PR #53, opened 2015-01-04
Let FindOpenSSL.cmake only search in Hunter install location.

alamaison:
Previously, Hunter would build and install OpenSSL, but then `find_package(OpenSSL)` would find a different version of the binaries installed elsewhere.  This is caused by FindOpenSSL.cmake looking in a set of hardcoded `HINTS` and `PATHS` other than the `OPENSSL_ROOT` set by Hunter.

Even though the Hunter install location was being searched first, the problem manifested because `find_library` favours shared libraries.  It would search the Hunter OpenSSL installation, see only static libs, and continue looking in the HINTS until finding a shared library,

This commit removes these extra paths from the search.

### DIFF
--- a/cmake/find/FindOpenSSL.cmake
+++ b/cmake/find/FindOpenSSL.cmake
@@ -13,6 +13,7 @@
 # Copyright 2006-2009 Kitware, Inc.
 # Copyright 2006 Alexander Neundorf <neundorf@kde.org>
 # Copyright 2009-2011 Mathieu Malaterre <mathieu.malaterre@gmail.com>
+# Copyright 2015 Alexander Lamaison <alexander.lamaison@gmail.com>
 #
 # Distributed under the OSI-approved BSD License (the "License");
 # see accompanying file Copyright.txt for details.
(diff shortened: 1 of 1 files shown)


cmake/find/FindOpenSSL.cmake
```diff
@@ -268,9 +232,7 @@ function(from_hex HEX DEC)
 endfunction()
 
 if (OPENSSL_INCLUDE_DIR)
-  if (_OPENSSL_VERSION)
-    set(OPENSSL_VERSION "${_OPENSSL_VERSION}")
-  elseif(OPENSSL_INCLUDE_DIR AND EXISTS "${OPENSSL_INCLUDE_DIR}/openssl/opensslv.h")
+  if(OPENSSL_INCLUDE_DIR AND EXISTS "${OPENSSL_INCLUDE_DIR}/openssl/opensslv.h")
```
ruslo: `OPENSSL_INCLUDE_DIR` already checked here, so probably it's simply:
```cmake
if(EXISTS "${OPENSSL_INCLUDE_DIR}/openssl/opensslv.h")
```

ruslo:
@alamaison Hi, Alex. Can you take a look at this PR. It looks good for me I'm just think that one 'if' variable is redundant (see my note)

alamaison:
Sorry, I forgot about this PR.  Fixed now.

ruslo:
Travis build failed because of incorrect REGEX in 'FindOpenSSL'. I've made a [patch](https://github.com/ruslo/hunter/commit/6475dde309c92449368504df73b48ce925f12dfe), please rebase this PR to fix the CI error.

PR #59, opened 2015-03-31
Added Expat XML parser.

alamaison:
http://expat.sourceforge.net/

Usage:
```cmake
hunter_add_package(EXPAT)

find_package(Expat REQUIRED)

add_executable(foo foo.c)
target_compile_definitions(foo PRIVATE XML_STATIC)
target_link_libraries(foo PRIVATE ${EXPAT_LIBRARIES})
target_include_directories(foo PRIVATE ${EXPAT_INCLUDE_DIRS})
```


examples/Expat/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+# All rights reserved.
+
+cmake_minimum_required(VERSION 3.0)
+project(example-hunter-expat)
+
+# Emulate HunterGate:
+# * https://github.com/hunter-packages/gate
+include("../common.cmake")
+
+hunter_add_package(EXPAT)
```
ruslo: I think it must be 'hunter_add_package(Expat)' here. Probably on OS like Mac or Windows it doesn't matter but on case sensitive platforms like Linux it will be an error.

alamaison:
Thanks.  Fixed.

ruslo:
@alamaison Hi, Alex! I've fixed some issues with applying toolchain flags to C project (see [details](https://github.com/ruslo/hunter/issues/269#issuecomment-159488545)). Now some CI testing start failing at `analyze` toolchain (issue was that wrapper for clang analyzer was not used for C code but only for C++ code).

See `Expat` errors: https://travis-ci.org/ruslo/hunter/jobs/93227575#L606

I will disable it for now. Let me know if you want to fix it - I will enable it back after update.

Thanks, Ruslo

alamaison:
> Let me know if you want to fix it - I will enable it back after update.

Sorry @ruslo, I've not had much free time lately.  Will look at this if I get a chance, but it might not be soon :(  Looks like issues in the upstream projects.

PR #65, opened 2015-04-09
Handle skipping toolchain verification internally

alamaison:
Specifying HUNTER_SKIP_TOOLCHAIN_VERIFICATION should be something Hunter does internally, when necessary.  Client projects shouldn't have to do this themselves.


ruslo:
Does make sense, thanks.

PR #89, opened 2015-04-22
Initial cvmatio and dlib packages

headupinclouds:
Provide initial hunter packages for cvmatio (Matlab *.mat => opencv cv::Mat) and header only dlib module.


cmake/projects/cvmatio/hunter.cmake
```diff
@@ -0,0 +1,43 @@
...
+  set(HUNTER_CMAKE_PROJECTS_CVMATIO_HUNTER_CMAKE_ 1)
+endif()
+
+# Load used modules
+include(hunter_add_version)
+include(hunter_download)
+include(hunter_pick_scheme)
+include(hunter_add_package)
+
+hunter_add_package(TIFF)
```
ruslo: There is no need for dependencies here, they will be taken from package itself

cmake/projects/cvmatio/hunter.cmake
```diff
@@ -0,0 +1,43 @@
...
+    URL
+    "https://github.com/headupinclouds/cvmatio/archive/v1.0.18.tar.gz"
+    SHA1
+    4bf8f9e9edc34a346220dc8ea307185d95283e93
+)
+
+# Pick a download scheme
+hunter_pick_scheme(
+    DEFAULT
+    url_sha1_release_debug # For separate release/debug generators, like Makefile
```
ruslo: now unified scheme can be used, see updated wiki: https://github.com/ruslo/hunter/wiki/usr.adding.new.package

cmake/projects/dlib/hunter.cmake
```diff
@@ -0,0 +1,38 @@
...
+    VERSION
+    "18.14"
+    URL
+    "https://github.com/davisking/dlib/archive/v18.14.tar.gz"
+    SHA1
+    85db9a8107cf5f65006921783164a0b2c0f3a5c0
+)
+
+# Pick a download scheme
+hunter_pick_scheme(DEFAULT url_sha1_no_install) # (for cmake modules and header-only libraries)
```
ruslo: let's make config-mode from this package since it's header only library if must be a trivial task. What headers need to be installed? All '*.h' from [dlib](https://github.com/davisking/dlib/tree/master/dlib) directory?

headupinclouds: Yes.  All the dlib files (mix of *.hpp and *.cpp) should be installed such
that files are accessible via the initial dlib path:

#include <dlib/image_processing.h>

#include <dlib/all/source.cpp>

There is a CMakeLists.txt and some bundled jpeg, png, and zlib code, but
that is only for building the sample applications and test code.

Some discussion here:

http://sourceforge.net/p/dclib/discussion/442517/thread/efc8d59d/

and here:

http://dlib.net/compile.html

ruslo: Open: https://github.com/ruslo/hunter/issues/95

PR #91, opened 2015-04-23
Avoid null check for IPHONESIMULATOR_ARCH since we're supporting single ...

headupinclouds:
...architecture native builds on ios (i.e., arm64, etc)


ruslo:
@headupinclouds Thanks for the quick fix, see more wider changes also: https://github.com/ruslo/hunter/commit/061d8a984bb6b6164654f367f0e5de26ad873c9f

PR #118, opened 2015-04-27
Added wxWidgets.

alamaison:
Usage:
    hunter_add_package(wxWidgets)
    find_package(wxWidgets REQUIRED core base)
    include(${wxWidgets_USE_FILE})
    add_executable(foo foo.cpp)
    target_link_libraries(foo ${wxWidgets_LIBRARIES})

Depends on GTK and OpenGL (GL and GLU) on Linux.  GTK and OpenGL aren't in Hunter.


ruslo:
Nice :)

alamaison:
This doesn't Mac because the CMakeified wxWidgets hasn't had its OSX Bakefiles ported (we don't have Macs to test the result).  It may work if GTK is present, but that hasn't been tested.

ruslo:
New page: https://github.com/ruslo/hunter/wiki/pkg.wxwidgets

ruslo:
> Depends on GTK and OpenGL (GL and GLU) on Linux. GTK and OpenGL aren't in Hunter.

* https://github.com/ruslo/hunter/issues/119
* https://github.com/ruslo/hunter/issues/120

ruslo:
> It may work if GTK is present, but that hasn't been tested.

I saved the note: https://github.com/ruslo/hunter/wiki/pkg.wxwidgets#issues

PR #135, opened 2015-05-04
libm fixes for ccv build on Linux (correct repository/link)

headupinclouds:


cmake/projects/ccv/hunter.cmake
```diff
@@ -18,13 +18,13 @@ hunter_add_version(
     PACKAGE_NAME
     ccv
     VERSION
-    "0.7-p4"
+    "0.7-p5"
      URL
-     "https://github.com/headupinclouds/ccv/archive/v0.7-p4.tar.gz"
+     "https://github.com/hunter-packages/ccv/archive/v0.7-p5.tar.gz"
```
ruslo: @headupinclouds Heh :) now I think this change is not in release: https://github.com/headupinclouds/ccv/commit/c25b2fd1615f5c3477a60f9120a966d1c47efcdd

Issue #121, opened 2015-04-28
ABI detection failures


alamaison:
Sometime, while building dependencies, I see that detecting compiler ABI fails for a subset of the dependencies.  For others in the same build, it works fine.  Two questions:

1) Should the build not be aborted if the ABI detection fails.  I just had a really subtle bug that manifested really late on: ABI detection failed which turned off exception handling.  The build continued, everything looked OK, but really it was hoplessly broken.

2) Does Hunter ever calculate the location of CMake subprocesses?  I did a lot of grepping but couldn't find anywhere it did that, but the failure seems to be caused by CMake calculating its own location relative to the current build directory and not being able to find its own modules:
```
Detecting CXX compiler ABI info failed to compile with the following output:
Change Dir: C:/Users/Alexander/.hunter/_Base/7018ac8/bf3e2d1/a44f91b/Build/wxWidgets/Build/wxWidgets-Debug-prefix/src/wxWidgets-Debug-build/CMakeFiles/CMakeTmp

Run Build Command:"C:/Program Files (x86)/Microsoft Visual Studio 8/Common7/IDE/devenv.com" "CMAKE_TRY_COMPILE.sln" "/build" "Debug" "/project" "cmTryCompileExec3546700233"


Microsoft (R) Visual Studio Version 8.0.50727.867.

Copyright (C) Microsoft Corp 1984-2005. All rights reserved.

1>------ Build started: Project: cmTryCompileExec3546700233, Configuration: Debug x64 ------

1>Compiling...

1>Microsoft (R) C/C++ Optimizing Compiler Version 14.00.50727.762 for x64

1>Copyright (C) Microsoft Corporation.  All rights reserved.

1>cl /Od /D "_DEBUG" /D "CMAKE_INTDIR=\"Debug\"" /D "_MBCS" /FD /RTC1 /MDd /Fo"cmTryCompileExec3546700233.dir\Debug\\" /Fd"cmTryCompileExec3546700233.dir\Debug\vc80.pdb" /c /Zi /TP   /MP /MP

1>   "..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\Program Files (x86)\CMake\share\cmake-3.2\Modules\CMakeCXXCompilerABI.cpp"

1>CMakeCXXCompilerABI.cpp

[...]

PR #141, opened 2015-05-08
Pass C/CXX ABI info to subprojects from top-level configuration.

### LINKED ISSUE #122: Forward all ABI info
Save ABI info on first run and forward it to all projects.

According to documentation: https://github.com/ruslo/CMake/blob/c90633df9eacb3063669ac0aad9506cea1885582/Modules/CMakeDetermineCompilerABI.cmake

Next optimization is possible:
* Save variables that is set by this module on first run, list:
  * `CMAKE_{C,CXX}_SIZEOF_DATA_PTR`
  * `CMAKE_{C,CXX}_COMPILER_ABI`
  * `CMAKE_{C,CXX}_IMPLICIT_LINK_LIBRARIES`
  * `CMAKE_{C,CXX}_IMPLICIT_LINK_DIRECTORIES`
  * `CMAKE_{C,CXX}_IMPLICIT_LINK_FRAMEWORK_DIRECTORIES`
  * `CMAKE_{C,CXX}_LIBRARY_ARCHITECTURE`
* Set variables to skip checking on next run:
 * `CMAKE_{C,CXX}_ABI_COMPILED`

Test it on visual studio 8 2005 with winsparkle project. See bug: https://github.com/ruslo/hunter/issues/121

alamaison:
Speeds up the build slightly, because it doesn't have to be recalculated in every subproject.  Also helps avoid #121 because the top-level project is less likely to run into the long-path problem while compiling the ABI test.

Closes #122.

### DIFF
--- a/cmake/modules/hunter_create_cache_file.cmake
+++ b/cmake/modules/hunter_create_cache_file.cmake
@@ -51,6 +51,51 @@ function(hunter_create_cache_file cache_path)
       "set(HUNTER_STATUS_DEBUG \"${HUNTER_STATUS_DEBUG}\" CACHE INTERNAL \"\")\n"
   )
 
+  # Pass compiler ABI info through to avoid recalculating in every project
+  foreach(lang C CXX)
+    if(DEFINED CMAKE_${lang}_SIZEOF_DATA_PTR)
+      file(
+          APPEND
+          "${temp_path}"
+          "set(CMAKE_${lang}_SIZEOF_DATA_PTR \"${CMAKE_${lang}_SIZEOF_DATA_PTR}\" CACHE INTERNAL \"\")\n"
+      )
+    endif()
+    if(DEFINED CMAKE_${lang}_COMPILER_ABI)
+      file(
+          APPEND
+          "${temp_path}"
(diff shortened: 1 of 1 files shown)


cmake/modules/hunter_create_cache_file.cmake
```diff
@@ -51,6 +52,58 @@ function(hunter_create_cache_file cache_path)
       "set(HUNTER_STATUS_DEBUG \"${HUNTER_STATUS_DEBUG}\" CACHE INTERNAL \"\")\n"
   )
 
+  # Pass compiler ABI info through to avoid recalculating in every project
+  foreach(lang C CXX)
+    if(DEFINED CMAKE_${lang}_SIZEOF_DATA_PTR)
```
ruslo: Why we need to check that variable is defined? May be just set it to cache unconditionally?

alamaison: > Why we need to check that variable is defined? May be just set it to cache unconditionally?

Becuase [CMakeDetermineCompilerABI.cmake sets it](https://github.com/Kitware/CMake/blob/master/Modules/CMakeDetermineCompilerABI.cmake#L65) only if `ABI_SIZEOF_DPTR` or `CMAKE_${lang}_SIZEOF_DATA_PTR_DEFAULT` are true.  Maybe one of those is always true, but I thought it safer not to assume.

alamaison:
Thanks again.

PR #165, opened 2015-06-10
Adding sparsehash/ package

aadityakalsi:
I probably have to fix this ```hunter.cmake```, since this is header only, and we don't need ```Release``` and ```Debug``` variants.

Other than that, I think this works well.

Sparsehash has BSD 3 clause license, so the license will have to go along with the source wherever used.


ruslo:
> since this is header only, and we don't need Release and Debug variants

In general header only library can have different additional defines. E.g. for Debug variant `FOO_DEBUG=1`. This option can be specified in installed `FooTargets-debug.cmake`. Also since library is header only it "builds" quite fast, so in my opinion this optimization doesn't worth any efforts.

ruslo:
@aadityakalsi Merged. Thanks!

PR #176, opened 2015-06-17
Updated hunter recipe to use the patched v1.58.0-p0 boost release

headupinclouds:
(Note: previous VERSION tag for v1.58.0.tar.gz file was also 1.58.0-p0, so that will be left as is for the new v1.58.0-p0.tar.gz file)

Just tested the 64 bit android build and all is working. 

Since the VERSION isn't changing in the update, this would require manually deleting the cache, but this only impacts 64 bit android builds that weren't working before this anyway, so I think we are okay.  Keeping the VERSION in sync with the RELEASE is probably the right thing...

### DIFF
--- a/cmake/projects/Boost/hunter.cmake
+++ b/cmake/projects/Boost/hunter.cmake
@@ -23,9 +23,9 @@ hunter_add_version(
     VERSION
     "1.58.0-p0"
     URL
-    "https://github.com/hunter-packages/boost/archive/v1.58.0.tar.gz"
+    "https://github.com/hunter-packages/boost/archive/v1.58.0-p0.tar.gz"
     SHA1
-    bab1bf06fabccbe7843350023b9393a7ab4e670b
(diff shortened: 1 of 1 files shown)


ruslo:
>  previous VERSION tag for v1.58.0.tar.gz file was also 1.58.0-p0, so that will be left as is for the new v1.58.0-p0.tar.gz file

Ooops... by bad.

> Since the VERSION isn't changing in the update, this would require manually deleting the cache

This should work without deleting cache. Cache designed so it can be shared between several `hunter-id`, i.e. same version `boost 1.58` can refer to different locations and hence have different archive sha1 values. The only flaw cache has is when you change internals of the hunter archive (like schemes). I will think about adding check of this files if we have more problems with it.

ruslo:
Though you have to clean `_Base/.../<config-id>` if you use develop version cloned from git.

Issue #181, opened 2015-06-22
Building boost (1.58.0-p0) on visual studio 2008 fails due to configuration checks being reported as errors


oliverdaniell:
Visual studio 2008
Boost 1.58.0-p0

It looks the errors from the boosts compiler checks are falling through to cmake when it is built with vs 2008. This does not occur with vs 2013. As a result cmake stops the build. It should be safe to ignore these errors as the install step still works as expected.

```
1>Performing install step for 'Boost'
1>32.cpp
1>Performing configuration checks
1>    - 32-bit                   : yes
1>arm.cpp
1>libs\config\checks\architecture\arm.cpp(12) : fatal error C1189: #error :  "No
t ARM"
1>    - arm                      : no
1>mips1.cpp
1>libs\config\checks\architecture\mips1.cpp(10) : fatal error C1189: #error :  "
Not MIPS1"
1>    - mips1                    : no
1>power.cpp
1>libs\config\checks\architecture\power.cpp(13) : fatal error C1189: #error :  "
Not PPC"
1>    - power                    : no
1>sparc.cpp
1>libs\config\checks\architecture\sparc.cpp(10) : fatal error C1189: #error :  "
Not SPARC"
1>    - sparc                    : no
1>x86.cpp
```

oliverdaniell:
It looks like this is due to this bug http://lists.boost.org/boost-users/2013/04/78408.php.

oliverdaniell:
Also relevant https://git.backbone.ws/cmake/cmake/commit/45d2966dcc0ebc08a925d6ac2ba1831549faa345?diffmode=sidebyside

oliverdaniell:
It looks like the following fix works. I've tested it on vs 2008 only

url_sha1_boost.cmake.in

```
 if("@MSVC@")
   set(env_cmd "@HUNTER_MSVC_VCVARSALL@" "@HUNTER_MSVC_ARCH@")
+  set(msvc_cmd set VS_UNICODE_OUTPUT=)
 else()
   set(env_cmd "")
+  set(msvc_cmd "")
 endif()
```

```   INSTALL_COMMAND
-    ${install_cmd}
+	INSTALL_COMMAND ${msvc_cmd} COMMAND ${install_cmd}
```
Apologies for the lack of a proper patch, I need to lean how to use github properly

ruslo:
@alamaison Is this related to https://github.com/ruslo/hunter/pull/145 ?

ruslo:
[...]

PR #182, opened 2015-06-29
fix for boost vs unicode output

oliverdaniell:
Fix for https://github.com/ruslo/hunter/issues/181, tested on visual studio 2008


ruslo:
@oliverdaniell Thank you!

PR #184, opened 2015-06-29
added full path to cl in boost

oliverdaniell:
Not sure why this works :(
Tested on "Visual Studio 12 2013 Win64" and "Visual Studio 12 2013"


ruslo:
@oliverdaniell tested with 1.55 and 1.58 boost. Thanks!

PR #191, opened 2015-07-01
Pass sysroot to boost compiler for internal link tests (cross compila…

headupinclouds:
…tion scenario)


ruslo:
This change will break Visual Studio build. See: https://github.com/ruslo/hunter/commit/af66eeb51988e4e09b6d1895a0f8a314703ecf3e

PR #196, opened 2015-07-05
Added unpack install scheme plus fixes to Findodb*

pretyman:
Here are the packages for C++ Object-Relational Mapping (ORM) odb - http://www.codesynthesis.com/products/odb/

This commit contains:
* SQLite3 with autotools and flags -DSQLITE_ENABLE_UNLOCK_NOTIFY=1 -DSQLITE_ENABLE_RTREE=1 (maybe this package is unwanted by hunter, since amalgamation of SQLite3 is usually preferred)
* odb core runtime
* odb compiler
* odb-sqlite for persisting in SQLite3 databases 
* examples/odb-sqlite with a working example of the compiler command
* find scripts for SQLite3, odb and odb-compiler

From the previous code review I have added:
* url_sha1_unpack_install.cmake.in : unpacks and install a tarball

Removed the addition of ODB_ROOT to CMAKE_FIND_ROOT_PATH of
* Findodb-compiler.cmake
* Findodb.cmake


ruslo:
> The pull request ended up being big, but there should be less stuff to change this time. :-)

Of course it's convenient for test/review reasons to have small commits. SQLite part applied: https://github.com/ruslo/hunter/commit/b89aa8bf66f41dba611016148a8e9bf6e2912f0d

ruslo:
odb-compiler: https://github.com/ruslo/hunter/commit/b81384d95e9be1a6769c7d1ec668adf7b8d2bb26

ruslo:
Findodb: https://github.com/ruslo/hunter/commit/8bb44d1595fd1903be79876bca4dd6ea20f3beef

ruslo:
odb-sqlite: https://github.com/ruslo/hunter/commit/35f507236f5b83f9acbea50d2fa9726e8da7eb1f

ruslo:
Remove old autotools modules: https://github.com/ruslo/hunter/commit/f7b70eba3df3229403d14a8f88fd9c7d229bd125

ruslo:
> maybe this package is unwanted by hunter, since amalgamation of SQLite3 is usually preferred

We can use unpack scheme here I guess

pretyman:
Ah, ok, just by separate commits would have helped then? Because as I open the pull request, all commits get taken into account, and I ended up seeing one big diff, however I missed the point where you could pick the separate commits.

> We can use unpack scheme here I guess

I don't use the amalgamation myself, but could try it at some point and see how it could fit in to have the define flags for  `SQLITE_ENABLE_UNLOCK_NOTIFY` and `SQLITE_ENABLE_RTREE`. Also I am unsure how to pass the amalgamation to `odb-sqlite`. So these 2 points need to be cleared to go the amalgamation route.

pretyman:
Thanks for taking the pull request and извините за ошибки, которые я делал, everything worked well in end :+1:

ruslo:
> I don't use the amalgamation myself

Me too.

> Also I am unsure how to pass the amalgamation to odb-sqlite

So I guess you can add new package `sqlite-amalgamation` with unpack-only scheme which can be used like this:
```cmake
hunter_add_package(sqlite-amalgamation)
add_target(... ${SQLITE-AMALGAMATION_ROOT}/sqlite.c)
```

pretyman:
That would make sense

PR #197, opened 2015-07-05
Added odb-sqlite example from official odb example in a separate pull request

pretyman:
Added to examples/odb-sqlite an example of using odb-sqlite with hunter


ruslo:
https://github.com/ruslo/hunter/commit/4efc7883768edd0d2e10217dd3fed92d46c22d79

PR #199, opened 2015-07-10
Autotools parallel build with HUNTER_JOBS_OPTION

pretyman:
I just noticed while working on the Qt project that I was not calling `make` with the `-j ${HUNTER_JOBS_OPTION}`.

So I updated the autotools to use it. However Qt and Autotools based project are calling this `make` command. I will have to think of a way to avoid this repetition.


pretyman:
я совсем дурак, I opened the pull request against master *again*, sorry about that

PR #209, opened 2015-07-26
Autotools defaults to static libraries

pretyman:
Configure Autotools generate a static lib by default, as this is Hunter's convention

Tested with the odb-sqlite example


ruslo:
Thanks

PR #213, opened 2015-07-29
add hunter_cacheable(Eigen)

headupinclouds:
See documentation: https://github.com/ruslo/hunter/wiki/dev.modules#hunter_cacheable

### DIFF
--- a/cmake/projects/Eigen/hunter.cmake
+++ b/cmake/projects/Eigen/hunter.cmake
@@ -12,6 +12,7 @@ include(hunter_add_version)
 include(hunter_cmake_args)
 include(hunter_download)
 include(hunter_pick_scheme)
+include(hunter_cacheable)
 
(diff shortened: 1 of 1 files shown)


cmake/projects/Eigen/hunter.cmake
```diff
@@ -13,6 +13,8 @@ include(hunter_cmake_args)
 include(hunter_download)
 include(hunter_pick_scheme)
```
ruslo: we need `include(hunter_cacheable)` here

headupinclouds:
@ruslo: everything seems to pass except the expat test w/ the default toolchain.  from what i can see, expat doesn't even use eigen.

ruslo:
> everything seems to pass except the expat test w/ the default toolchain

Just a Travis CI glitches. Thanks, merged.

PR #214, opened 2015-07-30
Autotools iOS

pretyman:
Ура!! У нас iOS support!

Ok, this is the real thing, I managed to build odb-sqlite and run it in the iPhone :-)

For reviewing puporses: 
`hunter_autotools_project.cmake` - I pushed the shared code to the top, then dealt with the platform specific at the bottom
`scripts/autotools-lipo.cmake.in` - script that copies all non *.a to thei final destination and `lipo`s all *.a files together, might need to rename it to reflect its complete behavior

At the end, all built directories are removed as expect, seems to work with cacheable packages too


pretyman:
I can re-issue the pull-request with a commit for each file, if you find it easier to review this way

ruslo:
> I can re-issue the pull-request with a commit for each file, if you find it easier to review this way

No need. Looks good like that.

ruslo:
> Ура!! У нас iOS support!

Thanks for working on this! Looks awesome indeed :)

pretyman:
Извините, эту неделю я сдал вам много работу.

I did the merge of the header files, it seems to work well.

Any other different files in the multi-arch tree, I move them to the final installation directory appending a dash and their architecture.

I am sorry, I did quite a few changes, if you think it is easier for you to review, I can squash commit and commit one file at a time.

I chose to install the multi-arch tree into `@HUNTER_PACKAGE_INSTALL_DIR@/multi-arch`, then I move all files which are the same, merge different files with the rules above, but however, there *could* be some case where different file names are created for different arch, for reasons we don't know, so I manually clean `@HUNTER_PACKAGE_INSTALL_DIR@/multi-arch` empty directories leaving non empty ones intact.

Like this, if `@HUNTER_PACKAGE_INSTALL_DIR@/multi-arch` exists, its content can be used to investigate what filenames were different for different archs. It sounded more reasonable then building the multi-arch inside the `SOURCE_DIR` of the universal `ExternalProject_Add` target, which gets deleted by CMake.

ruslo:
> if you think it is easier for you to review, I can squash commit and commit one file at a time

@pretyman I think it's quite a huge change so I have some thoughts about history clean-ups. Let's make it work/test first.

pretyman:
> I think it's quite a huge change so I have some thoughts about history clean-ups. Let's make it work/test first.

Ok, let's hammer this into shape, then I'll create a new pull request with a single commit for each file to avoid the history mess.

ruslo:
> with a single commit for each file to avoid the history mess.

It doesn't always good for reviewing. I will add some notes later.

ruslo:
> I will add some notes later

Let's start with this: https://github.com/ruslo/hunter/wiki/dev.contribution

pretyman:
Хорошо, could you also add to https://github.com/ruslo/hunter/wiki/dev.contribution your vim configuration for tabs to spaces for `*.cmake` and `CMakeLists.txt`? Also the highlighter of trailing spaces? I searched the web, but couldn't get to work what I found.

ruslo:
@pretyman Done: https://github.com/ruslo/hunter/wiki/dev.contribution#vim-hints

pretyman:
Спасибо большое, работает. I ended up using `autocmd Filetype cmake` at the beginning to have the rules applied to cmake files only: 

```vim
autocmd BufRead,BufNewFile CMakeLists.txt setlocal filetype=cmake
autocmd Filetype cmake highlight TabsAndTrailingSpaces ctermbg=red guibg=red
autocmd Filetype cmake match TabsAndTrailingSpaces /\t\| \+$/
autocmd Filetype cmake let &colorcolumn=join(range(81,9999),",")
autocmd Filetype cmake setlocal ts=2 sts=2 noexpandtab sw=2
```

pretyman:
Thanks for the review, I will close this pull request for now and open a new one with the dev.contribution guidelines actively taken into account when I advance. Sorry for the trouble, I am yet not very acquainted with Git :-)

ruslo:
> Thanks for the review

Actually thank **you** for working on this :)

> Sorry for the trouble, I am yet not very acquainted with Git

No problems. Let me know if you have any questions.

PR #215, opened 2015-07-31
Added header only cereal c++11 serializtion lib and travis example app

headupinclouds:
original: https://github.com/USCiLab/cereal
hunter fork: https://github.com/headupinclouds/cereal


ruslo:
@headupinclouds Thanks. Wiki page: https://github.com/ruslo/hunter/wiki/pkg.cereal

PR #217, opened 2015-08-09
Package geos 3.4.2

pretyman:
This package is interesting because:
* creates a script file in the `bin` dir, which won't be `lipo`'ed
* needs some extra linking flags on Android


pretyman:
I pushed a simple example to trigger the build.

cmake/find/Findgeos.cmake
```diff
@@ -0,0 +1,78 @@
...
+  if(NOT GEOS_C_LIBRARY)
+    message(FATAL_ERROR "GEOS_C_LIBRARY not found!")
+  endif()
+  add_library("geos::c" UNKNOWN IMPORTED)
+  set_target_properties(
+      "geos::c"
+      PROPERTIES
+        IMPORTED_LOCATION "${GEOS_C_LIBRARY}"
+  )
+  add_dependencies(
```
ruslo: add_dependencies make sense only for building targets. Usually dependency set by CMake automatically (like in target_link_libraries) and needed only in very special cases when it's quite hard to find out that automatically (like dependency between several ExternalProject_Add targets). If you want to link `geos` target to every target that use `geos::s` you can use IMPORTED_LINK_INTERFACE_LIBRARIES: https://github.com/ruslo/hunter/commit/c783edc6cf36b709bdd6b5526aeb4a149dfe4e60

ruslo:
> creates a script file in the bin dir, which won't be lipo'ed

https://github.com/ruslo/hunter/issues/218

ruslo:
> needs some extra linking flags on Android

https://github.com/ruslo/hunter/issues/219

cmake/find/Findgeos.cmake
pretyman: ooops, thanks, I got mixed up with that one.

PR #221, opened 2015-08-16
qtquickcontrols fully depends on qtdeclarative

pretyman:
qtquickcontrols fully depends on qtdeclarative


ruslo:
@pretyman Merged, thank you.

PR #233, opened 2015-09-15
Add Avahi + dbus dependency support 

### LINKED ISSUE #228: Add dbus
See: https://github.com/ruslo/hunter/pull/227

daminetreg:
As of #227 this contains Avahi and all it's dependencies : 
  - intltoolize
  - libdaemon
  - dbus

Avahi is taken from the official webpage, however dbus had to be patched to use the cmake build, as their cmake build was not in the top-level folder. ExternalProject_Add doesn't support projects without top-level cmake, therefore the patched dbus is available here, and I think it would make sense to move it to the hunter-packages organization : https://github.com/daminetreg/dbus/tree/feature/small-cmake-patch-for-hunter.

Updating dbus is easy as the changes done are only auto replace of CMAKE_SOURCE_DIR / CMAKE_BINARY_DIR into dbus_SOURCE_DIR and dbus_BINARY_DIR + a top-level CMakeLists.

I've made a tag of the last stable dbus release + the hunter specific changes : https://github.com/daminetreg/dbus/releases/tag/dbus-1.10.0-hunter-1

Hope this is good now.
This closes #228 .


cmake/projects/dbus/hunter.cmake
```diff
@@ -0,0 +1,27 @@
...
+
+hunter_cmake_args(
+    dbus 
+    CMAKE_ARGS
+      DBUS_ENABLE_XML_DOCS=OFF 
+)
+
+hunter_pick_scheme(DEFAULT url_sha1_cmake)
+hunter_download(PACKAGE_NAME dbus
+  PACKAGE_DEPENDS_ON expat)
```
ruslo: @daminetreg This will not download `expat` package but only register it. I've update documentation to clarify this: https://github.com/ruslo/hunter/wiki/dev.modules#hunter_download

I've got an error while running `hunter_add_package(dbus)` only:

```
-- Check size of socklen_t
-- Check size of socklen_t - done
CMake Error at cmake/CMakeLists.txt:346 (message):
  expat not found!
```

Note that if `expat` will be installed system-wide it **will be** found. It's not good and may result conflicts, see [wiki](https://github.com/ruslo/hunter/wiki/usr.adding.new.package.with.dependencies).

By the was since `dbus` has an example already I think you can create separate pull request with only `dbus` package so we can avoid discussion of unrelated issues.

scripts/clear-all.sh
```diff
@@ -289,5 +289,5 @@ unset YACC
 unset arch
 unset variant
 
-# Problems with Xcode and boost
-export PATH=/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin
+# Problems with Xcode and boost and finding own binaries.
+export PATH=${HUNTER_PACKAGE_INSTALL_PREFIX}/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin
```
ruslo: Not needed

cmake/projects/dbus/hunter.cmake
daminetreg: I've put dbus in this same pull-request as it's a dependency of Avahi, don't you think it shall all get together ?

scripts/clear-all.sh
daminetreg: Sorry I forgot removing it.

cmake/projects/dbus/hunter.cmake
ruslo: @daminetreg But `dbus` doesn't depends on anything. I mean that we can add `dbus` package first and then add `Avahi`.

daminetreg: You're right :smile:. I'll open another PR only for Dbus then.

Regarding the fact expat get searched system-wide, it's something we could prevent well with CMake toolchain files options : 
```
set( CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY )
set( CMAKE_FIND_ROOT_PATH_MODE_PROGRAM BOTH )
set( CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY )
set( CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY )
```
What do you think ?

ruslo: > What do you think ?

Actually it's a nice idea but it has some implementation peculiarities, see: https://github.com/ruslo/hunter/issues/235

daminetreg:
I close this PR until dbus support is merged #234, I'll reopen it then as it bases on the dbus support.

ruslo:
> I close this PR until dbus support is merged

@daminetreg Merged

daminetreg:
Thank you very much for merging the dbus support, I made this PR again in #236. Sorry for creating so much PR. I hope we won't overflow the PR id storage. :smile:

PR #236, opened 2015-09-15
avahi dependency support

daminetreg:
Hi @ruslo ,

Continuing my PR collection : #227 #233 #234. :twisted_rightwards_arrows:

Finally a working Avahi integration thanks to your nice feedback and extensions to hunter_autotools_project.

I hope you won't get any missing dependencies, I checked and all libs (up to the optional gobject support one that I sadly) were found and taken from hunter install dir.

So normally it should always compile even on a naked system.

Cheers,

PR #237, opened 2015-09-16
avahi dependency support

daminetreg:
Same as #236 but targetting develop this time.

- I changed the include_directories in the Avahi example.
- I rebased on develop, and retested, it seems to work.


ruslo:
@daminetreg Thanks!

ruslo:
@daminetreg New wiki page: https://github.com/ruslo/hunter/wiki/pkg.avahi

daminetreg:
Cool that you add a doc for each package. I'll think to it for the next package.

ruslo:
@daminetreg Salut Damien! Just for your information I've added testing for Avahi package but all toolchains fail:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.155
* https://travis-ci.org/ingenue/hunter/builds/114462759

it means that any change for this package will not be tested.

Issue #123, opened 2015-04-28
Optimize HUNTER_BINARY_DIR layout


ruslo:
For now to avoid [path too long](https://github.com/ruslo/hunter/wiki/error.external.build.failed#windows) error external directory with short path can be specified: `HUNTER_BINARY_DIR`. But looks like Visual Studio 8 2005 use relative paths for sources, so in fact it becomes even more longer:
* Path to sources: `/<hunter-root>/_Base/<hunter-id>/<config-id>/<toolchain-id>/Build/<package>/Source`
* Path to build: `/<hunter-binary-dir>/<package>/Build`
* Relative: `/../../../../` (level up to `<hunter-root>`) + `<hunter-binary-dir>/<package>/Build`

Possible solution is to unpack sources to `HUNTER_BINARY_DIR` too:
* Path to sources: `/<hunter-binary-dir>/<package>/Source`
* Path to build: `/<hunter-binary-dir>/<package>/Build`

Note: how to deal with "unpack-only" packages from `/<hunter-root>/_Base/Download/<package>/Unpacked`?

Cyberunner23:
> Path to sources: ``` /<hunter-root>/_Base/<hunter-id>/<config-id>/<toolchain-id>/Build/<package>/Source```

> Possible solution is to unpack sources to HUNTER_BINARY_DIR too:
Path to sources: ```/<hunter-binary-dir>/<package>/Source ```

What i instead of unpacking them at a different place we unpack then at the usual place and symlink them to a simpler path and update the symlink as necessary?

Cyberunner23:
> 
    Path to sources: ```/<hunter-binary-dir>/<package>/Source```
    Path to build: ```/<hunter-binary-dir>/<package>/Build```

And also I suggest the paths being something like ```/<hunter-binary-dir>/Builds/<package>``` to keep hunter root somewhat clean when multiple packages are used.

ruslo:
> What if instead of unpacking them at a different place we unpack them at the usual place and symlink them to a simpler path and update the symlink as necessary?

what is the difference? `Source` directory will be removed after build so we not using more space

Cyberunner23:
Sorry I meant to include the other directories (```Build```) and so on, it could shorten the path length while still maintaining hunter-id, config-id and toolchain-id which I assume you need?

ruslo:
[...]

PR #239, opened 2015-09-21
Support for dependency-free nlohmann::json

daminetreg:
Hi @ruslo,

Hope you are well. :smile:

I cmaked the [nlohmann::json library](https://github.com/nlohmann/json) here : [nlohmann::json cmaked](https://github.com/nlohmann/json/compare/master...daminetreg:pull-request/support-cmake-package-file).

Then I made it available with hunter, therefore this pull-request. I would like to move my json fork in the hunter-packages organization if you accept, and I parallely open a PR by [nlohmann::json library #123](https://github.com/nlohmann/json/pull/123) to get the changes mainline.

Cheers,


ruslo:
> I would like to move my json fork in the hunter-packages organization

Created: https://github.com/hunter-packages/json

ruslo:
```
+set (json_VERSION "1.0.0-rc1")
```

If you will set version in `project` it will be used by `write_basic_package_version_file`. See [documentation](http://www.cmake.org/cmake/help/v3.0/module/CMakePackageConfigHelpers.html).

ruslo:
```
+  set(CMAKE_CXX_FLAGS "-std=c++11 -stdlib=libc++")
```
user may set `CMAKE_CXX_FLAGS` from command line, if you will use `set` they will be lost. Should be at least `set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -stdlib=libc++")`. Also this stuff is quite tricky. I'm pretty sure it will not work on Linux (default lib is `libstdc++`).

Also see [wiki](https://github.com/ruslo/hunter/wiki/usr.adding.new.package#default-behaviour) about defaults.

ruslo:
```
+check_required_components(nlohmann_json) 
```

Thanks, missed this stuff from documentation. I'm personally using [this](https://github.com/forexample/package-example) project as a template. Updated: https://github.com/forexample/package-example/commit/cf2ea1d6a209fb9eca2ab83fdd0ac15fe4d3e807

examples/nlohmann-json/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+
+hunter_add_package(nlohmann-json)
+find_package(nlohmann-json)
+
+add_executable(main main.cpp)
+
+target_link_libraries(main
+  nlohmann-json::nlohmann-json)
+set_target_properties(main PROPERTIES
+  COMPILE_FLAGS ${NLOHMANN_JSON_DEFINITIONS})
```
ruslo: if compile flags are usage requirement we can add them in `*Config.cmake` by [INTERFACE_COMPILE_OPTIONS](http://www.cmake.org/cmake/help/v3.0/prop_tgt/INTERFACE_COMPILE_OPTIONS.html) property. Or better by [target_compile_options](http://www.cmake.org/cmake/help/v3.0/command/target_compile_options.html).

ruslo:
```
STRING(REPLACE "/O2" "/Od" CMAKE_CXX_FLAGS_RELEASE ${CMAKE_CXX_FLAGS_RELEASE})
```
A little bit confusing. What is the point of this line? Though it came from original code...

examples/nlohmann-json/CMakeLists.txt
daminetreg: I tried with COMPILE_FLAGS and this was not a cmake whitelisted prop and couldn't find out the right one. Thanks for pointer INTERFACE_COMPILE_OPTIONS.

daminetreg:
> user may set CMAKE_CXX_FLAGS from command line, if you will use set they will be lost. Should be at least set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -stdlib=libc++"). Also this stuff is quite tricky. I'm pretty sure it will not work on Linux (default lib is libstdc++).

> Also see wiki about defaults.

Yes I didn't change it as it came from the original code. I find it strange but it might be some support of libstdc++ which is missing which is in libc++ that the original code use. I removed these switches, as for me both libstdc++ libc++ get the test passed to the same level of success.

Speaking for clang 3.5.1 rc2.

> STRING(REPLACE "/O2" "/Od" CMAKE_CXX_FLAGS_RELEASE ${CMAKE_CXX_FLAGS_RELEASE})

> A little bit confusing. What is the point of this line? Though it came from original code...

Looking at the history it look like the author disabled the MSVC optimization, it might be due to bugs that those incurred? I'm unsure, hopefully @nlohmann can tell us why. As you can see in this commit 
https://github.com/nlohmann/json/commit/c89ca716c5c9d9782db78f89c271bd831f5f96bc.

daminetreg:
> If you will set version in project it will be used by write_basic_package_version_file. See documentation.

Good to know, it however makes usage of -rc1 suffix impossible there. But that's not a problem as it's anyway not allowing greater-than version check. Thanks.

daminetreg:
> Thanks, missed this stuff from documentation. I'm personally using this project as a template. Updated: forexample/package-example@cf2ea1d

I'll take this one as basis in the future, nice organization name. :smile:

nlohmann:
I had no time to follow this thread yet. But just as info: the MSVC flags were set to compile my code at AppVeyor.

ruslo:
> But just as info: the MSVC flags were set to compile my code at AppVeyor.

@nlohmann I've got assertion, see https://github.com/nlohmann/json/issues/125

daminetreg:
Thanks @nlohmann for joining the thread :smile:. I'll make tries to see why it doesn't compile with all optimizations enabled on MSVC.

daminetreg:
Thanks @nlohmann for joining the discussion. :smile: And thanks @ruslo for debugging and merging. :+1:

VladEvka:
@daminetreg, don't build on Mac OS X. 
OS X: El Cspitan
cmake version 3.3.2

Error:
clang: error: no such file or directory: ' -std=c++11'

Fix bug:
 https://github.com/hunter-packages/json/blob/hunter/cmake/modules/nlohmann-json-config.cmake.in#L15

set(NLOHMANN_JSON_DEFINITIONS "@CMAKE_CXX_FLAGS@")
to
set(NLOHMANN_JSON_DEFINITIONS @CMAKE_CXX_FLAGS@)

ruslo:
> set(NLOHMANN_JSON_DEFINITIONS "@CMAKE_CXX_FLAGS@")

Actually we don't need this stuff at all. Can be added automatically...

ruslo:
Just for your information though library is header only we still need to download > 70 MB because of 'test' and 'benchmarks' directories:
```
> du -hs *
4.0K	CMakeLists.txt
4.0K	LICENSE.MIT
4.0K	Makefile
 16K	README.md
4.0K	appveyor.yml
 57M	benchmarks
4.0K	cmake
1.5M	doc
440K	src
 14M	test
```

ruslo:
> clang: error: no such file or directory: ' -std=c++11'

Fixed: https://github.com/ruslo/hunter/releases/tag/v0.12.32

ruslo:
> need to download > 70 MB

Archive is 16 MB, but anyway

VladEvka:
@ruslo, ths!!

nlohmann:
Hi there, it's been a while since PR https://github.com/nlohmann/json/pull/123 has been opened. I am not sure what it is good for, but it breaks the AppVeyor build. Could you please check if all proposed changes are actually required?

ruslo:
@nlohmann Compile flags in good vs. broken build are quite the same, I think problem is C++ code.

@daminetreg I think you need to rebase your pull request to fetch new updates from C++ sources.

ruslo:
> >    clang: error: no such file or directory: ' -std=c++11'

> Fixed: https://github.com/ruslo/hunter/releases/tag/v0.12.32

@daminetreg It make sense to pull this fixes too (see branch `hunter` in https://github.com/hunter-packages/json)

daminetreg:
Thanks @ruslo for the fixes.

I'll reopen another PR from the hunter-packages version.

@nlohmann I'll check for appveyor, but I think this was the exception handling model chosen that I removed : `/EHsc`, I think it's something we shouldn't had to the exported compile flags and only use it in the unit test compilation. I'll ensure appveyor build again.

PR #240, opened 2015-09-22
Fixed MSVC architecture detection regular expression

PyrO70:
MSVC regular expression doesn't account for the remaining '0' when
parsing "Visual Studio 10 2010..". The 3rd regex remove [1-9] where it
should be [0-9] as above comment shows the remaining '0'


ruslo:
@PyrO70 Nice catch, thanks. Please set destination branch to `develop` (see [contributing](https://github.com/ruslo/hunter#contribution)).

PR #241, opened 2015-09-22
Fixed MSVC architecture detection regular expression

PyrO70:
MSVC regular expression doesn't account for the remaining '0' when
parsing "Visual Studio 10 2010..". The 3rd regex remove [1-9] where it
should be [0-9].


ruslo:
@PyrO70 Thanks. Just for your information I've added a test: https://github.com/ruslo/hunter/commit/a40830ef8ce073efbbddc8e886e5863e4f10b17d

PR #242, opened 2015-09-24
Adds dbus version with valid default system_bus socket path

daminetreg:
Just fixed a default setting in dbus CMakeLists that is useful, as otherwise dbus uses an hunter install-prefixed dbus system socket which is not what some want when using dbus, as the standard use case is to communicate with the currently running dbus instance.


ruslo:
Thanks

daminetreg:
thanks for merging :)

PR #244, opened 2015-09-27
add latest version of armadillo library

ThomasFeher:
Does not use the build-in cmake files because they are ment to build a
convenience library that automatically determines installed dependencies
and links those.
This would prevent the possibility to install those dependencies via
hunter, because the cmake script would not find them. (Right??)

I will add older versions later.

Should the Armadillo repository be transferred to https://github.com/hunter-packages ?

### DIFF
--- a/cmake/projects/Armadillo/hunter.cmake
+++ b/cmake/projects/Armadillo/hunter.cmake
@@ -1,0 +1,23 @@
+if(DEFINED HUNTER_CMAKE_PROJECTS_ARMADILLO_HUNTER_CMAKE_)
+	return()
+else()
+	set(HUNTER_CMAKE_PROJECTS_ARMADILLO_HUNTER_CMAKE_ 1)
+endif()
+
+include(hunter_add_version)
+include(hunter_download)
+include(hunter_pick_scheme)
+
+hunter_add_version(
+	PACKAGE_NAME
+	armadillo
+	VERSION
+	"5.600.2"
+	URL
+	"https://github.com/thomasfeher/armadillo/archive/v5.600.2.tar.gz"
+	SHA1
+	2cb74a9228ffe9ccf4b48d7cccf344365867020b
+	)
+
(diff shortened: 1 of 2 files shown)


ruslo:
> Should the Armadillo repository be transferred to https://github.com/hunter-packages ?

@ThomasFeher Ok, see https://github.com/hunter-packages/armadillo

> This would prevent the possibility to install those dependencies via
hunter, because the cmake script would not find them

To be precise if they are not installed system-wide script will fail. If they are installed system-wide this may or may not conflict. See https://github.com/ruslo/hunter/wiki/usr.adding.new.package.with.dependencies. Of course it's better to add all dependencies to Hunter first but non-hunterized version can be used for now as a workaround.

> Does not use the build-in cmake files because they are ment to build a
convenience library...

It's better to patch CMake code than adding new scheme. I hope that sometime other schemes will be removed and only `url_sha1_cmake.cmake` will left. From what I see only headers of library is used. You can add an option to the CMake code (like `ARMADILLO_INSTALL_ONLY_HEADERS`) and install `INTERFACE` library. There are a lot of boilerplate code but It's quite trivial, see example: https://github.com/ruslo/leathers/blob/master/CMakeLists.txt#L91

ThomasFeher:
Ok, I will have a look into the install interface library approach. This will mainly mean bypassing all cmake code and just adding the install target. I will also try to get this upstream, but it will take a while, because I'm quite busy at the moment.

Just to clarify: Armadillo is a header only library. All dependencies are optional. The cmake script is, in my opinion, just for quickly testing it. For  serious use one will just use the headers and configure the used dependencies as needed. On Windows the cmake script won't even do anything, see https://github.com/hunter-packages/armadillo/blob/master/CMakeLists.txt#L59 .

ruslo:
> Ok, I will have a look into the install interface library approach

@ThomasFeher Cool, thanks

ruslo:
Seems abandoned. Feel free to reopen on updates.

PR #268, opened 2015-11-23
Add Async++ package

alapshin:
I had to [fork](https://github.com/alapshin/asyncplusplus/tree/hunter) original [Async++](https://github.com/Amanieu/asyncplusplus) repository to create tarball release and change Async++Config.cmake install path according to https://github.com/forexample/package-example

There is one problem. Async++ links with threading library using
```
set(THREADS_PREFER_PTHREAD_FLAG ON)
find_package(Threads REQUIRED)
target_link_libraries(Async++ PUBLIC Threads::Threads)
```
and attempt to link with it using
```
target_link_libraries(${PROJECT_NAME} Async++::Async++)
```
lead to error
```
CMake Error at src/CMakeLists.txt:5 (add_executable):
  Target "foo" links to target "Threads::Threads" but the target
  was not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?
```
As workaround I have to add to project CMakeLists.txt
```
find_package(Threads REQUIRED)
```
Is it library error? Can I fix it?

By the way, thanks for great project. I have to come back to С++ after couple of years of Android development and was missing how easy it was using 3rdparty libraries.


ruslo:
> I had to fork original Async++ repository to create tarball release

This place can be used for such tasks: https://github.com/hunter-packages

I've made a fork: https://github.com/hunter-packages/asyncplusplus
You have permissions to make a releases there.

ruslo:
> Is it library error? Can I fix it?

You can use [find_dependency](https://cmake.org/cmake/help/v3.3/module/CMakeFindDependencyMacro.html) macro in `*Config.cmake` file.

By the way you can add example for testing,
* to directory: https://github.com/ruslo/hunter/tree/master/examples
* instructions: https://github.com/ruslo/hunter/wiki/dev.hunter.root#jenkinspy

One more note: `Threads::Threads` available since [CMake 3.1](https://cmake.org/cmake/help/v3.1/module/FindThreads.html), there is no such target in [CMake 3.0](https://cmake.org/cmake/help/v3.0/module/FindThreads.html). Please add `cmake_minimum_required(VERSION 3.1)` to the `*Config.cmake` file.

ruslo:
> By the way, thanks for great project

Thanks for contribution, enjoy :)

cmake/projects/Async++/hunter.cmake
```diff
@@ -0,0 +1,25 @@
...
+    PACKAGE_NAME
+    Async++
+    VERSION
+    "0.0.1-hunter"
+    URL
+    "https://github.com/alapshin/asyncplusplus/archive/v0.0.1-hunter.tar.gz"
+    SHA1
+    e013afbfb12e5843dc0b3d543c101ed28c325c1d
+)
+
```
ruslo: Since this library installs `*Config.cmake` we can add [hunter_cacheable](https://github.com/ruslo/hunter/wiki/usr.adding.new.package#important) here for optimal usage.

alapshin:
Could you grant me push access to repository https://github.com/hunter-packages/asyncplusplus or should I fork it and send pull request?

ruslo:
> Could you grant me push access to repository

Oops, I forget to add a repository to the team :) Sorry, try again.

alapshin:
Thanks, I pushed new packages version and fixed pull request.

cmake/projects/Async++/hunter.cmake
```diff
@@ -0,0 +1,26 @@
+if(DEFINED HUNTER_CMAKE_PROJECTS_EXAMPLE_HUNTER_CMAKE_)
```
ruslo: Leave your copyrights here

cmake/projects/Async++/hunter.cmake
```diff
@@ -0,0 +1,26 @@
...
+    PACKAGE_NAME
+    Async++
+    VERSION
+    "0.0.2-hunter"
+    URL
+    "https://github.com/hunter-packages/asyncplusplus/archive/v0.0.2-hunter.tar.gz"
+    SHA1
+    19ad060bed5addd8975f27757231ff610117aeb6
+)
+
```
ruslo: You can add `hunter_cacheable(Async++)` here.

examples/Async++/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+# Emulate HunterGate:
+# * https://github.com/hunter-packages/gate
+include("../common.cmake")
+
+project(download-asyncpp)
+
+# download Async++
+hunter_add_package(Async++)
+
+# now gtest can be used
```
ruslo: `s,gtest,Async++,`

ruslo:
New wiki page: https://github.com/ruslo/hunter/wiki/pkg.async--

ruslo:
I've got runtime error on Linux with GCC:
```
terminate called after throwing an instance of 'std::system_error'
  what():  Enable multithreading to use std::thread: Operation not permitted
Aborted (core dumped)
```
This can be fixed by setting `THREADS_PREFER_PTHREAD_FLAG` to `TRUE` (`-pthread` argument to linker, see [this OS answer](http://stackoverflow.com/a/19619065/2288008)), add either to CMakeLists.txt of example before `find_package` or to `*Config.cmake` file before `find_dependency`.

alapshin:
Added THREADS_PREFER_PTHREAD_FLAG to *Config.cmake files and bumped package version

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.12.28

ruslo:
Just for your information I've disabled next tests that failed ([commit](https://github.com/ruslo/hunter/commit/bb1e23351ac7980ba4f3dc90d6b8146f0d611539)):

* `libstdcxx` https://travis-ci.org/ruslo/hunter/jobs/92904105#L557
```
In file included from /Users/travis/build/ruslo/hunter/examples/Async++/asyncpp.cpp:2:
/Users/travis/build/ruslo/hunter/_testing/Hunter/_Base/d557a43/55402bc/da32312/Install/include/async++.h:25:10: fatal error: 'atomic' file not found
#include <atomic>
         ^
1 error generated.
```
*I guess this one can be ignored because of lack of c++11 support in current GCC*

* `ios` https://travis-ci.org/ruslo/hunter/jobs/92904108#L494
```
/Users/travis/build/ruslo/hunter/_testing/Hunter/_Base/ad711bf/55402bc/097785c/Build/Async++/Source/src/scheduler.cpp:68:8: error: thread-local storage is unsupported for the current target
static THREAD_LOCAL wait_handler thread_wait_handler = generic_wait_handler;
       ^
```
Seems library can't be used on iOS platform.

* `analyzer`  https://travis-ci.org/ruslo/hunter/jobs/92904109#L514
```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/algorithm:2840:52: warning: The result of the '>>' expression is undefined
    __mask0_ = __w0_ > 0 ? _Engine_result_type(~0) >> (_EDt - __w0_) :
                           ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
1 warning generated.
```
Probably a real error (?) Need to investigate.

Let me know if you plan to report/fix this. I'll revert testing back.

alapshin:
I agree that failing test due to missing atomic header should be ignored 
since library explicitly requires C++11 compiler.
As for analyzer warning, I'll look into it.

Sent with AquaMail for Android
http://www.aqua-mail.com

alapshin:
2) Looks like Apple's version of clang doesn't support thread_local from C++11 standard [SO answer](
http://stackoverflow.com/questions/28094794/why-does-apple-clang-disallow-c11-thread-local-when-official-clang-supports/29929949#29929949) While it's unfortunate I don't think it's a library problem since it specifically requires C++11-compliant compiler and standard library.

3) According to this (SO answer)[http://stackoverflow.com/questions/29128320/clang-static-code-analyzer-the-result-of-the-expression-is-undefined] this is false positive from clang-analyzer.
By the way does clang-analyzer have option to ignore warnings from system headers similar to gcc (system headers)[https://gcc.gnu.org/onlinedocs/cpp/System-Headers.html] or this is not possible and bad idea?

ruslo:
> According to this SO answer this is false positive from clang-analyzer

I've found this bug, think it relates: https://llvm.org/bugs/show_bug.cgi?id=17634
Especially: `The analyzer doesn't handle bit shifts right now`

> By the way does clang-analyzer have option to ignore warnings from system headers

No from what I see in [FAQ](http://clang-analyzer.llvm.org/faq.html#exclude_code). I guess using `__clang_analyzer__` is the only way to suppress message explicitly.

> or this is not possible and bad idea?

I think yes, in some cases it's a bad idea since same code can be used correctly (don't produce warning) and wrongly (produce warning). It depends on conditions/branches.

I've tried to add assert to the `alogorithm` header and it fix warning:
`_LIBCPP_ASSERT((_Edt - __w0_) < sizeof(_Engine_result_type), "")`
(using hint from llvm bugreport).

PR #283, opened 2015-12-05
Added Lua package

Dinno:


cmake/projects/Lua/hunter.cmake
```diff
@@ -0,0 +1,23 @@
...
+include(hunter_download)
+include(hunter_pick_scheme)
+
+hunter_add_version(
+    PACKAGE_NAME
+    Lua
+    VERSION
+    "5.3.2"
+    URL
+    "https://github.com/hunter-packages/lua/archive/5.3.2.tar.gz"
```
ruslo: I've applied some fixes for Linux:
* https://github.com/hunter-packages/lua/commit/54d84610f3a2fda8b0afdfc45d2977c27d788ea9
* https://github.com/hunter-packages/lua/commit/1054f9fa7a86dd15ecdc085430c038883bbdf0ae

Please update archive: https://github.com/hunter-packages/lua/releases/tag/v5.3.2-p0

cmake/projects/Lua/hunter.cmake
```diff
@@ -0,0 +1,23 @@
...
+    SHA1
+    5395f8637c14b41f950b28f2c1dba65f30fa7030
+)
+
+hunter_pick_scheme(DEFAULT url_sha1_cmake)
+hunter_download(PACKAGE_NAME Lua)
```
ruslo: "No newline at end of file" warning

cmake/projects/Lua/hunter.cmake
```diff
@@ -0,0 +1,23 @@
...
+    Lua
+    VERSION
+    "5.3.2"
+    URL
+    "https://github.com/hunter-packages/lua/archive/5.3.2.tar.gz"
+    SHA1
+    5395f8637c14b41f950b28f2c1dba65f30fa7030
+)
+
+hunter_pick_scheme(DEFAULT url_sha1_cmake)
```
ruslo: Please add `hunter_cacheable` here

cmake/projects/Lua/hunter.cmake
```diff
@@ -0,0 +1,23 @@
+if(DEFINED HUNTER_CMAKE_PROJECTS_LUA_HUNTER_CMAKE_)
```
ruslo: Feel free to leave a copyrights note

cmake/projects/Lua/hunter.cmake
```diff
@@ -0,0 +1,23 @@
...
+    PACKAGE_NAME
+    Lua
+    VERSION
+    "5.3.2"
+    URL
+    "https://github.com/hunter-packages/lua/archive/5.3.2.tar.gz"
+    SHA1
+    5395f8637c14b41f950b28f2c1dba65f30fa7030
+)
+
```
ruslo: Probably `hunter_configuration_types(Lua CONFIGURATION_TYPES Release)` need to be added here. By default debug + release variant is built but executable has same name for both (i.e. will be overwritten).

See wiki: https://github.com/ruslo/hunter/wiki/usr.adding.new.package#build-types

cmake/projects/Lua/hunter.cmake
Dinno: That's not so easy as it seems :) Many subtle details are revealed when you have to deal with multi-platform development. You make very good work. Thank you!

ruslo: > You make very good work. Thank you!

Enjoy! :)

ruslo: Please apply my Linux fixes by using this code:

```cmake
URL
"https://github.com/hunter-packages/lua/archive/v5.3.2-p0.tar.gz"
SHA1
39fc2bb5e7de0c9b4dd3940ccf99cef3b7cb4b04
```

You can leave version `5.3.2` or update it to `5.3.2-p0` to show that package is patched and not the original.

stohrendorf:
On my system, I need to additionally link to `dl` to make use of this package, because the Lua package makes use of `dlclose`.

ruslo:
> On my system, I need to additionally link to dl to make use of this package, because the Lua package makes use of dlclose

@stohrendorf How did you build it, on what system? Can you show minimal example? I've just found out that some definitions are missing in CMake build which makes Lua not use `dlclose`.

stohrendorf:
Sorry. After updating the Lua package reference to a more recent commit in by Hunter fork, the problem vanished.

ruslo:
@stohrendorf Good, thanks for the information.

Dinno:
I think, that we can merge this pull request when tests will go through.

ruslo:
> I think, that we can merge this pull request when tests will go through

Well, actually there is nothing to test with Lua so test result will not show much info :)

I've added example:
* https://github.com/ruslo/hunter/commit/c6f0556e8de8182bf3a3ef0e8ce10ffd2fa1d502

ruslo:
New wiki page: https://github.com/ruslo/hunter/wiki/pkg.lua

ruslo:
> I've added example

@Dinno Analyze toolchain failed:
```
/.../Lua/Source/src/loslib.c:274:24: warning: Access to field 'tm_sec' results in a dereference of a null pointer (loaded from variable 'stm')
    setfield(L, "sec", stm->tm_sec);
                       ^~~~~~~~~~~
```
Build log: https://travis-ci.org/ruslo/hunter/jobs/102250101
Will remove it for now: https://github.com/ruslo/hunter/commit/da1df8f9812b836281466166f02fcc3fad9a4d48

Dinno:
If stm is NULL it calls luaL_error, which in turn calls lua_error. As we can see from comment in it's code, lua_error never returns, but static analyzer does not know about that, perhaps because it is side effect of calling some kernel function behavior of which is not known to it.
```
LUA_API int lua_error (lua_State *L) {
  lua_lock(L);
  api_checknelems(L, 1);
  luaG_errormsg(L);
  /* code unreachable; will unlock when control actually leaves the kernel */
  return 0;  /* to avoid warnings */
}
```

ruslo:
> I think, that we can solve that with additional return before luaL_error

There is some macro to detect clang analyzer. You can protect additional code with it so it will be obvious why code added.

PR #284, opened 2015-12-06
Added freetype package

Dinno:


ruslo:
Few improvements:
* https://github.com/ruslo/hunter/commit/9dba9f74b976a5fc5f38aae57fc41ce9314e5931
* https://github.com/ruslo/hunter/commit/d1fb93e4180b52cd4cf760ecd600a065b64a8f3b
* https://github.com/ruslo/hunter/commit/6426a8476081bff949724c63dc52af657e142903

Simple example (only "link something" test important):
* https://github.com/ruslo/hunter/commit/20772e87052fd1bd7af420a306f424a58d8c9ad1

ruslo:
Wiki updated: https://github.com/ruslo/hunter/wiki/pkg.freetype

Dinno:
How does hunter_cacheable work? What does it cache?
And do we need to add example for every package? Or this package is chosen one?

ruslo:
> How does hunter_cacheable work? What does it cache?

It install package to some temporary folder first and save files from this folder to archive. Only then to the "global" install directory. In future if meta information about build matches then package can be reused by different `<root-id>` (i.e. when you change Hunter `SHA1`). Without it package will be rebuild. Some packages will not survive such procedure since hardcode some paths in headers/rpath/cmake-configs.

related:
* https://github.com/ruslo/hunter/wiki/usr.adding.new.package#important
* https://github.com/ruslo/hunter/wiki/dev.modules#hunter_cacheable

> And do we need to add example for every package?

It would be nice since it makes testing quite trivial:
* Manual testing https://github.com/ruslo/hunter/wiki/usr.adding.new.package#05-test
* Automatic testing with `jenkins.py` https://github.com/ruslo/hunter/wiki/dev.hunter.root#jenkinspy

> Or this package is chosen one?

No, adding example for package (if it's not exist already) depends on my mood and free time :)

Issue #12, opened 2014-10-03
Explicit in-project dependencies list


alamaison:
Is there a way to maintain the 'config.cmake' containing the versions used by a project in the project itself?

If I understand correctly, a project specifies the versions of the dependencies it needs by referring to a specific Hunter commit whose config.cmake specifies those versions.  The versions have to be locked in the Hunter, rather than in the project, because a single Hunter may be used by different projects and they share the same _Base directory when they build.  If two projects could specify different versions to be built in the same _Base, then other hunter pacakges that also needed that dependency wouldnt know which version to use.  Is that about right?

I'm wondering if there is some solution to this to make it easier to change dependency versions.  With Ruby Gemsets, for instance, all the developer has to do is change the version number and call 'bundle install'.  With Hunter if I want to, for instance, change the Boost dependency from 1.55 to 1.49, I have to check out the current hunter I'm using, update the config.cmake to point at version 1.49, check it in and push it.  Then I have to download it, calculate its new SHA1, update my project's CMakeLists.txt to point at the new hunter download location with the new SHA1.

Each project is only linked to its 'Hunterset' by that download location.  I forsee challenges managing which projects point at which Hunter downloads in the future.

Would there not be some way to have a 'canonical' Hunter repo with all the recipes to build all the dependency versions.  Then each project maintains a file (HunterSet.cmake?) that lists the versions used by that projects.  When building the dependencies, Hunter could use another level of indirection that uses the hash of all the dependencies in the HunterSet to create a new _Base directory for building.  That way, any projects sharing the same dependency versions would share a single build tree that is internally consistent.  The downside would be that any change to dependencies would start a whole new build tree.

Have I made any sense?

Issue #76, opened 2015-04-19
Add GMock


ruslo:
* https://code.google.com/p/googlemock/

Knitschi:
I included GMock in my [fork](https://github.com/Knitschi/hunter/releases/tag/0.12.33-GoogleTest) of hunter. I did it by creating my own release of the current [googletest](https://github.com/google/googletest) version. Unlike the 1.7.0 release the repository now contains GTest and GMock and a CMakeLists file that creates both projects. This makes integration with Hunter pretty straight forward. 

I also added this [FindGMock](https://github.com/triglav/cmake-findgmock) to support find_package(GMock). Integrating this into the official Hunter release should not be too much work.

ruslo:
Thanks for working on this!

> Unlike the 1.7.0 release the repository now contains GTest and GMock

*offtopic* I don't understand that move by GTest developers... Very strange, now git history of my fork is broken */offtopic*

> I also added this FindGMock to support find_package(GMock)

It will be much easier to generate that file automatically, see:
* https://github.com/ruslo/hunter/wiki/usr.adding.new.package#important
* https://github.com/ruslo/hunter/wiki/example.find_package.config

And it will produce much correct file. E.g. `FindGMock` will not work if debug library has prefix `-Debug` and not `d`. Also non-standard variant like `RelWithDebInfo` and `MinSizeRel` will not be found. + More and more issues.

If you'll apply this approach it would be awesome. Anyway feel free to send pull request.

Knitschi:
I will take a look at this and see if i can come up with something that is worth a pull request.

Knitschi:
I now created a new [hunter](https://github.com/Knitschi/hunter/releases/tag/v0.12.33-GoogleTest.2-pr) release with a new version of [GoogleTest](https://github.com/Knitschi/googletest/releases/tag/1.8.0-pr-hunter-2) to which I added the cmake code that adds config file package support. But there are still a few issues for which I am not sure if they need to be fixed.

* GTest and GMock each have an CMakeLists file. Both files now contain the same cmake package code which I more or less copied from your [example](https://github.com/forexample/package-example). This smells a little like code duplication. Maybe one could place the stuff in the parents CMakeLists file but I am not sure if this is the right way to do it.

* The cmake variables gtest_INCLUDE_DIRS and gtest_LIBRARIES are not set after calling find package. I am not sure if they should be set if one does everything right. The include directories are however set correctly.

[...]

Issue #56, opened 2015-03-23
`URL/SHA1` for `hunter_config`


ruslo:
Allow extension of space of versions by adding custom version with URL/SHA1 in `hunter_config`:

```
hunter_config(
    Foo
    VERSION 1.2.3new
    URL "https://.../1.2.3new.tar.gz"
    SHA1 "abc123..."
)
```
##

ruslo:
Allow SHA1 of commit of git repo:
```cmake
hunter_config(Foo GIT_SHA1 xxx GIT_REPO https://.../Foo.git)
```

Allow getting GIT_SHA1 from submodule:
```cmake
hunter_config(Foo GIT_SUBMODULE)
```

ruslo:
Implementation notes:
* [hunter_add_package](https://github.com/ruslo/hunter/blob/master/cmake/modules/hunter_add_package.cmake) need to detect that url and sha1 is provided by `hunter_config` and not by `hunter.cmake` and call `hunter_pick_scheme(DEFAULT url_sha1_cmake)` (supporting only CMake scheme) and `hunter_download` explicitly
* `hunter_config(... URL ... SHA1)` must set the same variables as a [hunter_add_version](https://github.com/ruslo/hunter/blob/master/cmake/modules/hunter_add_version.cmake) do, i.e. `HUNTER_<package>_URL` and `HUNTER_<package>_SHA1`
* version of the package should be equal to `SHA1` (`HUNTER_<package>_VERSION`)
* [hunter_calculate_config_sha1](https://github.com/ruslo/hunter/blob/master/cmake/modules/hunter_calculate_config_sha1.cmake) module need to find out that we are adding new package. Probably some new variable needed

Extra (related: https://github.com/ruslo/hunter/issues/138):
* script to package a project to 'tar.gz' archive, calculate SHA1 of archive and emit code for importing to config, like:
```
> hunter-dev.py --create-package /path/to/Foo
hunter_config(
    Foo
    URL /path/to/this/dir/Foo.tar.gz
    SHA1 xxxxx
)
> hunter-dev.py --create-package /path/to/Foo >> /path/to/local/config/cmake/Hunter/config.cmake
```

Note that every package creation will result new `<config-id>`, i.e. root will be rebuilt from scratch (cacheable packages will be unpacked from `Cache/raw/*.tar.gz` archive). Further improvements: https://github.com/ruslo/hunter/issues/204

andoks:
[...]

PR #295, opened 2016-01-07
External repos

mdavezac:
The objective is to allow for external repositories of recipes. In so far as I understand @alamaison's proposal in #12, this pull request goes some way towards separating two concerns, i.e. (i) the package manager, (ii) the managed recipes. It makes the following possible:

- repos for sub-communities more specialised than what the main hunter repo might cater to
- private repos for private people/organisations

Repos are specified by adding a directory to `HUNTER_RECIPE_DIRS` before any call to `hunter_add_package`:

~~~CMake
HunterGate(...)
list(APPEND HUNTER_RECIPE_DIRS "/path/to/recipes/")

project(...)
hunter_add_package(...)
~~~

Each directory has the following format, following the scheme in hunter itself:

~~~
/root/
   |--- configs/
      |--- defaults.cmake  # defaults of packages in this repo
   |--- projects/
      |--- projectA/
         |--- hunter.cmake
         |--- component0/
            |--- hunter.cmake
         |--- projectB/
            |--- hunter.cmake
~~~

For what it's worth, there is a test of this approach at UCL/GreatCMakeCookoff#40.
Yep, it should really be here, but I couldn't quite figure out how testing works in hunter...
Any pointers appreciated!


headupinclouds:
Very useful feature.

> Yep, it should really be here, but I couldn't quite figure out how testing works in hunter...
Any pointers appreciated!

If you are referring to how to run the hunter/example unit tests, I struggled with that one previously.  From a recent post by @ruslo in the #76 GMock thread, [this](https://github.com/ruslo/hunter/wiki/dev.hunter.root) wiki page discusses how to run the local hunter testing .

```
# check you have python3
> python3 --version
Python 3.4.3

# download build.py script
> git clone https://github.com/ruslo/polly
> export PATH="`pwd`/polly/bin:$PATH"
> which build.py
/.../polly/bin/build.py

# clone Hunter repository
> git clone https://github.com/ruslo/hunter
> cd hunter

# run build for some package, like GTest
# environment variable TOOLCHAIN will be used as an argument for `build.py --toolchain`
[hunter]> TOOLCHAIN=default PROJECT_DIR=examples/GTest ./jenkins.py --verbose

# script will pack current archive in `_testing` directory
-- [hunter] Initializing Hunter workspace (f0550216489047a310a27c2b5ac95c70e7e878bf)
-- [hunter]   /.../hunter/_testing/hunter.tar.gz
-- [hunter]   -> /.../_testing/Hunter/_Base/Download/Hunter/unknown/f055021
```

ruslo:
@mdavezac We will have next type of problems with such approach:

Extra recipes from `root-A` have package `Foo` with version `1.0` and SHA1 = ABC.
Extra recipes from `root-B` have package `Foo` with version `1.0` but with patch so SHA1 = CBA

The root directory will be choosed by:
* SHA1 of Hunter archive (same for both `root-A` and `root-B`) 
* SHA1 of unified `config.cmake` (since version is the same for both A and B we will have same SHA1 too!)
* SHA1 of toolchain (same too).

Hence after `root-A` is built there will be no instructions triggered for `root-B` and non-patched version will be used. Let me know if I'm missing something.

By the way here is feature request with implementation notes that can do quite the same job:
* https://github.com/ruslo/hunter/issues/56

Beautifier commit cherry picked to `develop`, thanks:
* https://github.com/ruslo/hunter/commit/049d7a21eb17bde7d891953768eec62b4b673f51

mdavezac:
If I understand @ruslo's comment, it is that with separate repos of recipes, we may have recipes over-shadowing each other if they declare the same package and the same version number. There is no one-to-one mapping between code and package name + version number. 

Frankly, in some ways this is a feature, not a bug. If the same package declares the same version number, then the user most likely expects only one package, not two. It could mean hell to try and link a program to two ever-so-slightly different versions of the same library. Which versions overshadows which is then set by ordering the directories in HUNTER_RECIPE_DIRS.

That said, I think in the short term the simplest approach is to exit with an error message if overshadowing happens, possibly to the point where a package can only be in a single repo at a time (whether or not they declare the same version). This would ensure the uniqueness of the mapping from  package+version to actual code.

Later on, we could add correct versioning to the external repos and add provenance to the information kept with a package. Then a fuse would blow only if a package+version is requested with different provenances. This would ensure that although over-shadowing is allowed, it should remain consistent for a given HUNTER_ROOT.

ruslo:
> Frankly, in some ways this is a feature, not a bug

From my point of view it's a design bug, not feature. For now there are no such "features" in Hunter and never will be. Currently everything will work correctly, both versions will live in separate `hunter-id` - hence in separate directories. In cache the SHA1 of archive will be saved (along with many other parameters) so cache will be in correct state too. Again, I think the easiest way to implement it is to follow https://github.com/ruslo/hunter/issues/56.

> It could mean hell to try and link a program to two ever-so-slightly different versions of the same library

This conflict about two independent projects with different `Foo` packages which have same version number `1.0`, it's not one project with two versions (this conflict is not possible by design).

>  If the same package declares the same version number, then the user most likely expects only one package, not two

User may not even know about this conflict. For instance there is project `A` which use patched `Boost` and version is `1.59-p0`. And there is project `B` which use different patch of `Boost` but same version `1.59-p0`. If you start with the empty Hunter directory and build `A` - everything will work fine, if you start with empty Hunter directory and build `B` - everything will work fine. But if you start with `A`, then build `B` - `B` will use patched version of `Boost` which is specified by `A`.

> I think in the short term the simplest approach is to exit with an error message if overshadowing happens

The problem is that you can't detect it?

mdavezac:
The latest commits forces the repos to be orthogonal. By this I mean that a project, as identified exclusively by its name (and not version), can only live in a single repo at a time.


I've also thrown schemes into the mix, with the ability for repos to add their own schemes. Once again, they a scheme is enforced to live in a single repo.

However, schemes that live in projects/PROJECTNAME/schemes/schemename.cmake.in take priority. I thought I saw that feature request in an issue but can't find it now. If it's not wanted, I can remove it from this pull request.

Would this be a sensible solution for now, until I can grok #56 better? (assuming the tests pass, of course).

ruslo:
> The latest commits forces the repos to be orthogonal

Doesn't solve the issue. I've created repos for testing:
* Experimental Hunter version with your changes: https://github.com/ruslo/hunter/releases/tag/v0.0.0-experimental
* Package `Foo` (which is not in Hunter) with versions `1.0-patch-A` and `1.0-patch-B`: https://github.com/forexample/temp-foo/releases
* Package `Boo` use `Foo` release `1.0-patch-A`, names it locally `1.0-p0`: https://github.com/forexample/temp-boo
* Package `Bar` use `Foo` release `1.0-patch-B`, names it locally `1.0-p0`: https://github.com/forexample/temp-bar
* Testing script: https://gist.github.com/ruslo/f5c6880462230c8d4dd6

```bash
> mkdir hunter-test
> cd hunter-test
[hunter-test]>  wget https://gist.githubusercontent.com/ruslo/f5c6880462230c8d4dd6/raw/8a870041499f37169df13384bdedc37e8d64e3b5/run-test.sh
[hunter-test]> chmod +x run-test.sh
[hunter-test> ./run-test.sh
```

Result:
```
========== Generate Boo
========== Build Boo
========== Test clean Boo
...
test 1
    Start 1: BooTest

1: Test command: /.../_builds/boo
1: Test timeout computed to be: 9.99988e+06
1: Foo say: A
```
If Hunter directory is clean, then Boo start build and correct version of Foo used, prints `A`

Same for Bar (Hunter directory cleaned before build):
```
========== Generate Bar
========== Build Bar
========== Test clean Bar
...
test 1
    Start 1: BooTest

1: Test command: /.../_builds/boo
1: Test timeout computed to be: 9.99988e+06
1: Foo say: B
```

But if we will not clean-up after building `Bar` and start `Boo` again, wrong version of `Foo` will be used:
```
========== Generate Boo
========== Build Boo
========== Test dirty Boo
...
test 1
    Start 1: BooTest

1: Test command: /home/ruslan/work/test/hunter-test/_builds/boo
1: Test timeout computed to be: 9.99988e+06
1: Foo say: B
```
As you can see it picks version of `Foo` which installation is triggered by `Bar`.

> I've also thrown schemes into the mix, with the ability for repos to add their own schemes.

This will add additional moves to implementation, just for your information #56 limited only for CMake standard schemes.

mdavezac:
Okay, thanks for laying out for me. It's very helpful. Now, just have to figure it out.

ruslo:
Closing. For better approach and implementation details see https://github.com/ruslo/hunter/issues/56

PR #297, opened 2016-01-07
Changed source of package to patch project

Dinno:


cmake/projects/freetype/hunter.cmake
```diff
@@ -15,9 +15,9 @@ hunter_add_version(
     VERSION
     "2.6.2"
     URL
-    "http://sourceforge.net/projects/freetype/files/freetype2/2.6.2/freetype-2.6.2.tar.bz2/download"
+    "https://github.com/hunter-packages/freetype/archive/hunter.tar.gz"
```
ruslo: This file will change from time to time. For example if I add some commit to branch `hunter` then SHA1 of archive will change => download will fail with hash mismatch. You need to create release and add stable/release archive here.

Dinno: I will do it. But i want to fix the problem with test build first.

Dinno:
A fixed build and changed source of package to tag v2.6.2-p0

ruslo:
Merged. Please set destination branch `develop` next time, see https://github.com/ruslo/hunter/wiki/dev.contribution. Thanks!

ruslo:
https://github.com/ruslo/hunter/releases/tag/v0.12.34

PR #302, opened 2016-01-14
Added package odb-pgsql

pretyman:
This package depends on package PostgreSQL, since it is odb support for the PostgreSQL database


cmake/projects/odb-pgsql/hunter.cmake
```diff
@@ -0,0 +1,33 @@
...
+else()
+  set(HUNTER_CMAKE_PROJECTS_ODB-PGSQL_HUNTER_CMAKE 1)
+endif()
+
+include(hunter_add_version)
+include(hunter_download)
+include(hunter_pick_scheme)
+
+hunter_add_package(odb)
+hunter_add_package(PostgreSQL)
```
ruslo: Not needed here since we have them in scheme already

cmake/projects/odb-pgsql/hunter.cmake
```diff
@@ -0,0 +1,33 @@
...
+    URL
+    "http://www.codesynthesis.com/download/odb/2.4/libodb-pgsql-2.4.0.tar.bz2"
+    SHA1
+    4628d5e296da01dbaf8658fd402b4f709f30ea2d
+)
+
+hunter_pick_scheme(DEFAULT url_sha1_odb-pgsql_autotools)
+hunter_download(
+    PACKAGE_NAME odb-pgsql
+    PACKAGE_DEPENDS_ON odb PostgreSQL
```
ruslo: Not needed

cmake/projects/odb-pgsql/hunter.cmake
pretyman: Ok, I used the `odb-sqlite` template, where in the past we've decided to put dependencies in `projects/../hunter.cmake` and on the scheme script. I'll remove these.

cmake/projects/odb-pgsql/hunter.cmake
pretyman: This is from `odb-sqlite` template, where in the past we've decided to use `PACKAGE_DEPENDS_ON`.

pretyman:
I'll open a new pull request with your suggested changes for a clean history

PR #304, opened 2016-01-14
Added package odb-pgsql

pretyman:
This package depends on package PostgreSQL, since it is odb support for the PostgreSQL database


pretyman:
I must stop working at 5AM, I made the changes you suggested, but I managed to create a pull request with the same old, unchanged, code...

PR #314, opened 2016-01-29
Added package MySQL-client

pretyman:
Hi Ruslan,

I named this package MySQL-client and not only MySQL to leave the door open to whoever wants to contribute MySQL the embedded server library in the future.

Note that in the MySQL website they call it "MySQL connector", and the library goes by the name of `libmysqlclient`.

Just let me know if you would like different names for the Hunter package and CMake Find script. 

Note that MySQL does not install a Find or Config scripts.


pretyman:
There is a C++ connector as well: https://dev.mysql.com/downloads/connector/cpp/ so I guess I should somehow reflect that on the name, what do you think?

ruslo:
Test works for me.

@pretyman Should I merge that or you want to make some updates?

pretyman:
I think its best not to merge. I will update the CMake scripts. Would be great if anyone with experience with MySQL could help to give names for this package :-)
I'll sleep over it and update it in the morning. 
Thanks!

ruslo:
> Would be great if anyone with experience with MySQL could help to give names for this package

@pretyman I see no standard FindMySQL module so I guess you can name it as you like :)

PR #315, opened 2016-01-30
Added package MySQL-client

pretyman:
Left name as MySQL-client for the MySQL C Connector because of the name of library file.

MySQL provides a C++ Connector which has the `libcppcon` library file name, which depends on this C connector. If anyone would add the C++ connector in the future, I guess they could name the Hunter package `MySQL-cppcon` and create a target name `MySQL::cppcon` because this seems to be the vocabulary used for the library filename in the build instructions for the C++ Connector.


pretyman:
Opened this pull request towards the wrong branch. Sorry!

PR #316, opened 2016-01-30
Added package MySQL-client

pretyman:
Left name as MySQL-client for the MySQL C Connector because of the name of library file.

MySQL provides a C++ Connector which has the `libcppcon` library file name, which depends on this C connector. If anyone would add the C++ connector in the future, I guess they could name the Hunter package `MySQL-cppcon` and create a target name `MySQL::cppcon` because this seems to be the vocabulary used for the library filename in the build instructions for the C++ Connector.

PR #317, opened 2016-01-30
Added package odb-mysql

pretyman:
This package depends on `MySQL-client`, so that should be merged first.

No need to add page on the wiki for this package, I shall rename the wikipage of `odb-pgsql` to `odb` and treat all odb package information in there together with instructions to use the `odb-compiler` package. I'll get around to do that after I finish my university term.


ruslo:
> should be merged first

Ok, will do after testing: https://github.com/ruslo/hunter/pull/316#issuecomment-177125622

ruslo:
Just for the record: depends on https://github.com/ruslo/hunter/pull/316

pretyman:
Not sure how is this package going to be tested, but could we use the same configuration from https://github.com/ruslo/hunter/pull/316 ? Do you think of anyway to inherit the testing configuration from the dependencies of a package?

ruslo:
> Not sure how is this package going to be tested

Same way. Separate branch with example.

> but could we use the same configuration from #316

Actually we have to exclude broken toolchains since it's a dependency.

> Do you think of anyway to inherit the testing configuration from the dependencies of a package?

Not sure I understand what you mean by inheriting. There will be separate branch with `PROJECT_DIR=examples/odb-mysql` in .travis.yml/appveyor.yml. Dependency will be downloaded automatically.

PS I see some files from merged pull request. Can you please do rebase? Thanks.

pretyman:
> Not sure I understand what you mean by inheriting.

Sorry, it is that I don't fully understand how the testing works in Appveyor. My question is: since this package depends on MySQL-client, and MySQL-client failed some tests, would the testing of this package try to compile the MySQL-client dependency and fail the same tests? Or it would use the cache to get MySQL-client and by doing this not failing the MySQL-client tets? I'm guessing it would use the cache, but I couldn't find an example to be sure.

> PS I see some files from merged pull request. Can you please do rebase? Thanks.

Done. No, thank you :+1:

ruslo:
Testing now:
* https://github.com/ingenue/hunter/branches/all?utf8=%E2%9C%93&query=pkg.odb.mysql

pretyman:
Linux test failed because the compiler being downloaded is i686 and not x64, I'll fix that.

Windows tests under VS should fail because autotools integration is not supported under Visual Studio. Mingw failed, but I have no idea why. I never used `odb` on Windows, so I don't know how to integrate it better into hunter.

PR #323, opened 2016-02-07
Fix OpenSSL builds with MSVC.

alamaison:
OpenSSL is doind clang-cl detection by calling `cl.exe` with certain
arguments that aren't valid for MS cl.exe.  I'm not sure when this
behaviour started.  Just as happens with Boost, the instance of msvc
that runs the ExternalProject defines VS_UNICODE_OUTPUT to redirect all
output to a pipe.  That means is sees these non-fatal errors that are
happening all the way down in the OpenSSL configure step, and thinks the
build has failed.

This change works around the problem in the same way as the Boost scheme
by setting LOG_CONFIGURE, which wraps the ExternalProject build in a
script that insulates it from msvc's environment variables.

### DIFF
--- a/cmake/schemes/url_sha1_openssl_windows.cmake.in
+++ b/cmake/schemes/url_sha1_openssl_windows.cmake.in
@@ -64,6 +64,14 @@ else()
   set(do_ms do_ms.bat)
 endif()
 
+if("@MSVC@")
+  # Logging as Workaround for VS_UNICODE_OUTPUT issue:
+  # https://public.kitware.com/Bug/view.php?id=14266
+  set(log_opts LOG_CONFIGURE 1)
+else()
+  set(log_opts "")
+endif()
+
 ExternalProject_Add(
     "@HUNTER_EP_NAME@"
     URL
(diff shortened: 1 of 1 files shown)


ruslo:
Thanks, I've started OpenSSL testing of this PR here: https://ci.appveyor.com/project/ingenue/hunter/build/1.0.103

> OpenSSL is doind clang-cl detection by calling cl.exe with certain
arguments that aren't valid for MS cl.exe

Do you have an idea how it can be tested on AppVeyor?

alamaison:
> Do you have an idea how it can be tested on AppVeyor?

Not really.  I'm not sure if VS_UNICODE_OUTPUT is just a VS2005 thing

ruslo:
> Thanks, I've started OpenSSL testing of this PR here: https://ci.appveyor.com/project/ingenue/hunter/build/1.0.103

Passed. Merging.

Thanks, Alex!

ruslo:
Release: https://github.com/ruslo/hunter/releases/tag/v0.12.43

PR #324, opened 2016-02-07
Fix wxWidgets and Winsparkle for current Hunter.

alamaison:


cmake/projects/WinSparkle/hunter.cmake
```diff
@@ -17,7 +17,7 @@ hunter_add_version(
     VERSION
     "0.4.0"
     URL
-    "https://github.com/alamaison/winsparkle/archive/v0.4-hunter-3.tar.gz"
+    "https://github.com/alamaison/winsparkle/archive/v0.4-hunter-4.tar.gz"
     SHA1
     8e7790c93fe7ffeae529621a7fc93ea10e9d2af6
```
ruslo: I think SHA1 must change (?)

alamaison:
Yes, of course.  Thanks for spotting.

ruslo:
Thanks, merged. Looks like there is no tests for these packages: https://github.com/ingenue/hunter/branches/all feel free to add.

PR #329, opened 2016-02-21
Add WTL (Windows Template Library).

alamaison:


ruslo:
New wiki page: https://github.com/ruslo/hunter/wiki/pkg.wtl

PR #330, opened 2016-02-21
Add Boost.Process.

alamaison:
Like Boost.Compute, Boost.Process is not (yet) an official Boost library.

I must be doing something stupid because when I try to build the included example, I get errors like:
```
CMake Error at CMakeLists.txt:18 (add_executable):
  Target "foo" links to target "Boost::system" but the target was not found.
  Perhaps a find_package() call is missing for an IMPORTED target, or an
  ALIAS target is missing?
```

Boost.Process depends on system, filesystem and iostreams but, becuse the example uses `find_package(BoostProcess CONFIG REQUIRED)` I thought I didn't have to repeat the dependency and it would get picked up from BoostProcess automatically.  Is that not the case?


ruslo:
> I thought I didn't have to repeat the dependency and it would get picked up from BoostProcess automatically

Yes user don't have to do it - it's upstream responsibility. In your case since boost.process use boost.system you should add `find_package(Boost COMPONENTS system)` to `BoostProcessConfig.cmake` generated file.

If you take a look at this template project: https://github.com/forexample/package-example then `find_package` should be placed to [Config.cmake.in](https://github.com/forexample/package-example/blob/cf2ea1d6a209fb9eca2ab83fdd0ac15fe4d3e807/Foo/cmake/Config.cmake.in).

Note that usually you should add `find_dependency` but I think this command doesn't support MODULE/CONFIG switch and COMPONENTS so you can use just `find_package`.

alamaison:
Of course.  Thanks.  I always forget that bit.  I guess I find it odd that cmake can collect all those transitive usage requirements into a targets file, but not the find_package dependencies.  Oh well.

I've updated the PR

ruslo:
New wiki page: https://github.com/ruslo/hunter/wiki/pkg.boost.process

Issue #97, opened 2015-04-23
Add protobuf (google protocol buffers)


ruslo:
* [Official](https://github.com/google/protobuf) non-cmake

For package: https://github.com/ruslo/hunter/issues/79

KerstinKeller:
In the meantime, protobuf seems to support cmake for Version 3. It is still in beta stage, but it might be worth to have another look at the package.
I would be very interested in hunter supporting protobuf.

ruslo:
> protobuf seems to support cmake for Version 3

I've just noticed that there is a `cmake` directory :) Probably we just need to add a top `CMakeLists.txt` with one command `add_subdirectory(cmake)` and everything will just work.

@KerstinKeller Can you test this solution? Hints: https://github.com/ruslo/hunter/wiki/usr.adding.new.package

KerstinKeller:
Yes, I will have a look. It seems not quite as easy though (at least for
VS2012):
To compile protobuf, you're required to manually check out goolemock/gmock
into the main directory. There is however a realease tar.gz which contains
this folder already, but I cannot compile because of the
"_VARIADIC_TEMPLATE_MAX" default 5 in VS2012 .

I also tried with VS2015 Community (which seems not to be supported by
Hunter yet - probably cmake\modules\hunter_setup_msvc_vcvarsall.cmake will
have to be adapted).

I will  investigate further and update the issue accordingly.

2015-09-30 12:02 GMT+02:00 Ruslan Baratov <notifications@github.com>:

ruslo:
VS2015 should work fine with Hunter. From what I've remember while fixing this bug: https://github.com/forexample/hunter-simple/issues/2

ruslo:
> To compile protobuf, you're required to manually check out goolemock/gmock into the main directory

@KerstinKeller I think `BUILD_TESTING=OFF` might be helpful here?

KerstinKeller:
Ok, I was able to successfully install with hunter and `BUILD_TESTING=OFF`. I changed this in the protobuf cmake. Is there also an option to pass this from protobuf/hunter.cmake file?

And what is the general procedure to make this available then? Need to fork the protobuf project to integrate the changes (mainly additional CmakeLists.txt with `add_subdirectory(cmake)`  in the main directory)?

ruslo:
[...]

PR #331, opened 2016-02-22
Add Google Protocol Buffer

tatraian:
I have added Google's Protocol Buffer to hunter.

There is some potential problem with it:
First, the protocol buffer itself doesn't compile under linux, with c++11/clang/libc++. I opened an issue and sent a fix for it, but no answer yet.

Second. ProtBuf has cmake based build system, but it is placed into the cmake subdirectory and I used the autotool based builder instead.


ruslo:
> ProtBuf has cmake based build system, but it is placed into the cmake subdirectory and I used the autotool based builder instead

It means that windows will not be supported and `*Config.cmake` files will not be installed effectively making this package non-cacheable and non-relocatable. There is a discussion already: https://github.com/ruslo/hunter/issues/97 and fork for patches: https://github.com/hunter-packages/protobuf

> I opened an issue and sent a fix for it, but no answer yet

You can use hunterized fork if it make a fix in fact.

Also please use `develop` branch for pull request, see:
* https://github.com/ruslo/hunter/wiki/dev.contribution
* http://0.readthedocs.org/en/latest/git.html#pick-target-branch

tatraian:
Sorry, I forgot read your dev.contribution wiki... Next time I will use the development branch...

We can close this request if you have an issue for this as well.

In google/protobuf#1266 you can see the compile error fix's life. I think it is better to get it from the original package (if they accept it).

ruslo:
> I think it is better to get it from the original package (if they accept it)

Of course it's better to pick fixes from original, I'm just saying that hunterized fork can be used for fast experimenting/fixes.

> We can close this request if you have an issue for this as well

Okay, closing. Feel free to continue discussion in https://github.com/ruslo/hunter/issues/97

PR #333, opened 2016-02-28
Add latest release of libssh2 - 1.7.0.

alamaison:


ruslo:
All windows tests fail: https://ci.appveyor.com/project/ingenue/hunter/build/1.0.134

alamaison:
OpenSSL on Windows depends on crypt32.lib but FindOpenSSL.cmake doesn't expose that dependency.  Therefore packages that link statically with openssl may have link errors.  I've fixed this now in libssh2 by explicitly linking that library.

ruslo:
> I've fixed this now in libssh2 by explicitly linking that library

Alternatively you can patch `FindOpenSSL.cmake` module. It's a part of Hunter:
* https://github.com/ruslo/hunter/blob/master/cmake/find/FindOpenSSL.cmake

Also there is a chance that this module is out-of-date and contains some bugs like you've described.

ruslo:
Testing:
* https://travis-ci.org/ingenue/hunter/builds/117109150
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.297

alamaison:
> Alternatively you can patch FindOpenSSL.cmake module. It's a part of Hunter:

I thought about that but decided not to because the change had to be made in upstream libssh2 anyway for non-Hunter use.

> Also there is a chance that this module is out-of-date and contains some bugs like you've described.

True, but this particular bug isn't fixed in the latest CMake version either.  Thought the latest CMake version does have new stuff like an import library for OpenSSL.  What is the reason for keeping a custom FindOpenSSL.cmake?  Is it to lock it down so it only ever searches the Hunter dirs?

ruslo:
> Thought the latest CMake version does have new stuff like an import library for OpenSSL

Actually it is a nice improvement. Probably it will make OpenSSL cacheable/uploadable. Feel free to send PR with update, will see how it works.

>  What is the reason for keeping a custom FindOpenSSL.cmake?

Standard version doesn't respect `OPENSSL_ROOT`. And since the minimum version for Hunter is `3.0` for now it should be shipped with Hunter anyway, even if it will be fixed upstream.

ruslo:
> Testing

All tests passed. Merging.

ruslo:
Release: https://github.com/ruslo/hunter/releases/tag/v0.14.3

Issue #310, opened 2016-01-19
Add ZeroMQ package


LegalizeAdulthood:
http://zeromq.org/

There appears to be multiple C++ bindings for ZeroMQ:

http://zeromq.org/bindings:cpp

ruslo:
Added by @tatraian

PR #334, opened 2016-03-01
ZeroMQ library

tatraian:
I created a package for ZeroMQ. ZeroMQ has cmake based build BUT it doesn't generate ProjectConfig.cmake, but generates pkgconfig. I used the last since it is implemented:) and I'm lame for creating ProjectConfig.cmake...

I also added an example as well. As I see there is an issue #310 also wishes this package.

### DIFF
--- a/cmake/configs/default.cmake
+++ b/cmake/configs/default.cmake
@@ -108,7 +108,6 @@ hunter_config(spdlog VERSION 1.0.0-p0)
 hunter_config(szip VERSION 2.1.0-p1)
 hunter_config(wxWidgets VERSION 3.0.2)
 hunter_config(ZeroMQ VERSION 4.1.4)
-hunter_config(zmqpp VERSION 4.1.2)
 
 if(MINGW OR MSVC80)
   hunter_config(GTest VERSION 1.7.0-hunter-6)
--- a/cmake/configs/default.cmake
+++ b/cmake/configs/default.cmake
@@ -107,6 +107,8 @@ hunter_config(sparsehash VERSION 2.0.2)
 hunter_config(spdlog VERSION 1.0.0-p0)
 hunter_config(szip VERSION 2.1.0-p1)
 hunter_config(wxWidgets VERSION 3.0.2)
+hunter_config(ZeroMQ VERSION 4.1.4)
+hunter_config(zmqpp VERSION 4.1.2)
 
 if(MINGW OR MSVC80)
   hunter_config(GTest VERSION 1.7.0-hunter-6)
--- a/examples/ZeroMQ/CMakeLists.txt
+++ b/examples/ZeroMQ/CMakeLists.txt
@@ -1,0 +1,20 @@
+cmake_minimum_required(VERSION 3.2)
+
+# Emulate HunterGate:
+# * https://github.com/hunter-packages/gate
+include("../common.cmake")
+
+project(download-ZeroMQ)
+
+hunter_add_package(ZeroMQ)
+
+find_package(PkgConfig)
+set(PKG_CONFIG_USE_CMAKE_PREFIX_PATH ON)
+pkg_search_module(ZeroMQ REQUIRED libzmq)
+
+include_directories(${ZeroMQ_INCLUDE_DIRS})
+link_directories(${ZeroMQ_LIBRARY_DIRS})
+
+
+add_executable(version main.cpp)
+target_link_libraries(version ${ZeroMQ_LIBRARIES})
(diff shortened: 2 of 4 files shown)


LegalizeAdulthood:
Are you building only the C API or the C++ API as well?  See #310

tatraian:
No, this is the pure ZeroMQ.

C++ interfaces:
cppzmq: As I see this is the C++ interface that is presented in examples. This is a header only library and can be add with file copy to ZeroMQ. (But --- this is my opinion --- it does not give much more then the C api)
zmqpp: Is a much thicker C++ API (we plan to use this), but its CMake scripts not looks like cross platform (it gives gcc compatible parameters with platform conditions). Furthermore it doesn't create package so after building you have to search the header and libraries with find_file and find_library. I don't know what is the policy of hunter about these issues. Since we plan to use this library, I made a fork ([tatraian/zmqpp](https://github.com/tatraian/zmqpp) )  from it and currently I write the modifications of CMake file.
azmq: This is the marriage of Boost's Asio and ZeroMQ. It is also a header only library with CMake scripts, but it doesn't create package.
CpperoMQ: This is also a header only library with additional features without any build scripts.

Since you can choose among several libraries, I planed to add zmqpp as a separated project.

cmake/configs/default.cmake
```diff
@@ -107,6 +107,8 @@ hunter_config(sparsehash VERSION 2.0.2)
 hunter_config(spdlog VERSION 1.0.0-p0)
 hunter_config(szip VERSION 2.1.0-p1)
 hunter_config(wxWidgets VERSION 3.0.2)
+hunter_config(ZeroMQ VERSION 4.1.4)
+hunter_config(zmqpp VERSION 4.1.2)
```
ruslo: See no `zmqpp` package in this patch

examples/ZeroMQ/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+# * https://github.com/hunter-packages/gate
+include("../common.cmake")
+
+project(download-ZeroMQ)
+
+hunter_add_package(ZeroMQ)
+
+find_package(PkgConfig)
+set(PKG_CONFIG_USE_CMAKE_PREFIX_PATH ON)
+pkg_search_module(ZeroMQ REQUIRED libzmq)
```
ruslo: It will not work for Windows, right? If yes then it's better to add `hunter_report_broken_package` to the `hunter.cmake` so the error reported before build started (and message will be more clearer). Like this https://github.com/ruslo/hunter/blob/c006432102cf21bf5171d38fd0a8fc7c4a3a183f/cmake/projects/BZip2/hunter.cmake#L18

tatraian:
Sorry Ruslo! That was a bug. I made the two library together... I cleared it out from defult.cmake now.

ruslo:
> I used the last since it is implemented:) and I'm lame for creating ProjectConfig.cmake

It's not the best option but if it will work I will merge it. Though it's better to add this functionality since it will be easier for users to use `find_package(... CONFIG)`, package will be cacheable and dependent packages can be cacheable too.

If you (or someone else) want to update this in future here are wiki page with hints:
* https://github.com/ruslo/hunter/wiki/example.find_package.config

ruslo:
>  I made a fork

There is a central place for experimental forks for hunterizing: https://github.com/hunter-packages I can create new repository if you want and give you rights to write.

ruslo:
Testing now: https://github.com/ingenue/hunter/branches/all?utf8=%E2%9C%93&query=pkg.zeromq.pr-334

tatraian:
Ok. Than create a zeromq fork please and then I try to patch its CMakeLists.txt file...

tatraian:
ZeroMQ have different repositories for releases (so the development repository is absolutely independent of releases). Please clone the zeromq4-1 repository.

ruslo:
> Please clone the zeromq4-1 repository

Here is the fork: https://github.com/hunter-packages/zeromq4-1

> give you rights to write

@tatraian Check your email

ruslo:
> Testing now: https://github.com/ingenue/hunter/branches/all?utf8=%E2%9C%93&query=pkg.zeromq.pr-334

Some linux tests passed, merging

ruslo:
All Windows tests failed: https://ci.appveyor.com/project/ingenue/hunter/build/1.0.146
All OSX tests failed: https://travis-ci.org/ingenue/hunter/builds/113099814
Android and analyze toolchains failed on Linux:
* https://travis-ci.org/ingenue/hunter/jobs/113099819
* https://travis-ci.org/ingenue/hunter/jobs/113099820

ruslo:
New wiki page: https://github.com/ruslo/hunter/wiki/pkg.zeromq

PR #337, opened 2016-03-03
ZeroMQ now can be used as CMake package

tatraian:
I updated the hunter scripts and patched libzmq, so now I hope we can use it with find_package CMake command.

### DIFF
--- a/cmake/projects/ZeroMQ/hunter.cmake
+++ b/cmake/projects/ZeroMQ/hunter.cmake
@@ -14,9 +14,9 @@ hunter_add_version(
   VERSION
   "4.1.4"
   URL
-  "https://github.com/zeromq/zeromq4-1/archive/v4.1.4.tar.gz"
+  "https://github.com/hunter-packages/zeromq4-1/archive/v4.1.4-hunter.tar.gz"
   SHA1
-  840b24394e68450f46178c16878b0f9dc163e406  
+  aeb0650b1d1b2e5f7c9da81d4e294716eb16e91e
   )
 
 hunter_pick_scheme(DEFAULT url_sha1_cmake) 
--- a/cmake/projects/ZeroMQ/hunter.cmake
+++ b/cmake/projects/ZeroMQ/hunter.cmake
@@ -14,9 +14,9 @@ hunter_add_version(
   VERSION
   "4.1.4"
   URL
-  "https://github.com/hunter-packages/zeromq4-1/archive/v4.1.4-hunter.tar.gz"
+  "https://github.com/hunter-packages/zeromq4-1/archive/v4.1.4-hunter-p1.tar.gz"
   SHA1
-  aeb0650b1d1b2e5f7c9da81d4e294716eb16e91e
+  be57db45f4f998c04d1804132b1302f923405fa4
   )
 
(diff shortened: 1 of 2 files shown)


ruslo:
Thanks, I've enables all CI matrix back to see if it fix some failures. Testing now:
* https://github.com/ingenue/hunter/branches/all?utf8=%E2%9C%93&query=pkg.zeromq.pr-337

tatraian:
It's not looking good. As I see linux builds failed because statically linked version doesn't find librt. That was not problem on my ubuntu box... I will fix this tomorrow.

Looks like ZeroMQ cannot be compiled on android. I think it is independent from me and it is better to extract mobile platforms from CI matrix...

On Windows I don't know what is the problem currently, and I don't have Windows environment. But this must be built on Windows. I hope in the next week I have VS2015 and then I try to figure out what is the problem.

ruslo:
> As I see linux builds failed because statically linked version doesn't find librt

Not "find" but "link". I think if you add `rt` as a usage requirement then it will work, i.e.
```cmake
if(UNIX AND NOT APPLE)
    target_link_libraries(libzmq PUBLIC rt)
endif()

tatraian:
The library itself was correctly linked, so there is no problem with libzmq. But I added two example applications. The first is dynamically linked to libzmq and the second is statically linked to libzmq-static. The compile error happened at the second. Now I add a huge amount of code to the example CMakeLists.txt to find and link the same libraries to the static version that are linked to libzmq.so as well. But if you prefer I can eliminate the static version.

ruslo:
> The library itself was correctly linked, so there is no problem with libzmq

The logic is next: if `main.cpp` of example is not using any function from `rt` library then the reason of link errors is that `libzmq` use something from `rt` => must provide `rt` as a usage requirement => `CMakeLists.txt` of `libzmq` should be patched.

I guess static version of `libzmq` installed fine because it's okay to have unresolved symbols for library, the real error occurs when you try to link everything into final executable.

Dynamic version of `libzmq` works fine because dynamic libraries contain info about dependency (run `ldd libzmq.so`), so it can be loaded automatically.

tatraian:
Ok. Then I don't understand (know) the next: When you create a static library (libzmq-static) then it is just an archive of compiled objects. In this case does the archive itself contain any information about libraries that required for final linking (As I know it doesn't)? And If the static archive does not give you this dependency, then can CMake provide it (for ZeroMQ::libzmq-static target)? I proposed not. So in the example I searched all libraries that were linked to libzmq.so and I linked them to statically linked example binary.

tatraian:
Sorry. I think I understood... In case of static libraries the `target_link_libraries` command does not make real linking but the target will know its dependencies. In the deep that means that the CMake will generate `set_target_properties(IMPORTED_LINK_INTERFACE_LIBRARIES_* ...)` property for the target in the ProjectTarget-*.cmake files.

Please restart the test on this version. On Windows it will fail, but I hope I fixed the compile error on linux.

ruslo:
>Please restart the test on this version

Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.151
* https://travis-ci.org/ingenue/hunter/builds/113713499

ruslo:
> Testing

Much better! Now only android, ios and analyze fail.

PR #340, opened 2016-03-05
Trial version for Google Protocol Buffer

tatraian:
This is a simple example and project file for protobuf testing.
It refers to changes of: hunter-packages/protobuf#1


tatraian:
A known bug: protobuf cannot be compiled with clang/libc++. I wrote a patch for this google/protobuf#1299.
I will backported to hunterized branch as well.

examples/Protobuf/ProtoBuffer.diff
```diff
@@ -0,0 +1,12 @@
+diff --git a/cmake/configs/default.cmake b/cmake/configs/default.cmake
```
ruslo: I think this file was added by mistake

tatraian: Yes. Sorry. I removed it.

ruslo:
> Yes. Sorry. I removed it.

Thanks, looks good now. I'll wait for merging https://github.com/hunter-packages/protobuf/pull/1 first.

tatraian:
I updated the hash and url. I made Protobuf cacheable and I moved the cmake args from config/default.cmake to Protobuf's hunter.cmake.

ruslo:
Thanks, testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.162
* https://travis-ci.org/ingenue/hunter/builds/114567693

ruslo:
> Thanks, testing

All windows tests fails. According to the logs somebody use `MD` instead `MT` (or vice versa). Effectively violating [respect defaults](https://github.com/ruslo/hunter/wiki/usr.adding.new.package#default-behaviour) rule. MinGW build failed because of timeout. Is there a way to reduce number of thing to build? May be removing examples or other stuff?

Travis CI is much better. iOS, Android and analyze toolchains fail.

ruslo:
New wiki page: https://github.com/ruslo/hunter/wiki/pkg.protobuf

PR #342, opened 2016-03-07
odb-compiler linux: changed from i686 to x64

pretyman:


cmake/projects/odb-compiler/hunter.cmake
```diff
@@ -29,9 +29,9 @@ elseif(CMAKE_HOST_UNIX)
       VERSION
       "2.4.0"
       URL
-      "http://www.codesynthesis.com/download/odb/2.4/odb-2.4.0-i686-linux-gnu.tar.bz2"
```
ruslo: Probably one more condition should be used here like
```cmake
if(CMAKE_HOST_UNIX)
  if(ARCHITECTURE_I686) # ???
    set(... ...-i686-...)
  else()
    set(... ...-x86_64-...)
  endif()
endif()
```

pretyman: where is `ARCHITECTURE_I686` set? I cannot find it in the docs nor in the source code (I jut grep'ed for it, though, didn't look any deeper) nor on Google.

On the other hand, I naively assume that most systems are 64bits nowadays, I just put the wrong compiler by mistake in the past because I wasn't compiling on Linux at the time.

ruslo: > where is ARCHITECTURE_I686 set?

Nowhere, it's just first thing cames to mind. That's why I placed `???` near usage. Sorry for confusion.

I'm not 100% sure but you can use `CMAKE_SIZEOF_VOID_P`:
* http://stackoverflow.com/a/4029607/2288008

ruslo:
> should I have double quotes around ${odb_linux_compiler_url} ?

Not necessary but in my opinion it's a good habit:
* http://0.readthedocs.org/en/latest/cmake.html#enclose-path-with-double-quotes

pretyman:
Ok, made the changes. One test failed on Travis but I believe it is not related to the commit:

`The command "wget "https://github.com/ruslo/polly/archive/v${POLLY_VERSION}.tar.gz"" failed and exited with 4 during .`

cmake/projects/odb-compiler/hunter.cmake
pretyman: I guess this would be the responsibility of the toolchain, no?

ruslo: > I guess this would be the responsibility of the toolchain, no?

Yes. As far as I understand value of `CMAKE_SIZEOF_VOID_P` will depend on toolchain you're using since it determined by `try_compile`. Or you mean we need host architecture?

ruslo:
> One test failed on Travis but I believe it is not related to the commit

Yes, it's not related. I restarted this job.

cmake/projects/odb-compiler/hunter.cmake
pretyman: I would guess that's dependent on the host architecture, you are right. I don't know how to achieve that since the `CMAKE_SIZEOF_VOID_P` is dependent on the target as far as I understood it.

pretyman:
Cool, done, can we merge? :-)

I have more stuff to merge after odb-mysql, but need to fix that first ;-)

ruslo:
> can we merge

Please remove `ARCHITECTURE_I686` since there is no such variable.

cmake/projects/odb-compiler/hunter.cmake
ruslo: > is dependent on the target as far as I understood it

Okay, then `CMAKE_SIZEOF_VOID_P` should not be used here. I don't know what is the best way to solve it. I've searched a little bit and found [CMAKE_HOST_SYSTEM_PROCESSOR](https://cmake.org/cmake/help/latest/variable/CMAKE_HOST_SYSTEM_PROCESSOR.html) variable, it return `x86_64` on my Linux.

One more note is that using `url_sha1_unpack_install` scheme is not quite correct since it mix host and target executables. I haven't designed this part in Hunter for now, planning to make some improvements in future.

pretyman: I have been cross compiling `odb` to the Raspberry Pi with no problems with the current design, so you can keep this mind

pretyman:
> Please remove ARCHITECTURE_I686

Ruslaaaaaan, you make me crazy!! :laughing: But yeah, done! :+1:

cmake/projects/odb-compiler/hunter.cmake
ruslo: > to the Raspberry Pi with no problems with the current design

This is because there are no conflicts yet. I'm talking that in general such approach will not work correctly. For example you have a tool `mytool` which should be run on Linux host which depends on `libz.so` shared library. And you want to use shared zlib in your Android project which is also will be installed with name `libz.so` to the same location. Solution is to have separate host and target install directories.

ruslo:
> you make me crazy

Sorry for confusion but I can't allow commit variable without meaning and name that I've invented myself :)

ruslo:
Testing:
* https://github.com/ingenue/hunter/branches/all?utf8=%E2%9C%93&query=pr-342

Issue #338, opened 2016-03-03
Add ZMQPP package


tatraian:
ZMQPP package is a C++ front-end for ZeroMQ. Please create a hunter-packages repository for it
(https://github.com/zeromq/zmqpp) and I commit the changes that required for hunter compiling. After I will come with hunter pull request.
I don't have Windows compiler currently I could test it only on Linux.
##

ruslo:
> Please create a hunter-packages repository

Done: https://github.com/hunter-packages/zmqpp

Issue #322, opened 2016-02-05
Apply CMAKE_POSITION_INDEPENDENT_CODE-like variables to custom schemes


ruslo:
Such variables like `CMAKE_POSITION_INDEPENDENT_CODE` doesn't change `CMAKE_CXX_FLAGS` so will not affect non-CMake custom schemes. Need to introduce converter module which will read such variables and modify `CMAKE_CXX_FLAGS` accordingly.

Workaround:
- Set `-fPIC` explicitly in toolchain
##

ruslo:
\+  `CMAKE_CXX_STANDARD`

ruslo:
Implemented here:
* https://github.com/ruslo/hunter/blob/64ce65719d28cc841d834cd5b813038effc1b128/cmake/modules/hunter_dump_cmake_flags.cmake#L124-L138

PR #343, opened 2016-03-07
ZMQPP library addition

tatraian:
This is a c++11 library, so it won't compile on not C++11 environments. It basis on ZeroMQ that currently doesn't compile on Windows. This is the first pull request of issue #338.

### DIFF
--- a/cmake/projects/ZMQPP/hunter.cmake
+++ b/cmake/projects/ZMQPP/hunter.cmake
@@ -5,6 +5,7 @@ else()
 endif()
 
 include(hunter_add_version)
+include(hunter_cacheable)
 include(hunter_download)
 include(hunter_pick_scheme)
 
@@ -20,5 +21,5 @@ hunter_add_version(
   )
 
 hunter_pick_scheme(DEFAULT url_sha1_cmake)
-
+hunter_cacheable(ZMQPP)
 hunter_download(PACKAGE_NAME ZMQPP)
--- a/cmake/projects/ZMQPP/hunter.cmake
+++ b/cmake/projects/ZMQPP/hunter.cmake
@@ -15,9 +15,9 @@ hunter_add_version(
   VERSION
   "4.1.2"
   URL
-  "https://github.com/hunter-packages/zmqpp/archive/4.1.2-hunter.tar.gz"
+  "https://github.com/hunter-packages/zmqpp/archive/4.1.2-hunter-p2.tar.gz"
   SHA1
-  f7fb7ffda0d34fcc61b562e73f3e548490076ea9
+  549d2d0078ed636d90ed0febe687a058bdd0d64d
   )
 
 hunter_pick_scheme(DEFAULT url_sha1_cmake)
--- a/cmake/projects/ZMQPP/hunter.cmake
+++ b/cmake/projects/ZMQPP/hunter.cmake
@@ -14,9 +14,9 @@ hunter_add_version(
   VERSION
   "4.1.2"
   URL
-  "/home/kobold/Developments/zmqpp.tar.gz"
+  "https://github.com/hunter-packages/zmqpp/archive/4.1.2-hunter.tar.gz"
   SHA1
-  4934c38a30a7f6eb54acc7f13c713bad5c838daf
+  f7fb7ffda0d34fcc61b562e73f3e548490076ea9
   )
 
 hunter_pick_scheme(DEFAULT url_sha1_cmake)
--- a/cmake/projects/ZMQPP/hunter.cmake
+++ b/cmake/projects/ZMQPP/hunter.cmake
@@ -1,0 +1,24 @@
+if(DEFINED HUNTER_CMAKE_PROJECTS_ZMQPP_HUNTER_CMAKE_)
+  return()
+else()
+  set(HUNTER_CMAKE_PROJECTS_ZMQPP_HUNTER_CMAKE_ 1)
+endif()
+
+include(hunter_add_version)
+include(hunter_download)
+include(hunter_pick_scheme)
+
+hunter_add_version(
+  PACKAGE_NAME
+  ZMQPP
+  VERSION
+  "4.1.2"
+  URL
+  "/home/kobold/Developments/zmqpp.tar.gz"
+  SHA1
+  4934c38a30a7f6eb54acc7f13c713bad5c838daf
+  )
+
+hunter_pick_scheme(DEFAULT url_sha1_cmake)
+
+hunter_download(PACKAGE_NAME ZMQPP)
--- a/examples/ZMQPP/CMakeLists.txt
+++ b/examples/ZMQPP/CMakeLists.txt
@@ -4,12 +4,11 @@ cmake_minimum_required(VERSION 3.2)
 # * https://github.com/hunter-packages/gate
 include("../common.cmake")
 
-project(download-ZeroMQ)
+project(download-ZMQPP)
 
-hunter_add_package(ZeroMQ)
 hunter_add_package(ZMQPP)
 
-find_package(ZMQPP)
+find_package(ZMQPP CONFIG REQUIRED)
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
 add_executable(version main.cpp)
--- a/examples/ZMQPP/CMakeLists.txt
+++ b/examples/ZMQPP/CMakeLists.txt
@@ -1,0 +1,16 @@
+cmake_minimum_required(VERSION 3.2)
+
+# Emulate HunterGate:
+# * https://github.com/hunter-packages/gate
+include("../common.cmake")
+
+project(download-ZeroMQ)
+
+hunter_add_package(ZeroMQ)
+hunter_add_package(ZMQPP)
+
+find_package(ZMQPP)
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+add_executable(version main.cpp)
+target_link_libraries(version ZMQPP::zmqpp)
--- a/cmake/configs/default.cmake
+++ b/cmake/configs/default.cmake
@@ -111,9 +111,7 @@ hunter_config(sparsehash VERSION 2.0.2)
 hunter_config(spdlog VERSION 1.0.0-p0)
 hunter_config(szip VERSION 2.1.0-p1)
 hunter_config(wxWidgets VERSION 3.0.2)
-hunter_config(ZeroMQ VERSION 4.1.4-p2)
 hunter_config(ZMQPP VERSION 4.1.2)
-hunter_config(Protobuf VERSION 3.0.0-beta-2)
 
 if(MINGW)
   # 1.60.0 broken:
--- a/cmake/configs/default.cmake
+++ b/cmake/configs/default.cmake
@@ -111,6 +111,9 @@ hunter_config(sparsehash VERSION 2.0.2)
 hunter_config(spdlog VERSION 1.0.0-p0)
 hunter_config(szip VERSION 2.1.0-p1)
 hunter_config(wxWidgets VERSION 3.0.2)
+hunter_config(ZeroMQ VERSION 4.1.4-p2)
+hunter_config(ZMQPP VERSION 4.1.2)
+hunter_config(Protobuf VERSION 3.0.0-beta-2)
 
 if(MINGW)
   # 1.60.0 broken:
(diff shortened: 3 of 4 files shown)


cmake/projects/ZMQPP/hunter.cmake
```diff
@@ -0,0 +1,24 @@
...
+  VERSION
+  "4.1.2"
+  URL
+  "https://github.com/hunter-packages/zmqpp/archive/4.1.2-hunter.tar.gz"
+  SHA1
+  f7fb7ffda0d34fcc61b562e73f3e548490076ea9
+  )
+
+hunter_pick_scheme(DEFAULT url_sha1_cmake)
+
```
ruslo: Since `ZMQPPConfig.cmake` installed this package can be cacheable: https://github.com/ruslo/hunter/wiki/dev.modules#hunter_cacheable

examples/ZMQPP/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+cmake_minimum_required(VERSION 3.2)
+
+# Emulate HunterGate:
+# * https://github.com/hunter-packages/gate
+include("../common.cmake")
+
+project(download-ZeroMQ)
+
+hunter_add_package(ZeroMQ)
```
ruslo: Dependencies will be installed automatically, this line can be removed.

examples/ZMQPP/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+hunter_add_package(ZMQPP)
+
+find_package(ZMQPP)
```
ruslo: Add `CONFIG REQUIRED`

tatraian: It has been fixed. But Huston I have a problem. The cache is missing in ZeroMQ too. Can I commit fix here, or do I create a new pull request.

ruslo:
> The cache is missing in ZeroMQ too

Please open new pull request. In this case it will be easier to test change.

tatraian:
PR #344 has been opened. I think until ZeroMQ cannot caching this example won't work.

ruslo:
> PR #344 has been opened

Merged

ruslo:
Testing:
* https://github.com/ingenue/hunter/branches/all?utf8=%E2%9C%93&query=pr-343

ruslo:
> Testing

All windows tests failed. Example of error message:
```
CUSTOMBUILD : CMake error : install(EXPORT "ZeroMQTargets" ...) includes target "libzmq" more than once in the export set.
```

Some linux/osx passed, some failed:
* https://travis-ci.org/ingenue/hunter/builds/114704154

@tatraian Branch `develop` updated. Please rebase.

tatraian:
ZeroMQ currently runs only linux with hunter. I fixed some Windows build issues on it, but currently it doesn't find windows symbol database. I had Windows (enlarged my C partition at home...), so now I can fix these issues.I open new pull request(s) for ZeroMQ's windows build.

ruslo:
> I open new pull request(s) for ZeroMQ's windows build

Just for the record, depends on: https://github.com/hunter-packages/zeromq4-1/pull/1

tatraian:
And the other dependency: #355

ruslo:
> And the other dependency: #355

Merged

ruslo:
@tatraian I see some new commits, wonder if it is ready?

tatraian:
Naturally no! I think I will create a new pull request for ZMQPP library, but just after Easter!

ruslo:
> Naturally no

Okay, no problems. Just asking :)

ruslo:
I can't merge it anyway, it has conflicts.

tatraian:
motto: better later than never

I rebased the ZMQPP branch to develop. This version works for me on linux but C++11 is required for it.

examples/ZMQPP/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+# * https://github.com/hunter-packages/gate
+include("../common.cmake")
+
+project(download-ZMQPP)
+
+hunter_add_package(ZMQPP)
+
+find_package(ZMQPP CONFIG REQUIRED)
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
```
ruslo: This flag should be set in toolchain since it affects compatibility. I'm planning to add examples with in depth explanation in future: https://github.com/ruslo/hunter/issues/32. Toolchain `default` should fail and excluded from testing and uploading. This will show the fact that some of the library in system require C++11 support. At least please add:
```cmake
string(COMPARE EQUAL "${CMAKE_TOOLCHAIN_FILE}" "" no_toolchain)
if(no_toolchain)
  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
endif()
```
Also note that you need to check that flag is appropriate for the current compiler (e.g. `-std=c++11` is not for MSVC). Probably it's better to use abstraction [CMAKE_CXX_STANDARD](https://cmake.org/cmake/help/latest/variable/CMAKE_CXX_STANDARD.html):
```cmake
string(COMPARE EQUAL "${CMAKE_TOOLCHAIN_FILE}" "" no_toolchain)
if(no_toolchain)
  set(CMAKE_CXX_STANDARD 11)
endif()
```

tatraian: Won't be better if I changed this something like this (before hunter_add_package):

```cmake
if(CMAKE_CXX_STANDARD LESS 11)
  message(FATAL_ERROR, "C++11 required for ZMQPP")
endif()
```

in this case the user will get a correct error message. Furthermore I'm not sure that add c++11 attribute to compiler flags just only for project won't cause silent or noisy link error...

Can I force this check somewhere in the ZMQPP hunter definition? (Earlier the ZMQPP's `CMakelists.txt`).

ruslo: > if(CMAKE_CXX_STANDARD LESS 11)

Toolchains add `-std=c++11` flag to `CMAKE_CXX_FLAGS` manually, `CMAKE_CXX_STANDARD` is not set. Using `CMAKE_CXX_STANDARD` require CMake 3.1+:
* https://cmake.org/cmake/help/v3.0/variable/CMAKE_CXX_STANDARD.html (not found)
* https://cmake.org/cmake/help/v3.1/variable/CMAKE_CXX_STANDARD.html (found)

Also this bug should be fixed first (when `CMAKE_CXX_STANDARD` set we need to add `-std=c++11` manually to custom schemes):
* https://github.com/ruslo/hunter/issues/322

However I think this is okay to have it in examples since last version of CMake used for testing/uploading (but still under `if(no_toolchain)` condition).

> in this case the user will get a correct error message

Usually such checks done in `hunter.cmake` file but I don't know how to do it nicely in this case. First problem is that we need to create module which read both `CMAKE_CXX_STANDARD` and `CMAKE_CXX_FLAGS`. Second problem is that when `hunter.cmake` loaded  `CMAKE_CXX_FLAGS` may be in modified state (i.e. may contain `-std=c++11` locally but not globally, well wrongly).

You can add check to C++ code. `__cplusplus` macro set to `201103L` for `c++11`.

pretyman: I think it will also be needed to patch `CMakeLists.txt` of `zmqpp` to remove the lines at https://github.com/hunter-packages/zmqpp/blob/hunter/CMakeLists.txt#L17-L21

```cmake
# prepare C++11
set(CMAKE_CXX_FLAGS "-std=c++11 ${CMAKE_CXX_FLAGS}")

# show all warnings
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -pedantic -Wextra")
```

So it takes whatever it set by the toolchain.

I guess when hunter moves to CMake 3.1, ideally we would add
```cmake
set_target_properties(... PROPERTIES
    CXX_STANDARD 11
    CXX_STANDARD_REQUIRED ON
)
```

ruslo: > I think it will also be needed to patch CMakeLists.txt of zmqpp to remove the lines at 

Yes. Again, at least under condition `if(no_toolchain)`.

> I guess when hunter moves to CMake 3.1, ideally we would add

I think this CMake feature leads to conflicts in some cases. Setting standard version locally either by modifying `CMAKE_CXX_FLAGS` or by `CXX_STANDARD` is a bad design. The reason is a compatibility of dependencies. If project `Foo` set `CXX_STANDARD 98` and `Boo` set `CXX_STANDARD 11` and both of them depends on header-only Boost then for `Foo` and `Boo` different incompatible structures/methods will be generated from Boost. The only way to fix this I see in setting it globally in toolchain (by `CXX_STANDARD` or by `CMAKE_CXX_FLAGS`).

tatraian: @ruslo please clarify! You wanna add c++11 flags in example `CMakeLists.txt` or I can raise error message but currently I have to check the `CMAKE_CXX_FLAGS` variable instead of `CMAKE_CXX_STANDARD` variable?

Now as I see the `CMakelists.txt` of zmqpp the building of ZMQPP will failed if C++11 is not enabled (on linux).

ruslo: > You wanna add c++11 flags in example CMakeLists.txt or I can raise error message but currently I have to check the CMAKE_CXX_FLAGS variable instead of CMAKE_CXX_STANDARD variable?

@tatraian This is up to you. Toolchains set `CMAKE_CXX_FLAGS` so you can check it. In general `CXX_STANDARD` can be checked too. Also this can be done for both in C++ code of the library instead of CMake code.

Set of `c++11` standard in local project should be done under condition `if(no_toolchains)` (both in example and top CMakeLists.txt of ZMQPP).

> Now as I see the CMakelists.txt of zmqpp the building of ZMQPP will failed if C++11 is not enabled (on linux).

This is okay since it is a requirement. Toolchain with name `default` must fail, I will exclude it from testing. Also adding `-std=c++11` may not help in all cases since old compilers from Travis environment may be missing some c++11 features anyway.

tatraian: I have a problem. I added the following lines to the example CMakelists.txt, before `hunter_add_package` command:

```cmake
if(NOT MSVC)
  string(COMPARE EQUAL "${CMAKE_TOOLCHAIN_FILE}" "" no_toolchain)
  if(no_toolchain)
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
    set(CMAKE_CXX_STANDARD 11)
  endif()
endif()
```
I debugged with messages that we enter in the second condition and set the variables but looks like it has no effect. I got the c++11 missing error message from the ZMQPP's CMakelists.txt. I tried to put this variable setting before including common.cmake but has no effect as well. What is wrong?

cmake/projects/ZMQPP/hunter.cmake
```diff
@@ -0,0 +1,25 @@
...
+hunter_add_version(
+  PACKAGE_NAME
+  ZMQPP
+  VERSION
+  "4.1.2"
+  URL
+  "https://github.com/hunter-packages/zmqpp/archive/4.1.2-hunter.tar.gz"
+  SHA1
+  f7fb7ffda0d34fcc61b562e73f3e548490076ea9
```
ruslo: Wrong SHA1. I got: `ec402dba0ea1c1b31ba5dfd31574d45e3c6a5583`

examples/ZMQPP/CMakeLists.txt
ruslo: > I added the following lines to the example CMakelists.txt, before hunter_add_package

> I got the c++11 missing error message from the ZMQPP's CMakelists.txt

Setting C++ flags locally in project doesn't affect C++ flags for 3rd party packages from Hunter. Toolchain should be used. As a workaround you can set `c++11` manually but protect it with `if(no_toolchain)` condition.

cmake/projects/ZMQPP/hunter.cmake
tatraian: Yes. Sorry this is my error. I forgot commit it.

ruslo:
Testing:
* https://travis-ci.org/ingenue/hunter/builds/119710974

ruslo:
> Testing

Some tests excluded:
* https://github.com/ingenue/hunter/commit/af8342f275f2090f6e20d9d8e5269df7d969efbf

ruslo:
Uploaded binaries matrix:
* https://travis-ci.org/ingenue/hunter/builds/119762873

New wiki, feel free to add more info:
* https://github.com/ruslo/hunter/wiki/pkg.zmqpp

tatraian:
Ok, thanks!

PR #353, opened 2016-03-13
SQLite3 in Linux to link with pthread and dl

pretyman:
Fix for Linux tests when linking to SQLite3 statically, they fail with unresolved symbols from `pthread` and `dl` library, such as:

https://travis-ci.org/ingenue/hunter/jobs/114990295
https://travis-ci.org/ingenue/hunter/jobs/114990296
https://travis-ci.org/ingenue/hunter/jobs/114990297

I'm sure if this is considered a proper fix or just a work-around. It was the best I could come up with.

### DIFF
--- a/cmake/find/FindSQLite3.cmake
+++ b/cmake/find/FindSQLite3.cmake
@@ -47,6 +47,20 @@ if(NOT TARGET "SQLite3")
           "${SQLITE3_LIBRARY}"
         DIRECTORY CACHE
     )
+
+    if(UNIX AND NOT APPLE AND NOT ANDROID) # if Linux
+      string(REGEX MATCH "\\.a$" _ends_in_dot_a "${SQLITE3_LIBRARY}")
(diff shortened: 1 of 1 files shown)


ruslo:
> I'm sure if this is considered a proper fix or just a work-around

You can use [FindThreads](https://cmake.org/cmake/help/latest/module/FindThreads.html) module and [CMAKE_DL_LIBS](https://cmake.org/cmake/help/latest/variable/CMAKE_DL_LIBS.html) variable.

pretyman:
Updated. I wasn't sure if I should have used `THREADS_PREFER_PTHREAD_FLAG` since I don't know if it would interfere with the toolchain or not.

Thanks

pretyman:
Sorry, updated again. I had the `find_package(Threads ...)` logic outside the if. It is fixed now.

ruslo:
Testing:
* https://github.com/ingenue/hunter/branches/all?utf8=%E2%9C%93&query=pr-353

ruslo:
> Testing

All tests passed.

PR #364, opened 2016-03-23
Added the Qt backend to OpenCV desktop

pretyman:
Depends on https://github.com/hunter-packages/QtCMakeExtra/pull/2 on OSX, or else run into linker errors on   `QPlatformPrintDevice`


cmake/configs/default.cmake
```diff
@@ -62,7 +62,7 @@ hunter_config(Libcxxabi VERSION ${HUNTER_Clang_VERSION})
 hunter_config(Libssh2 VERSION 1.7.0)
 hunter_config(Lua VERSION 5.3.2)
 hunter_config(MySQL-client VERSION 6.1.6)
-hunter_config(OpenCV VERSION 3.1.0-p1)
+hunter_config(OpenCV VERSION 3.1.0-p2 CMAKE_ARGS WITH_QT=YES)
```
ruslo: So it will trigger Qt build? I don't think we need to set this by default. User can set this locally if needed:

pretyman: Yes, for non iOS, Android and WinRT it would trigger Qt5 base component by default as per OpenCV's CMakeLists.txt:
`OCV_OPTION(WITH_QT             "Build with Qt Backend support"               OFF  IF (NOT ANDROID AND NOT IOS AND NOT WINRT) )`

My reasoning behind setting it by default is that the Qt backend is enabled by default when downloading OpenCV from the site and when downloading highgui module on the average linux distribution. So it would create a more seamless transition to hunter.

But I do understand the burden of having to compile Qt, as I've been testing this package, so I'm also happy with an entry in the wiki about it.

Your call.

ruslo: > Your call

Now it will break CI testing/uploading of OpenCV. I'm working on Qt uploading to cache from my local machine so my suggestion is to pause this merge until I've finished so we can have this cool automation working. Also I don't want to make a decision myself since I can't be a guy who is responsible for every package in system. I think I should delegate end verdict to the [OpenCV team](https://github.com/orgs/hunter-packages/teams/opencv): @headupinclouds what do you think?

ruslo: > My reasoning behind setting it by default

It should be set in `hunter.cmake` of OpenCV package:
* https://github.com/ruslo/hunter/wiki/usr.adding.new.package#cmake-options

pretyman: Previously I've tried setting it in `hunter.cmake` with `hunter_cmake_args ` as per your suggestion, but I felt that was rather fragile since I could not specify the version that those parameters would apply. 

If one would chose to use a previous version that was not patched for the Qt backend, such as 2.4 or 3.0, the build would break.

ruslo: > If one would chose to use a previous version in their local config that was not patched for the Qt backend, such as 2.4 or 3.0, the build would break

But same arguments can be applied for the opposite situation when I want to set compatible version and `WITH_QT=YES` will unexpectedly gone. I think leaving a note to the wiki is a reasonable choice. Also you can check version in `hunter.cmake`, see for example Qt:
* https://github.com/ruslo/hunter/blob/d6177f64253d19c3853a3511a8af82ea99d2e145/cmake/projects/Qt/hunter.cmake.in#L18

pretyman: I think putting it inside `projects/OpenCV/hunter.cmake` with an explicit test on `HUNTER_OpenCV_VERSION` would be the best solution then, as pointed out in the Qt example.

I'll drop a message when I update the code and we can leave this pull request on hold until the change of using Qt from cache and the feedback of @headupinclouds

pretyman:
Updated this pull request

ruslo:
> Updated this pull request

Thanks. I'm still working on Qt upload. Hope to fix it soon.

headupinclouds:
@pretyman , @ruslo : I just noticed I was mentioned in the previous discussion for this PR (dating back to Mar 23).  I apologize for the lack of response, I never saw it.  Is there still an open question about this?

ruslo:
> Is there still an open question about this?

Yep, should we set `WITH_QT=YES` as a default option. Pros: same configuration as default build. Cons: OpenCV will depend on Qt.

headupinclouds:
I don't have a strong opinion on this, since users can always override the default, but in a build-from-source scenario, I probably wouldn't make `WITH_QT=YES` the default option.  Preserving default options seems to be a good goal, but I think the cost is a little too high in this context, and the Qt highgui extensions are probably used by a very small # of users.  I see two main issues with a Qt dependency: incredibly long build times (not travis friendly (and potential frustration for users when it fails)) and potential package instability/complexity.  I also expect there is a limited # of folks who will tackle the learning curve and time required to maintain and troubleshoot hunter + Qt issues (waiting several hours before hitting a build error can be significant road block).  Since hunter is aiming to work everywhere, this seems to be a bit of a liability to me. The binary cache feature @ruslo is working on will definitely help, but will only cover a small # of platforms + toolchains.  I also think a hunter upgrade that triggers an additional Qt build would probably be unwelcome surprise to most users.

I would be interested to hear from other users on this.  @pretyman: Any comments?

pretyman:
As stated above, I would set it on by default because that would match the behavior when downloaded from the site, and that would create a seamless transition to hunter. When my classmates looked at my projects, they did look surprised I had no toolbar from HighGui.

But your @headupinclouds  points are valid, so I will remove this code and leave a note in the Wiki to add WITH_QT=YES to the package config.

ruslo:
> Qt version 5.5.1-cvpixelbuffer-2-p7 is not building in OSX 10.11 with polly toolchain osx-10-11. I have still to investigate why

Try this version: https://github.com/ruslo/hunter/releases/tag/v0.14.21 I forget to commit some fixes :)

pretyman:
Updated to remove the version check. When this pull request is merged I will add a entry for the wiki.

> Try this version: https://github.com/ruslo/hunter/releases/tag/v0.14.21 I forget to commit some fixes :)

Thanks! Qt compiled, however no Qt Widgets application fails to link because of the lack of https://github.com/hunter-packages/QtCMakeExtra/pull/2 , anyway to get that pull request merged? I cannot get any Qt Widgets application to link in OSX without it. Would you know if it works for you? `examples/qt-widgets` I know it won't link for me in OSX. When I compiled the OpenCV example in `examples/OpenCV` it fails with:

```
Undefined symbols for architecture x86_64:
  "QPlatformPrintDevice::createPageSize(QString const&, QSize const&, QString const&)", referenced from:
      QCocoaPrintDevice::createPageSize(OpaquePMPaper* const&) const in libqcocoa.a(qcocoaprintdevice.o)
  "QPlatformPrintDevice::QPlatformPrintDevice(QString const&)", referenced from:
      QCocoaPrintDevice::QCocoaPrintDevice(QString const&) in libqcocoa.a(qcocoaprintdevice.o)
  "QPlatformPrintDevice::QPlatformPrintDevice()", referenced from:
      QCocoaPrintDevice::QCocoaPrintDevice() in libqcocoa.a(qcocoaprintdevice.o)
      QCocoaPrintDevice::QCocoaPrintDevice() in libqcocoa.a(qcocoaprintdevice.o)
  "QPlatformPrintDevice::~QPlatformPrintDevice()", referenced from:
      QCocoaPrintDevice::QCocoaPrintDevice(QString const&) in libqcocoa.a(qcocoaprintdevice.o)
      QCocoaPrintDevice::~QCocoaPrintDevice() in libqcocoa.a(qcocoaprintdevice.o)
  "QPlatformPrinterSupport::createPrintDevice(QPlatformPrintDevice*)", referenced from:
      QCocoaPrinterSupport::createPrintDevice(QString const&) in libqcocoa.a(qcocoaprintersupport.o)
  "QPlatformPrinterSupport::createDefaultPrintDevice()", referenced from:
      vtable for QCocoaPrinterSupport in libqcocoa.a(qcocoaprintersupport.o)
  "QPlatformPrinterSupport::QPlatformPrinterSupport()", referenced from:
      QCocoaPrinterSupport::QCocoaPrinterSupport() in libqcocoa.a(qcocoaprintersupport.o)
      QCocoaPrinterSupport::QCocoaPrinterSupport() in libqcocoa.a(qcocoaprintersupport.o)
  "QPlatformPrinterSupport::~QPlatformPrinterSupport()", referenced from:
      QCocoaPrinterSupport::~QCocoaPrinterSupport() in libqcocoa.a(qcocoaprintersupport.o)
      QCocoaPrinterSupport::~QCocoaPrinterSupport() in libqcocoa.a(qcocoaprintersupport.o)
      QCocoaPrinterSupport::~QCocoaPrinterSupport() in libqcocoa.a(qcocoaprintersupport.o)
  "QPlatformPrintDevice::makeAndModel() const", referenced from:
      vtable for QCocoaPrintDevice in libqcocoa.a(qcocoaprintdevice.o)
  "QPlatformPrintDevice::defaultInputSlot() const", referenced from:
      QCocoaPrintDevice::loadInputSlots() const in libqcocoa.a(qcocoaprintdevice.o)
      QCocoaPrintDevice::defaultInputSlot() const in libqcocoa.a(qcocoaprintdevice.o)
  "QPlatformPrintDevice::defaultOutputBin() const", referenced from:
      QCocoaPrintDevice::loadOutputBins() const in libqcocoa.a(qcocoaprintdevice.o)
      QCocoaPrintDevice::defaultOutputBin() const in libqcocoa.a(qcocoaprintdevice.o)
  "QPlatformPrintDevice::isValidPageLayout(QPageLayout const&, int) const", referenced from:
      vtable for QCocoaPrintDevice in libqcocoa.a(qcocoaprintdevice.o)
  "QPlatformPrintDevice::supportedPageSize(QPageSize::PageSizeId) const", referenced from:
      vtable for QCocoaPrintDevice in libqcocoa.a(qcocoaprintdevice.o)
  "QPlatformPrintDevice::supportedPageSize(QSize const&) const", referenced from:
** BUILD FAILED **
```

ruslo:
> anyway to get that pull request merged?

I'm still working on Qt upload. You can link those libraries manually for now. This should be fixed only after we have full automated testing, otherwise this is endless `fix A` -> `broke B` -> `fix B` -> `broke A` game.

pretyman:
Ok, thanks!

ruslo:
@pretyman I'm ready to test it.

ruslo:
> @pretyman I'm ready to test it.

@pretyman I don't see such updates like "reactions", please write me a message.

Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.668
* https://travis-ci.org/ingenue/hunter/builds/144746373

pretyman:
Ok, sorry, I lost track of this, I guess you finished the cached Qt build?

Should I update the wiki on how to enable to Qt backend or best wait until you merge `develop` branch into `master`?

pretyman:
Hum.... I guess to test this pull request, it would need a config file with `WITH_QT=YES` 

As in `hunter_config(OpenCV VERSION 3.1.0-p2 CMAKE_ARGS WITH_QT=YES)` ? 

I would think a new branch in https://github.com/ingenue/hunter would be needed just to treat this scenario?

pretyman:
Travis iOS build on https://travis-ci.org/ingenue/hunter/builds/144746373 failed with `The job exceeded the maximum time limit for jobs, and has been terminated.` 

On the other hand, I don't think the iOS ( + other mobile platforms) would make sense here, since the Qt backend targets the desktop with QtWidgets. I can disable these toolchains, just get back to me on the config issue of the comment above.

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.14.30

ruslo:
> Should I update the wiki on how to enable to Qt backend

Yes, sure.

> I would think a new branch in https://github.com/ingenue/hunter would be needed just to treat this scenario?

Let's have same branch but with one more example.

> On the other hand, I don't think the iOS ( + other mobile platforms) would make sense here

No, OpenCV is totally ok with iOS, I just need to upload binary cache from local machine because of timeout: https://travis-ci.org/ingenue/hunter/builds/144939737. However if OpenCV + Qt on iOS doesn't make sense you can exclude it from testing.

pretyman:
> Let's have same branch but with one more example.

You mean I can have a `hunter_config(...)` call inside of the example's CMakeLists.txt to add the `CMAKE_ARGS WITH_QT=YES`? How should I approach it?

> However if OpenCV + Qt on iOS doesn't make sense you can exclude it from testing.

Sorry, yes, I meant OpenCV + Qt backend. I just never seen anyone use it in iOS up to now.

Can you please review the text below to be added to the OpenCV page on the wiki:

# Qt backend 

Qt backend support for desktop can be enabled by [creating a custom config](https://github.com/hunter-packages/gate/tree/develop#usage-custom-config) with the line:

 ```cmake
hunter_config(OpenCV VERSION ... CMAKE_ARGS WITH_QT=YES)
```

Where `...` should be replaced with a version of `3.1.0-p1` or higher

ruslo:
> You mean I can have a hunter_config(...)

Yes, add one more example OpenCV-with-Qt. You can set custom config.cmake location using TESTING_CONFIG_OPT.

> Where ... should be replaced with a version of 3.1.0-p1 or higher

Add here that `${HUNTER_OpenCV_VERSION}` can be used.

pretyman:
Updated wiki: https://github.com/ruslo/hunter/wiki/pkg.opencv

Issue #194, opened 2015-07-02
Correct reusing of Cache directory


ruslo:
While implementing the [cache feature](https://github.com/ruslo/hunter/issues/74) I've purposely omitted `SHA1 of internal dependent files (like build scheme)` step since I was thinking that if there will be updates on internal modules this updates will fix broken packages. And since broken package will not be cached there no need to track this dependency:

* try build package -> fail -> not cached
* update internal modules
* try build package -> ok -> cached (correct)

But looks like this design have some flaws, example: https://github.com/ruslo/hunter/issues/190

* try build boost (incorrect zlib link) -> ok -> cached (bad version!)
* update internal modules # *SHA1 of sources and arguments not change*
* try build boost -> reuse cache (no updates! package still broken)

Here are some proposals of how it can be fixed:

### A. Add SHA1 values of modules to cache metadata (manually)

Manually add the files that we think may affect the build. We can add any files we need and ignore others. Fine-tune but error-prone (we can miss something on updates). Not effective in some cases (adding comments to code result different SHA1). Also it's hard to track some dependencies (boost *may* depends on ZLIB, ZLIB depends on url_sha1_cmake scheme).

### B. Add SHA1 values of modules to cache metadata (automatically)

Hard to implement I guess. Also some files not needed, like if we add some print symbol to `hunter_status_debug` function which used almost everywhere but not change build behaviour.

### C. Add SHA1 of all internal files

Pack such directories as 'cmake/{find, modules, schemes, templates}' and calculate SHA1 of such archive (or even create separate git project HunterCore?). 100% correct but not optimal since there will be quite a lot of updates (I guess) and we will be close to situaltion when every release will use new SHA1 and cache will be useless.

### D. Internal version

There is variable `HUNTER_TOOLCHAIN_UNDETECTABLE_ID` in toolchain-id calculation module that can be used to update version of internal modules. I.e. when we hit "bad version cached" situation we can do "++internal_version" and new cache will be built. Not effective. I.e. we have bad Boost version cached but other packages that do not depends on Boost will be rebuild too.

I guess `D` variant implemented per-package is the most appropriate one - simple/effective and we have full-control.

ruslo:
New wiki section: https://github.com/ruslo/hunter/wiki/dev.modules#hunter_download

ruslo:
> I guess D variant implemented per-package is the most appropriate one - simple/effective and we have full-control

[...]

Issue #74, opened 2015-04-19
Binaries cache


ruslo:
Share built binaries between several `Hunter-ID` directories. For now sharing implemented only for downloaded sources (`${HUNTER_ROOT}/_Base/Download`).

Need to check this idea is working:
* Install package to package's specific location first => patch build schemes so they use `HUNTER_PACKAGE_INSTALL_DIR` instead of global `HUNTER_INSTALL_PREFIX`
* Pack `HUNTER_PACKAGE_INSTALL_DIR` to `*.tar.gz` archive and move it to cache `${HUNTER_ROOT}/_Base/Cache`
* Save meta information using formula:
 * SHA1 of archive of sources **+**
 * SHA1 of cache `*.tar.gz` of dependencies **+**
 * SHA1 of toolchain.info **+**
 * SHA1 of args.cmake (build options, cmake_args) **+**
 * SHA1 of internal dependent files (like build scheme) (?) **=**
 * SHA1 of cache `*.tar.gz`
* Unpack `*.tar.gz` archive from cache to `HUNTER_INSTALL_PREFIX` for each `Hunter-ID` build on cache hit (same meta information)

Some major problems to fix/implement:
* Track dependencies for packages. In general we must track all dependencies transitively. Probably introduce some internal variable that holds information about currently building package and when `hunter_download` triggered use this variable to store info `Foo depends on Boo`.
* Fix packages relocation problems. All packages must use CMake imported target format (like `ZLIB::zlib`). As far as I know this is the only sense way to make relocatable package that works nicely with transitive `find_package` commands with target usage requirement. Need to patch existing CMake builds + introduce imported version of `Find*` modules for non-CMake packages like Boost or OpenSSL

ruslo:
I'm testing this feature now. Code is in branch 'bin.cache'.
Release: https://github.com/ruslo/hunter/releases/tag/v0.10.0

As I've already mentioned not all the packages are relocatable by default.
Cache will be used only for packages that marked by [hunter_cacheable][1] command.
For now I've added some packages which I tested: TIFF, ZLIB and GTest.

I will add more so we can build GPUOpenCV from cache:
* OpenCV (fix [bug][2] + dependent [PNG][3])
* Boost (fix [bug][4])
* cvmatio (can be cached already?)
* GPUImage (can be cached already?)

and will close this bug, other packages will be added in future.

Some important changes which are visible by users:
* successful build of cacheable package:
```
[...]

PR #384, opened 2016-04-11
External repos -- trial 2

mdavezac:
@ruslo, This pull-request tries to address your comments from #295.

There are two main differences:
- hunter_config takes an additional (optional) argument specifying the directory of the repo. It is automatically added when computing the unified config. This makes sure (see caveat below) to recompile package that have the same name and version but essentially are different recipes (e.g. point to same version + patch). See the test in *tests/external_repositories*
- Package build-dirs also have the hash of the recipe in their name.

The main caveat I can think of right now is that if a user does not specify the repo in their local config, and yet change the number/order of repos in their CMakeLists.txt, they could be specifying different recipes without triggering a rebuild. I'll address ASAP.

Otherwise, the external repos are still specified by directory names, which is not as pretty as the rest of the hunter api.

Am I missing something else?


mdavezac:
I've taken of the caveat above, as well as some bugs.

At this point HUNTER_${pacakge}_RECIPE is set by the time the unified config is created. If a local config requests a specific repo, then that will be the one used. Otherwise, it should be the first package that comes up in HUNTER_RECIPE_DIRS, and last in the in main hunter repo. Builds will be triggered whenever HUNTER_${package}_REPIPE is changed, one way or another.

cmake/modules/hunter_calculate_config_sha1.cmake
```diff
@@ -20,57 +20,52 @@ function(hunter_calculate_config_sha1 hunter_self hunter_base user_config)
...
-    if(IS_DIRECTORY "${directory_with_projects}/${x}")
-      list(APPEND real_projects "${x}")
+  foreach(repo ${HUNTER_RECIPE_DIRS} "${hunter_self}/cmake")
+    if(NOT EXISTS "${repo}/projects")
+      hunter_internal_error("Directory `${repo}/projects` not exists")
+    endif()
+    if(NOT IS_DIRECTORY "${repo}/projects")
+      hunter_internal_error("Path `${repo}/projects` is not a directory")
     endif()
+    file(GLOB subdirectories RELATIVE "${repo}/projects" "${repo}/projects/*/")
```
ruslo: We are searching in local folder - using `file(GLOB ...)` may lead to the same problems as using `file(GLOB ...)` for collecting source files. If I have `Foo` and `Boo` in my local directory and I run cmake configure then when I do update (like add `Bar` locally or run git pull) there will be no reconfigure retriggered since cmake doesn't know that `Bar` is a dependency.

Note that it's not the case in current code since adding `Bar` affects Hunter-ID and `file(GLOB ...)` will be run on "static" directory.

mdavezac: Indeed, that is the problem with the current API. I am hopping you will accept  the current partial implementation, until I can have time to add an api for tracking remote repositories the same way hunter tracks itself, e.g. named releases with ids.

ruslo: I've not touched all the problems. Just pick fundamental one and not sure what technique are you planning to use to fix it. I'm still thinking that approach #56 is the easiest/correct one to implement.

mdavezac: Eventually, there would be an API to add repositories:

~~~
hunter_add_repository(
   URL https://somewhere.com/something.tar.gz
    SHA256 aaa
)
~~~

And the list of directories, as in the current implementation would not be available to users directly.
That way we can make sure that a repository is correctly versioned. It would downloaded to a directory with the hash in its name. Just the way hunter is.
Does that make sense?
Would it take care of this particular problem?

ruslo: > Eventually, there would be an API to add repositories
> URL https://somewhere.com/something.tar.gz

Is it a Hunter archive? If it's a new package archive how it differs from #56 approach?

> Would it take care of this particular problem?

Don't follow. Current problem is about doing GLOB. Approach #56 add every new package explicitly.

mdavezac:
The main difference with #56 are that:
- #56 allows for case by case only. It is practical only for  setting a particular version in a particular project. Hence it is not so DRY.
- #56 does not allow for new packages that are not in the main hunter repo (e.g. private packages)

cmake/modules/hunter_calculate_config_sha1.cmake
mdavezac: It would an archive of packages, separate from those in Hunter. It mainly meant to reference packages that are not currently in Hunter, whether because those packages are private, or because they have not met their way there (think specialised community). Once repos are added via the syntax above, the glob is as much of an issue as in the current code. The rebuild happens because any change to a repo means a change to its hash, as with HunterID.

ruslo:
> Hence it is not so DRY.

You can make it DRY. For example include global `config.cmake` from local `config.cmake`.

> does not allow for new packages that are not in the main hunter repo (e.g. private packages)

It does. At least it's planned:

> hunter_calculate_config_sha1 module need to find out that we are adding **new package**

mdavezac:
@ruslo,

The last few commits add the API I was talking about. I think it takes care of the issue you raised about `file(GLOB)`.

With this pull request, the standard way of adding a repository of new hunter packages (outside those built in hunter) would be as follows, in main CMakeLists.txt

~~~
include(HunterGate)
HunterGate(
    URL "https://github.com/ruslo/hunter/archive/v0.14.12.tar.gz"
    SHA1 "2d444140db0d53393b5490604006bb4f4896a444"
)
# Must happen after HunterGate and before project
include(hunter_add_repository)
hunter_add_repository(
    URL https://www.github.com/something/some_repo/archive/v0.14.12.tar.gz
    SHA1 "ceegg1a0db0d53393b5490604006bb4f4896a444"
)

project(...)
~~~ 

The repository gets installed to a subdirectory of _Base which includes the SHA1 in its name.

Each repo follows the directory structure of Hunter (projects/, configs/default.cmake, etc) with one difference. A file `init.cmake` lives in the root directory of the repo. It is included when initializing the repository. It should have the following form:

~~~
# These two lines are mandatory
include(hunter_initialize_repository)
hunter_initialize_repository(NAME fakey VERSION 0.0.1)

# Optionally, could add directory with Find X files
list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/Find")
~~~

The name and version can then be used to easily specify a project from a given repos in config.cmake:

~~~
hunter_config(packagename VERSION x.y REPOSITORY_NAME fakey REPOSITORY_VERSION 0.0.1)
~~~

What do you think?

mdavezac:
Also, I've added a test with a fake repository in `tests/hunter_add_repository`

cmake/modules/hunter_add_package.cmake
```diff
@@ -9,6 +9,17 @@ include(hunter_internal_error)
 include(hunter_status_debug)
 include(hunter_test_string_not_empty)
 
+function(_hunter_find_project_dir projectname output)
```
ruslo: Should be in separate file with name `hunter_find_project_dir.cmake`. See "one function - one file" rule:
* http://0.readthedocs.org/en/latest/cmake.html#library-of-cmake-extra-modules

mdavezac:
Moved functions to their own files and did some reformatting to follow the commit guidelines.

tests/external_repositories/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+foreach(project Foo Bar)
+  set(build_dir "${CMAKE_BINARY_DIR}/${project}")
+  execute_process(
+    COMMAND "${CMAKE_COMMAND}"
+    "-H${CMAKE_SOURCE_DIR}/${project}"
+      "-B${build_dir}"
+      "-G${CMAKE_GENERATOR}"
+    RESULT_VARIABLE ${project}_result
+  )
+  if(NOT${project}_result EQUAL 0)
```
ruslo: This will do not what you're expecting. It will check value of variable with name `NOTFoo_result` instead of checking variable with name `Foo_result` and doing negation.

cmake/modules/hunter_add_repository.cmake
```diff
@@ -0,0 +1,81 @@
...
+include(hunter_internal_error)
+include(hunter_fatal_error)
+include(hunter_test_string_not_empty)
+include(hunter_status_debug)
+include(hunter_status_print)
+include(hunter_install_repository)
+
+function(hunter_add_repository)
+  if(PROJECT_NAME OR NOT HUNTER_GATE_ROOT)
+    hunter_fatal_error(
```
ruslo: `WIKI` missing here, see https://github.com/ruslo/hunter/wiki/dev.modules#hunter_fatal_error. You can improve this later and use `hunter_internal_error` for now.

cmake/modules/hunter_add_repository.cmake
```diff
@@ -0,0 +1,81 @@
+include(CMakeParseArguments) # cmake_parse_arguments
+include(hunter_internal_error)
+include(hunter_fatal_error)
+include(hunter_test_string_not_empty)
+include(hunter_status_debug)
+include(hunter_status_print)
+include(hunter_install_repository)
+
+function(hunter_add_repository)
+  if(PROJECT_NAME OR NOT HUNTER_GATE_ROOT)
```
ruslo: With this check it's not possible to include subproject by `add_subdirectory` if both parent and child projects have `hunter_add_repository`. In this scenario subproject settings should be ignored since parent can trigger packages installation to root.

ruslo:
```cmake
hunter_add_repository(
    URL https://www.github.com/something/some_repo/archive/v0.14.12.tar.gz
    SHA1 "ceegg1a0db0d53393b5490604006bb4f4896a444"
)
```
Since now we moved to approach of creating extra archive I got the question why you can't simply do the fork of Hunter and create release with extra packages you want to have? Is the only reason that you don't want to do merges? You have to check conflicts anyway, you have to create releases anyway.

mdavezac:
why you can't simply do the fork of Hunter and create release with extra packages you want to have

Because then I couldn't easily enjoy the updates in Hunter itself.

    Is the only reason that you don't want to do merges?

Not necessarily, it depends.

There are at least four reasons that I can see. The top two are the main ones I'm concerned with. The third is a matter of taste and for you to decide. The fourth is the "why not fork" you are referring to. If it were the only reason, I would just fork indeed.

- it gives sub-communities the extra-flexibility of hosting their own archive repos for specialized codes while more general packages are kept in the main hunter archive. This is the way that [homebrew](http://brew.sh/) or [ubuntu "Personal Package Archive"](https://launchpad.net/ubuntu/+ppas)  operate. It makes it possible for the communities to coordinate their releases, and do so independently of hunter.
- it allows for private archives. I can tell the academics I work with that archives for their code (code A) will not be released until after their paper comes out, while still using hunter to pull it down in their other (code B).
- it could be used to separate the two concerns, code and archives in separate repos, so releases can be managed independently. 
- people can work on new archives while still benefiting from updates to Hunter, and only put back into hunter when they know the archive works. (Not all that different from forking and not a sufficient reason in itself for external archives)

cmake/modules/hunter_add_repository.cmake
mdavezac: I haven't worked with sub-projects in the past, so I'm not sure I understand the pros and cons completely. But I think that's what I want... If I'm not mistaken, this is also the behavior for HunterGate?

In practice, all `hunter_add_repository` must happen before the call to `hunter_finalize` (or more accurately, before the call to `hunter_calculate_config_sha1`) so that the extra packages appear  in the unified `config.cmake`.  Technically, `hunter_add_repository` could be allowed at any point until the first `hunter_add_package`, but forcing inclusion before `project` seems cleaner to me.

Does that sound right?

tests/external_repositories/CMakeLists.txt
mdavezac: Good catch. Corrected in 682db1a.

cmake/modules/hunter_add_repository.cmake
mdavezac: I've switched to `hunter_internal_error` in 4e599c1.

ruslo:
> Because then I couldn't easily enjoy the updates in Hunter itself

So it still means "want to avoid merging".

> it gives sub-communities the extra-flexibility of hosting their own archive repos for specialized codes while more general packages are kept in the main hunter archive

Same can be done with fork - you can pull updates from "main" repo without pushing changes back. Also in future I don't see it as a separate versions for specialized teams. I think everything should go back to upstream for being convenient to user.

> it allows for private archives

If you can create private archive for `hunter_add_repository` why you can't create private archive for `HunterGate`?

> it could be used to separate the two concerns, code and archives in separate repos, so releases can be managed independently

Not sure I understand that.

> people can work on new archives while still benefiting from updates to Hunter, and only put back into hunter when they know the archive works. (Not all that different from forking and not a sufficient reason in itself for external archives)

Yep, same as fork/merge.

To summarize: I'm not saying that it's not possible but trying to understand what is the scenarios where it will really help (and can't be solved by #56). Also in case of conflicts you still need to merge conflicting branches and resolve conflicts manually.

cmake/modules/hunter_add_repository.cmake
ruslo: > I haven't worked with sub-projects in the past, so I'm not sure I understand the pros and cons completely

This is a trivial way to add new stuff to local build. It's not a correct way in long terms but fast, good for experimenting.
So you have project with library `foo`:
```
cmake_minimum_required(VERSION 3.0)
project(Foo)
add_library(foo foo.cpp)
```

And you have project with executable `boo` and you want to use `foo`:
```
cmake_minimum_required(VERSION 3.0)
project(Boo)

add_subdirectory(/path/to/.../Foo) # or add_subdirectory(Foo) if it's local

add_executable(boo boo.cpp)
target_link_libraries(boo PUBLIC foo) # use foo from Foo project
```

ruslo: > this is also the behavior for HunterGate?

Only one `HunterGate` parameters will be used. Next called `HunterGate` will do nothing, effectively using settings from the very first.

> Does that sound right?

Should be used oncd - yes. Can be called once - no. You should detect that settings from `hunter_add_repository` already used by Hunter and next calls to `hunter_add_repository` should be ignored.

mdavezac:
@ruslo, does that mean you will never accept the pull-request, even code is up to snuff? Or should I invest time making sure `hunter_add_repository` can only be called once, as per your comment about multiple projects?

ruslo:
> does that mean you will never accept the pull-request, even if the code is up to snuff?

I will accept code without bugs, but it would be nice to have a strong long term use-case (again, in my opinion #56 fits for fast experiments better and it's simplier to implement). For your information I haven't reviewed all the code at this moment, so there may be more errors. Also anyway I can mention your fork somewhere in documentation (or in #56 for the start) so if somebody interested in this feature can try/patch your code.

> as per your comment about multiple projects?

One more scenario you might want to think is about parent/child dependencies. So assume we have package `Boo` which use private stuff added by `hunter_add_repository`. And we have package `Foo` that use `Boo` by `hunter_add_package(Boo)`. Since we should calculate `Config-ID` once then code from `Boo` will not be used, it means that any update of private stuff of `Boo` must be repeated in `Foo` too. This is quite true for `HunterGate` but with `HunterGate` function we need to update only one `URL`/`SHA1`, but if there will be N extra repositories we have to do N updates.

mdavezac:
I've modified the code so that `hunter_add_repository` can be called from sub-projects. Any call occurring  after the first call to `project` is ignored. There is no warning, since `hunter_add_repository` can be called multiple time to add multiple external package archives.

I'm still thinking through the parent/child dependencies scenario. Indeed, it is an annoying issue.

cmake/modules/hunter_create_cache_meta_directory.cmake
```diff
@@ -30,7 +30,7 @@ function(hunter_create_cache_meta_directory cache_directory result)
   file(COPY "${toolchain_info}" DESTINATION "${cache_meta_dir}")
 
   # Save package name and version
-  set(cache_meta_dir "${cache_meta_dir}/${HUNTER_PACKAGE_NAME}")
+  set(cache_meta_dir "${cache_meta_dir}/${HUNTER_PACKAGE_NAME}/${HUNTER_${HUNTER_PACKAGE_NAME}_RECIPE_HASH}")
```
ruslo: This will break cache feature. For example we have `Foo 1.0` in `hunter.cmake` and save it to cache, now we are adding `Foo 2.0` to `hunter.cmake`. Since `hunter.cmake` change => `*_RECIPE_HASH` change => `cache_meta_dir` changed => cache miss.

More info about cache mechanism may be found here:
* https://github.com/ruslo/hunter/wiki/dev.modules#hunter_download
* https://github.com/ruslo/hunter/issues/74
* https://github.com/ruslo/hunter/issues/194

ruslo:
One more thought just came to mind:

I'm still thinking that the only benefit of this approach is about avoiding `merge`/`release` step (vs just `release`). So what do you think about doing "merge" operation automatically? E.g. extend `HunterGate` so it accept several `URL`/`SHA1` and unpack them to the new directory one by one. Or new function to create "merged" archive, like:

```cmake
HunterGateMixArchives(
    URL ${url-A} SHA1 ${sha1-A} # mainstream
    URL ${url-B} SHA1 ${sha1-B} # experimental
    RESULT_URL url RESULT_SHA1 sha1 # temporary mixed
)

HunterGate(URL ${url} SHA1 ${sha1} <more-options>)
```

?

mdavezac:
That's a sensible API, I think. Compared to the current implementation, it means moving `hunter_add_repository` to hunter gate.

Although, if by merge, you mean more of something like `git merge`, where hunter.cmake from one set of archives is merged into `hunter.cmake` from the builtin archives... Frankly, that would be difficult.

I'll look into the caching. Not too surprised I bungled it 😖

ruslo:
> Although, if by merge, you mean more of something like git merge

No I don't mean it. If you want `git merge` then, well... fork and do git merge! :) I think `HunterGateMixArchives` should just unpack archives one by one overwriting files if paths are the same. I think this may lead to some tricky stuff but if you want full control you should use more mature separate tools like git. The main benefits are (if we are talking about `HunterGateMixArchives` approach):
* `HunterGate` doesn't change
* Resulting "mixed" archives are predictable/stable
* Doesn't affect internal Hunter stuff

mdavezac:
The main problem I see with that approach is 
i. merging the default.cmake into a single unified file (minus user provided file?)  is now the job of huntergate. I'm still in favor of keeping archives sets orthogonal, as much as possible, so it is necessary to unify them.
ii. if two repos provide the same package, then there is no way to select the one that gets overwritten (yes, that's for non-orthogonal archive sets. But flexibility is nice.) 

It is probably cleaner to have hunter-gate download all the archive sets, but I'm not so sure about just dumping them all into the same directory. I still think archives across several directories is simpler and more flexible.

mdavezac:
I've modified the caching such that it depends on the name of the repo (not the version or hash). I do not understand enough about the caching mechanism to be completely sure it is correct. But it should basically mirror the current mechanism, while adding a repo specificity.

ruslo:
> I've modified the caching such that it depends on the name of the repo

It should not depend on repository. Cache is shared between several `Hunter-ID`, i.e. repository from where package came should not affect cache internal structure layout.

mdavezac:
I was thinking that within a repo it is easy to ensure package names are consistent. This is not necessarily true across repositories, and so packages should be named after the repo where they are held.

If you're not convinced, let me know and I'll go back to the original implementation.

ruslo:
> I was thinking that within a repo it is easy to ensure package names are consistent

Sorry, I don't follow. As I said logic is next: cache used to share binaries between several `Hunter-ID`, hence between different extra repositories too, hence there should be no names of repositories or `Hunter-ID` saved in meta information. Also users are free to name extra repositories as they want, so there is no guarantee that same name will be used for same URL.

mdavezac:
I've moved back to the original implementation for caching

cmake/modules/hunter_create_cache_meta_directory.cmake
```diff
@@ -29,7 +29,6 @@ function(hunter_create_cache_meta_directory cache_directory result)
   endif()
   file(COPY "${toolchain_info}" DESTINATION "${cache_meta_dir}")
 
-  # Save package name and version
```
ruslo: Please return it back. It doesn't relate to this change.

ruslo:
Hi,

Thanks for the update. Just for your information I'll be off for some time,
will be able to do checks about 12-15 May. Thanks again.

Ruslo

cmake/modules/hunter_create_cache_meta_directory.cmake
mdavezac: Okay, it's back to the original

ruslo:
Just found some time for review. @mdavezac are you still interested in this feature?

mdavezac:
indeed! Sorry, it dropped of my radar for a bit there

cmake/modules/hunter_add_repository.cmake
```diff
@@ -0,0 +1,83 @@
...
+
+  # set root for all repositories (except builtin)
+  if(NOT HUNTER_REPOSITORY_ROOT)
+    set(HUNTER_REPOSITORY_ROOT "${HUNTER_GATE_ROOT}/_Base/repositories")
+    set(HUNTER_REPOSITORY_ROOT "${HUNTER_REPOSITORY_ROOT}" PARENT_SCOPE)
+  endif()
+
+  set(rootdir "${HUNTER_REPOSITORY_ROOT}/${har_SHA1}")
+  if(NOT EXISTS "${rootdir}/Done")
+    hunter_install_repository("${rootdir}" "${har_URL}" "${har_SHA1}")
```
ruslo: You're modifying shared directory, you should lock it before.

cmake/modules/hunter_install_repository.cmake
```diff
@@ -0,0 +1,87 @@
...
+      "    \"\"\n"
+      "    INSTALL_COMMAND\n"
+      "    \"\"\n"
+      ")\n"
+  )
+
+  if(HUNTER_STATUS_DEBUG)
+    set(logging_params "")
+  else()
+    # set(logging_params OUTPUT_QUIET)
```
ruslo: You should either remove this line or uncomment it.

cmake/modules/hunter_download.cmake
```diff
@@ -142,9 +142,11 @@ function(hunter_download)
   set(HUNTER_GLOBAL_SCRIPT_DIR "${HUNTER_SELF}/scripts")
   set(HUNTER_PACKAGE_SCRIPT_DIR "${HUNTER_PACKAGE_SETUP_DIR}/scripts/")
   set(HUNTER_PACKAGE_HOME_DIR "${HUNTER_TOOLCHAIN_ID_PATH}/Build")
+  set(repository_hash "${HUNTER_${HUNTER_PACKAGE_NAME}_REPOSITORY_HASH}")
+  hunter_test_string_not_empty("${repository_hash}")
   set(
       HUNTER_PACKAGE_HOME_DIR
-      "${HUNTER_PACKAGE_HOME_DIR}/${HUNTER_PACKAGE_NAME}"
+      "${HUNTER_PACKAGE_HOME_DIR}/${HUNTER_PACKAGE_NAME}/${repository_hash}"
```
ruslo: What if package's dependency changed? Recipe of package doesn't change (hence hash is not affected) but it should be moved to the separate directory.

mdavezac:
@ruslo, sorry about the silence. I hope I can get back to this in September.

cmake/modules/hunter_download.cmake
mdavezac: As far as I understand, hunter resolve this problem by using a different root directory for each hunter release?
To emulate that here would mean setting a different hunter root directory for each possible set of (hunter release, repo A release, repo B release...). That would become quite cumbersome since packages would not be reused from one project to the next (unless the repos are exactly the same).

The other option it to read through the graph of dependencies. But this graph is not explicitly available, is it?

Did I understand this right?

ruslo: > The other option it to read through the graph of dependencies. But this graph is not explicitly available, is it?

This graph is not static. You can't predict will your package (or it's next version) will depends on package Foo. First you have a code:

```cmake
if(MYPACKAGE_MORE_DEPS)
  hunter_add_package(Boo)
endif()
```

and next version is

```cmake
if(MYPACKAGE_MORE_DEPS)
  hunter_add_package(Boo)
  hunter_add_package(Foo) # new dependency!
endif()
```

There is a graph for cached packages, but not all packages are cacheable and graph can be used only after all packages requested by current project will be built.

As a general note I don't think the approach you're trying to implement is doable. You're trying to hack the really basic things Hunter relies on. I recommend you to think about creating Hunter archive on-the-fly either by [mixing several Hunter archives](https://github.com/ruslo/hunter/pull/384#issuecomment-213397810) or by [adding custom hunter.cmake recipies](https://github.com/ruslo/hunter/pull/295#issue-125450831).

PR #387, opened 2016-04-17
Move odb schemes

pretyman:
Move schemes of odb-boost and odb-mysql into the new layout under their respective projects


ruslo:
Testing:
* https://travis-ci.org/ingenue/hunter/builds/123650942

Issue #398, opened 2016-04-21
Add autobahn|cpp library


tatraian:
I would like to add autobahn|cpp library to hunter.

As I see it has the following dependencies:
 - [x] boost [already in hunter]
 - [x] openssl [already in hunter]
 - [x] ZLib [already in hunter]
 - [x] websocketpp [must add to hunter]
 - [x] Msgpack [must add to hunter]

and the package itself

 - [x] autobahn-cpp

I already created [websocketpp](https://github.com/tatraian/websocketpp/tree/0.7.0-hunter) fork and transformed it to hunter compatible. Please create a websocketpp fork into hunter-packages and I send the pull request.

ruslo:
> Please create a websocketpp fork into hunter-packages and I send the pull request

Done

tatraian:
Thanks, PR arrived: #400

tatraian:
Please add msgpack-c and autobahn-cpp repositories to hunter-packages.

ruslo:
> Please add msgpack-c and autobahn-cpp repositories to hunter-packages

This forks:
* https://github.com/msgpack/msgpack-c
* https://github.com/crossbario/autobahn-cpp

?

tatraian:
Yes. Sorry I forgot the urls...

ruslo:
Done:
* https://github.com/hunter-packages/autobahn-cpp
* https://github.com/hunter-packages/msgpack-c

tatraian:
Just for note:
websocketpp version that handles dependencies is arriving in #404 
The other dependency: msgpack-s is arriving in #406

tatraian:
Arriving in PR: #564

PR #400, opened 2016-04-21
websocketpp addition

tatraian:
I added websocketpp to hunter together a simple example.

You can see the ci results [here](https://travis-ci.org/tatraian/hunter_test/builds/124765516) and [here](https://ci.appveyor.com/project/tatraian/hunter-test).

(see issue: #398)


tatraian:
Ups. I see that currently I only install the includes (this is an include only library). But It has ZLIB, OpenSSL and Boost dependency. They were used only during example building, but in philosophy of hunter I think it should be better to add these libraries as linked libraries to the interface target. In this it will be enought to link the final code only to websocketpp.

So I think you can wait for this fix in the hunter-package/websocketpp package.

ruslo:
New wiki page: https://github.com/ruslo/hunter/wiki/pkg.websocketpp

ruslo:
> But It has ZLIB, OpenSSL and Boost dependency. They were used only during example building, but in philosophy of hunter I think it should be better to add these libraries as linked libraries to the interface target

Does websocketpp itself depends on these libraries or the examples depends on them? If just examples then you should add something like `WEBSOCKETPP_BUILD_EXAMPLES=OFF`.

ruslo:
New CI branch:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.400
* https://travis-ci.org/ingenue/hunter/builds/124876307

PR #412, opened 2016-05-12
Add fmt (cppformat) 3.0.0

dpantele:
fmt (previously cppformat) is a nice and fast prtintf/iostreams replacement

Url: http://fmtlib.net/

It defines two exported targets: fmt and fmt-header-only. It is slightly different from what is used in hunter usually (something like fmt::fmt and fmt::header-only), but otherwise everything works out-of-box.


dpantele:
Sry, will reopen another one into develop

PR #415, opened 2016-05-15
Add GLM package

### LINKED ISSUE #281: Add glm package
http://glm.g-truc.net/0.9.7/index.html
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28776893-add-glm-package?utm_campaign=plugin&utm_content=tracker%2F1032790&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1032790&utm_medium=issues&utm_source=github).
</bountysource-plugin>

tbastos:
Fixes #281


ruslo:
Thanks, please set target branch to `develop` (see [wiki](https://github.com/ruslo/hunter/wiki/dev.contribution)).

cmake/projects/GLM/hunter.cmake
```diff
@@ -0,0 +1,25 @@
...
+    VERSION
+    "0.9.7.4"
+    URL
+    "https://github.com/g-truc/glm/archive/0.9.7.4.tar.gz"
+    SHA1
+    cd3d91c1361de38cb2f7603ac06d185d8e10dc01
+)
+
+hunter_cacheable(GLM)
+hunter_configuration_types(Lua CONFIGURATION_TYPES Release)
```
ruslo: s,Lua,GLM,

ruslo: What is the reason of forcing Release here? It doesn't compile for Debug?

tbastos: It's a header only library, so no need to install the same headers twice to the same place.
Yes I noticed the Lua typo a bit too late...

tbastos:
Argh, and I'm in the process of adding other packages on master... when I finish the new packages I'll merge the changes into develop and send a new pull request with several new packages, if that's ok.

That [wiki](https://github.com/ruslo/hunter/wiki/dev.contribution) page is not visible enough. You should have a CONTRIBUTING.md file at the root as explained here: https://help.github.com/articles/setting-guidelines-for-repository-contributors/

Honestly though I would just drop the "develop" branch... it's an unnecessary complexity. The master branch should be the only "eternal" branch, while all development is done in temporary branches/forks and merged back into master as often as possible to minimize integration problems. Pull requests and automated testing already gives you enough protection. This will make things much simpler for you and all contributors...

ruslo:
> You should have a CONTRIBUTING.md file at the root as explained here: https://help.github.com/articles/setting-guidelines-for-repository-contributors/

Oh... I have this file before but removed it. Returning it back: https://github.com/ruslo/hunter/commit/f0a8d0c70e02f3671067a391cdd34d0151e6db73

> Pull requests and automated tests already give you enough protection on master

No. Even if 2 pull requests doesn't break build the merge commit of them may. Also I do experiments in `develop` sometimes.

cmake/projects/GLM/hunter.cmake
ruslo: Ok, feel free to add comment about it.

tbastos:
As you wish... but experiments are better done in temporary branches. The chance of having two simultaneous non-conflicting pull requests break each other is low, especially if you review each pull request well (something you should do anyway). Meanwhile having an eternal "develop" branch requires more work and will probably bring you (and/or contributors) problems every so often. Just mark my words ;)

ruslo:
> but experiments are better done in temporary branches

Okay, I agree

> The chance of having two simultaneous non-conflicting pull requests break each other is low, especially if you review each pull request well (something you should do anyway).

I'm using this flow in all my project, it may be true for Hunter but not for others. I simply don't like an idea of pushing untested code to master :)

> Meanwhile having an eternal "develop" branch requires more work and will probably bring you (and/or contributors) problems every so often.

What work? The only problem I'm experiencing is that users who contribute first time do pick `master` branch as a target often. Usually it's users fault since I've mention it in `CONTRIBUTING` (in this particular case it's my fault).

Anyway I will think about it. May be will change this in future.

tbastos:
It's not untested though, you seem to have fairly good automated tests. If you think they are not enough, just make them better. The more automated things are, the more hunter will evolve. Right now you have this manual step of testing things on "develop" and manually pushing to "master". This won't scale if you start to receive many pull requests. It's better for scale to trust your tests, merge to master as fast as possible and then fix any (rare) problem that may happen. You will get more users and more contributors, because people want to see their new packages and updates merged as soon as possible for their projects.

ruslo:
> It's not untested though

It is untested. Assume we have state A in develop and have two pull requests P1 and P2. Automated testing do test A + P1 and A + P2. When I do merge them I have state A + M where M is the merge commit of P1 and P2. M is **new** untested commit. In case of `develop` -> `master` flow we test P1, P2 and do test M. Then do merge M to `master`: `git merge --ff-only`. As you said it's quite unlikely case when problems may happens but it may.

> Right now you have this manual step of testing things on "develop"

What step is manual?

> manually pushing to "master"

It's true. I don't know how to fix it easily since we need to manage two services Travis CI and AppVeyor. So it should be some external service that check last state of branch `develop` and do merge if everything OK.

> This won't scale if you start to receive many pull requests

Why? What's the difference?

tbastos:
hehehe I know, my point is that you're worrying about something very unlikely, and "perfection is the enemy of progress". What's the worst that can happen if you merge a pull-request into master and something breaks? Someone will find the problem and fix it. The benefits of accepting contributions faster (including fixes) outweigh the benefit of "never breaking the master" (something you can't guarantee anyway).

BTW you can re-run the tests in a Pull Request after merging another PR. What doesn't scale is manually testing the "develop" branch and manually merging into "master". If you automate the merges it will scale, but personally I don't think the extra complexity is justified. If someone updates hunter and suddenly cannot build their projects anymore, they can just rollback to an earlier version :-)

ruslo:
As I said I'm using this flow in several projects so I'll be talking in general terms now.

> Someone will find the problem and fix it

Not exactly. Someone will clone my repo with default branch set to `master`, probably integrate it into some code and run compilation, then something will fail. Then user need to figure out that it's my project fault, report a bug... Or some user want to try my project, clone default `master` run and it fails, he decide that project is crap and move to another one :) Anyway fixing a bug may not be a trivial task so you have to revert. And so on. That's why having "blessed" repository is a good idea, it "just works".

> Removing barriers for contribution will also bring more contributors, obviously. The simpler things are, the better

What barriers? Set target branch to `develop` is a barrier? Developer open pull request - the rest is mine responsibility.

> BTW you can re-run the tests in a Pull Request after merging another PR

How? I'm not awared of it.

> What doesn't scale is manually testing the "develop" branch

Why manually? What does it mean?

> If you update hunter and suddenly cannot build your projects anymore, you can just rollback to an earlier version :-)

See above.

tbastos:
> Not exactly. Someone will clone my repo with default branch set to master, probably integrate it into some code and run compilation, then something will fail. Then user need to figure out that it's my project fault, report a bug... Or some user want to try my project, clone default master run and it fails, he decide that project is crap and move to another one :) Anyway fixing a bug may not be a trivial task so you have to revert. And so on. That's why having "blessed" repository is a good idea, it "just works".

If you have enough contributors that won't happen. Contributors will notice the problems and fix them. I would focus on attracting contributors and wouldn't be afraid of the scenario you described.

> What barriers? Set target branch to develop is a barrier? Developer open pull request - the rest is mine responsibility.

It's an extra complexity that you have to explain to every new contributor, and it gives the impression that their contributions won't be merged into master so soon. Contributors will assume that the develop branch may be unstable and you may take some time to merge their PRs into master.

Ignore what I said about re-running the tests of a PR. Even if you restart the tests, they won't include the new code merged upstream. The only solution is to ask the author of the other PRs to pull changes from upstream, and then that will re-run the tests. But I bet that would seldom be necessary in practice.

ruslo:
> Contributors will notice the problems and fix them, with very high probability (proportional to the number of current users/contributors you have)

May be I don't need testing at all in this case? Why do writing unit-tests, setting up CI environment and so on if you have "a lot of" users who will fix all the problems...

> I would focus on attracting contributors and wouldn't be afraid of the scenario you described

Why I can't do the both? :)

> It's an extra complexity that you have to explain to every new contributor

Every project have some contributing rules. If developer can't handle "set target branch to develop" there is a high probability that he will violate all other rules.

> and it gives the impression that their contributions won't be merged into master so soon. Contributors will assume that the develop branch may be unstable and you may take some time to merge their PRs into master

Why do they will assume so. It's pretty clear:. "...against the branch develop. After a successful build this branch will be merged to master..." as wiki page states.

To summarize: I will think about changing `delelop` -> `master` flow in future. At least for the Hunter it make a little sense. But I don't see a big deal about fixing it now. I think it's not a significant barrier.

Issue #332, opened 2016-02-22
Add 'benchmark' (gbenchmark) package


headupinclouds:
##

ruslo:
Note: CMake based

yowidin:
I plan to implement this package but `benchmark` is very generic name. Any suggestions on a better one? `GBenchmark`, `google-benchmark`?

ruslo:
as far as I can see user need to include `benchmark/benchmark.h`, so it's okay to name it `benchmark`

ruslo:
WIP: https://github.com/ruslo/hunter/pull/672

ruslo:
GitHub: https://github.com/google/benchmark

isaachier:
Did this in #1088. Can be close.

PR #429, opened 2016-06-07
google/benchmark package implementation

yowidin:
See #332


cmake/find/Findbenchmark.cmake
```diff
@@ -29,7 +29,7 @@ if(BENCHMARK_INCLUDE_DIR AND BENCHMARK_LIBRARY)
   add_library("benchmark::library" UNKNOWN IMPORTED)
   set_target_properties("benchmark::library" PROPERTIES
       INTERFACE_INCLUDE_DIRECTORIES "${BENCHMARK_INCLUDE_DIR}"
-			INTERFACE_COMPILE_OPTIONS -fno-omit-frame-pointer
+      INTERFACE_COMPILE_OPTIONS -fno-omit-frame-pointer
```
yowidin: Probably a bit [more ](https://msdn.microsoft.com/en-us/library/2kxx5t2c.aspx) logic required for the MSVS generator:

cmake/find/Findbenchmark.cmake
```diff
@@ -0,0 +1,44 @@
...
+
+find_package(Threads REQUIRED)
+
+if(BENCHMARK_INCLUDE_DIR AND BENCHMARK_LIBRARY)
+  set(benchmark_FOUND TRUE)
+
+  add_library("benchmark::library" UNKNOWN IMPORTED)
+  set_target_properties("benchmark::library" PROPERTIES
+      INTERFACE_INCLUDE_DIRECTORIES "${BENCHMARK_INCLUDE_DIR}"
+      INTERFACE_COMPILE_OPTIONS -fno-omit-frame-pointer
```
yowidin: Probably a bit [more ](https://msdn.microsoft.com/en-us/library/2kxx5t2c.aspx) logic required for the MSVS generator:

ruslo: when this code execuded library is already installed, so this flag will have no effect. instead this flag will be added to current project, which will be kind of surprising I think

ruslo: Also such file can be generated automatically, see https://github.com/ruslo/hunter/wiki/usr.adding.new.package#important

examples/benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+# Copyright (C) 2016, Dennis Sitelew
+cmake_minimum_required(VERSION 3.0)
+
+if(NOT CMAKE_BUILD_TYPE)
+  message(WARNING "You shouldn't benchmark a debug build")
```
ruslo: why not? also it shouldn't be checked such way for multi-config generators, see http://stackoverflow.com/a/24470998/2288008

cmake/find/Findbenchmark.cmake
yowidin: This is the sole purpose of this flag: to add it to the current project.

Reasoning behind it: 
In my experience most of the time benhmarking is done parallel with disassembly and code analysis. Without this flag disassembly analysis is hard to impossible.

examples/benchmark/CMakeLists.txt
yowidin: I can recommend you a good video on benchmarking: ttps://www.youtube.com/watch?v=nXaxk27zwlk

In short: it doesn't make any sense you benchmarking nothing since it isn't the code you will run on your production system.

As for detecting debug build it isn't required, just a hit for developers who will look at examples and try to implement their own benchmarks. Maybe I should add more rationale behind this message or delete it entirely.

dpantele: @yowidin 
Sometimes you'd like to debug the benchmarking code itself. I think that devs are responsible enough to use Release version after they have already made an effort to use google/benchmark.

yowidin: @dpantele It's just a waning and in example code, not library itself. It doesn't force you to use release build. I don't think it will have any negative impact on the development process.

dpantele: @yowidin Interestingly enough the benchmark itself also have such warning: https://github.com/google/benchmark#debug-vs-release , https://github.com/google/benchmark/blob/1b263fe6d906bb0854b84247f1b395bbacd3b88e/src/reporter.cc#L53

Ok, maybe devs indeed should make such warning to themselves.

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.488
* https://travis-ci.org/ingenue/hunter/builds/136381918

ruslo:
> https://travis-ci.org/ingenue/hunter/builds/136381918

most of the toolchains fail. you can do fixes or exclude broken from testing

yowidin:
@ruslo I will look into this problem.
Should I eventually integrate toolchain tests into this package (an error if we know that `behncmark` isn't supported on android/apple)?

ruslo:
@yowidin if you don't want to fix it or some platforms are not supported by design you can exclude everything now. in future toolchains can be reverted back step-by-step. you can send patch to `ingenue/hunter` branch `pkg.benchmark`. please add notes in case somebody want to do fixes. see for example: https://github.com/ingenue/hunter/pull/16/files in case there is no toolchains for some OS (like Windows I guess) you can use `dummy` toolchain

yowidin:
@ruslo Seems like `benchmark` has system introspection tests which it obviously unable to run. I picked `android-ndk-r10e-api-19-armeabi-v7a-neon` and supplied CMake with a cache file with those variables set and was able to run an example on my android tablet. Is where any way to provide these kind of files to the `benchamrk` package while building, based on the used toolchain? Any suggestions on how to solve this kind of problems?

ruslo:
> Any suggestions on how to solve this kind of problems?

@yowidin sounds like a bug/missing feature of `benchmark` package itself, i.e. should be fixed in `benchmark`, not in toolchains

yowidin:
@ruslo Have to disagree. It is common to see system introspection tests in complex projects and the project authors cannot prepare tests for all the possible platforms. So the user of the library have to feed CMake with a valid cache file containing these test results. This is exactly what I have done in my android test. But I did it manually and only for the standalone `benchmark` library. See [this](http://www.vtk.org/Wiki/CMake_Cross_Compiling#System_introspection) link for more information about this issue.

Now knowing  that some packages may require these kind of cache files I assume `hunter` needs some kind of mechanism that provides library implementer a way to feed CMake with introspection test results depending on the used toolchain. Or as alternative give users a possibility to provide these cache files themselves. In our case since tests are build with the `polly` I guess `polly` requires a way to pass cache files to the underlying `hunter` libraries.

ruslo:
> So the user of the library have to feed CMake with a valid cache file containing these test results

sounds like a workaround, this is not how CMake usually work

> the project authors cannot prepare tests for all the possible platforms

they should, for all the platforms they promise to support

> See this link for more information about this issue

You don't need the cache file, just pass missing variables using `-D` or `set(... CACHE ...)`. In this case it is possible to add them to Polly.

Just curious what is the official way to build it for Android?

yowidin:
> Just curious what is the official way to build it for Android?

This library doesn't mention android anywhere explicitly but still builds in case if cache files provided. or in built manually. 

> sounds like a workaround, this is not how CMake usually work

That's  why I shared a link where it stated explicitly, that it is the expected way to cross-compile a library what not aware about a given platform :)

> You don't need the cache file, just pass missing variables using -D or set(... CACHE ...). In this case it is possible to add them to Polly.

I tested it before opening this issue and tried all the possible ways to make this library compile for android. CMake expects this kind of file. 

We could move this discussion in chat and talk a bit more about it directly.

ruslo:
> We could move this discussion in chat and talk a bit more about it directly

you can join gitter or we can use tox: https://docs.hunter.sh/en/latest/contacts.html

ruslo:
Seems abandoned. Feel free to reopen on updates.

Issue #428, opened 2016-06-06
Autotools bootsrap option.


Cyberunner23:
Some libraries that use autotools as a build system use some sort of bootstrap command such as `./autogen.sh` or `./bootstrap` before doing the usual
    
    ./configure
    make
    make install

(I'm looking at you libsodium)
So I was wondering if it would be possible to add some sort of option where one can set such boostrap command and make it run just before the ./configure. I was thinking of putting it in [hunter_autotools_project.cmake](https://github.com/ruslo/hunter/blob/master/cmake/modules/hunter_autotools_project.cmake) so we can easily make a schema without rederiving hunter_autotools_project.cmake for each bootstrap command.

ruslo:
> if it would be possible to add some sort of option where one can set such boostrap command

I guess it's possible. You can add option `BOOTSTRAP` to `hunter_autotools_project`. @pretyman what do you think?

pretyman:
Sorry, late for the party since I've been without internet. Yeah it makes total sense to be there 

👍

PR #430, opened 2016-06-07
Added option to have a bootstrap command in autotools based schemas.

Cyberunner23:
Includes a schema for projects using ./autogen.sh (Most common of the bootstrap commands as far as I know.) Discussion: #428

### DIFF
--- a/cmake/modules/hunter_autotools_project.cmake
+++ b/cmake/modules/hunter_autotools_project.cmake
@@ -271,7 +271,8 @@ function(hunter_autotools_project target_name)
     list(APPEND configure_command "PKG_CONFIG_PATH=${d1}:${d2}")
   endif()
 
-  if(NOT ("${PARAM_BOOTSTRAP}" STREQUAL ""))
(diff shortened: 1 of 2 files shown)


cmake/modules/hunter_autotools_project.cmake
```diff
@@ -267,6 +271,10 @@ function(hunter_autotools_project target_name)
     list(APPEND configure_command "PKG_CONFIG_PATH=${d1}:${d2}")
   endif()
 
+  if(NOT ("${PARAM_BOOTSTRAP}" STREQUAL ""))
```
ruslo: please use `string(COMPARE NOTEQUAL`, see http://0.readthedocs.io/en/latest/cmake.html#strequal

ruslo:
@Cyberunner23 thanks for update, just for your information I'm not receiving notifications in such cases, you have to do it manually, see http://0.readthedocs.io/en/latest/git.html#notify

Issue #248, opened 2015-10-02
Add Curl


jneidlinger:
It would be great to have libcurl support in Hunter!

https://github.com/bagder/curl

I love everything Hunter is trying to accomplish, thanks for all the hard work!

ruslo:
@jneidlinger What platform are you interested in? The last time I've tried to build Curl universal iOS library I've hit some troubles. Mainly because it generates different headers for different architectures. With this exception it should "just work" since it has a CMake build. Can you try: https://github.com/ruslo/hunter/wiki/usr.adding.new.package ?

jneidlinger:
Well, in a perfect world, Win/Lin/OSX + Android/IOS/WinRT

I've only ever built seperate libcurl iOS architectures and lipo'd them myself. Other's go the route of using a different build system, for example:

https://github.com/biasedbit/curl-ios-build-scripts

will build curl as a universal library for you

ruslo:
> The last time I've tried to build Curl universal iOS library I've hit some troubles. Mainly because it generates different headers for different architectures.

Here is the place to workaround Curl header issue that I was talking about:
https://github.com/gcesarmza/curl-android-ios/blob/503d7ef85a9650503a0620993c066576509f5d7e/curl-compile-scripts/build_iOS.sh#L89

ruslo:
> for example:
https://github.com/biasedbit/curl-ios-build-scripts
will build curl as a universal library for you

this script doesn't handle header issue, here is a bug: https://github.com/biasedbit/curl-ios-build-scripts/issues/24

Quote:
```
This is because curlbuild.h differs for the 32 bit and 64 bit versions,
but the build script assumes that they are the same, and uses just
the 32 bit version
```

jneidlinger:
Good find

ruslo:
> Well, in a perfect world, Win/Lin/OSX + Android/IOS/WinRT

I've never used WinRT, if it's supported by CMake should work too though. The rest platforms should work fine.

ruslo:
Added by #511. Not all toolchains are working, see testing branches for available list:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.914
* https://travis-ci.org/ingenue/hunter/builds/157820991

PR #456, opened 2016-07-11
Added CURL build

s-brain:
Added curl, static and shared build works fine on widows depending on hunter OpenSSL,
cannot seem to get static build in linux to depend on OpenSSL properly, but i have no real experience with auto tools, some guidance would be nice


cmake/projects/CURL/hunter.cmake
```diff
@@ -0,0 +1,43 @@
...
+    url_sha1_curl_windows
+)
+
+
+
+hunter_cacheable(CURL)
+
+hunter_download(PACKAGE_NAME CURL
+
+    PACKAGE_DEPENDS_ON OpenSSL ZLIB)
```
ruslo: Not needed

ruslo:
CURL is CMake based: https://github.com/curl/curl/blob/master/CMakeLists.txt
It means you don't need to use custom schemes. You just need to hunterize it: add hunter_add_packages({OpenSSL,ZLIB}). Also it would be nice to add a code that will install automatically generated CURLConfig.cmake instead of FindCURL.cmake (it doesn't support custom build types for example).

Also for iOS there is a problem with installing universal libraries because headers for 64 bit differes from headers of 32 bit architecture (see https://github.com/ruslo/hunter/issues/248).

s-brain:
Curl may look like a cmake enabled project but after implementing your suggestions i have come to notice it is appears not to be complete. 

one of the first messages is  

> CMake Warning at CMakeLists.txt:57 (message):
> 
>       the curl cmake build system is poorly maintained.  Be aware
> 

And also it does not build a static library correctly 

I cannot do IOS testing as i have not got a MAC so am unsure how to test this.

Also all the versioning and building seems to be fully incorporated into the tarball releases, so a lot of work bypassing this mechanism would have to be done to replicate this.
I would feel more comfortable following the offical releases.

I have created a package http://github.com/s-brain/curl  which encompasses my changes so far, but cannot decide how to go from here as the cmake version needs much more care and attention to get right, and if it is not currentlly supported fully by the CURL devs, i don't know how much to trust it,

ruslo:
> a lot of work bypassing this mechanism would have to be done to replicate this

Source are on GitHub, just fork and release. Push fixes to upstream.

> I would feel more comfortable following the offical releases

This look like a nice way to go only at the first glance, in fact this approach is quite unmaintainable. Just think about build types: Release, Debug, MinSizeRel, RelWithDebInfo. What about custom build types? Shared/static libraries, Android build, iOS build, iOS + fusing artifacts into universal libraries, on windows build with MinGW and so on. Custom scheme for Boost is one of the first scheme that was added and it's still not complete and I think it will be complete only when Boost will move to CMake. Custom schemes is something that need to be used when we really have no choice (like Qt) but not with the project that already have CMakeLists.txt

> I have created a package http://github.com/s-brain/curl which encompasses my changes so far

I have a fork already: https://github.com/hunter-packages/curl I've sent you an invitation, you can push changes here after you accept it. I think it make sense to fix the issues and push them upstream, then keep hunterized fork and use releases of it. Please keep branch `master` clean so we can always do `git merge --ff-only` from upstream.

s-brain:
Ok curl hunter-package is updated, pull  updated to enable cmake-scheme.
There is one issue i see which i don't know the best way to resolve, and that is making the curl find_package automatically run the ZLIB find package, as i currently need to have both find packages in my client CMakeList.txt file when building curl statically, even though the client should not need to care about it.

cmake/projects/CURL/hunter.cmake
```diff
@@ -0,0 +1,56 @@
...
+    VERSION
+    "7.49.1-DEV"
+    URL
+    "https://github.com/hunter-packages/curl/archive/hunter-7.49.1-v0.zip"
+    SHA1
+    fd72a153ca60fe85b57a2ac4eadedcfb38c3066d
+)
+
+
+if(BUILD_SHARED_LIBS)
```
ruslo: Should not be here. `BUILD_SHARED_LIBS` is for controlling **current** project, not 3rd party. CURL should respect default CMake feature: `BUILD_SHARED_LIBS`. See: https://github.com/ruslo/hunter/wiki/usr.adding.new.package#default-behaviour

cmake/projects/CURL/hunter.cmake
```diff
@@ -0,0 +1,56 @@
...
+
+
+
+include(hunter_add_version)
+include(hunter_cacheable)
+include(hunter_download)
+include(hunter_pick_scheme)
+
+
+hunter_add_package(OpenSSL)
```
ruslo: Not needed here. This should be done in `CMakeLists.txt` of CURL.

cmake/projects/CURL/hunter.cmake
```diff
@@ -0,0 +1,56 @@
...
+        CURL_STATICLIB=${BUILD_CURL_STATIC}
+ )
+
+
+
+# Pick a download scheme
+hunter_pick_scheme(DEFAULT url_sha1_cmake)
+hunter_cacheable(CURL)
+hunter_download(PACKAGE_NAME CURL
+                PACKAGE_DEPENDS_ON OpenSSL ZLIB LibSSH2)
```
ruslo: `PACKAGE_DEPENDS_ON` not needed

examples/CURL/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+
+# Emulate HunterGate:
+# * https://github.com/hunter-packages/gate
+include("../common.cmake")
+
+project(download-curl)
+
+hunter_add_package(CURL)
+find_package(CURL REQUIRED)
+find_package(ZLIB REQUIRED)
```
ruslo: Dependency of CURL should be loaded in `CURLConfig.cmake` automatically so users don't have to take care of it manually.

s-brain:
Addressed your comments, and also improved Curl CMakeLists.txt alongside your comments there too.
Thanks for the review, I think I have learned more about CMake trying to add to hunter than I had previously just using CMake, there are so many "hidden" gems in your code compared to documentation.

Thanks for making me persevere with updating CURLS CMakeLists.txt file I will consolidated the changes and see if CURL will accept them too, though I feel that your review will be better than theirs as i think the CMake file is a second class citizen over there.

ruslo:
Thanks. I've tried to run it on my Linux machine and it's failed because `hunter_add_package(LibSSH2)` used somewhere. Should be `hunter_add_package(Libssh2)`: https://github.com/ruslo/hunter/tree/master/cmake/projects/Libssh2

examples/CURL/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+cmake_minimum_required(VERSION 3.0)
+
+# Emulate HunterGate:
+# * https://github.com/hunter-packages/gate
+include("../common.cmake")
+
+project(download-curl)
+
+hunter_add_package(CURL)
+find_package(CURL REQUIRED)
```
ruslo: We have to use `find_package(CURL CONFIG REQUIRED)` to load `CURLConfig.cmake` instead of `FindCURL.cmake`

examples/CURL/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+
+
+
+message("CURL_INCLUDE_DIRS: ${CURL_INCLUDE_DIRS}")
+message("CURL_LIBRARIES: ${CURL_LIBRARIES}")
+message("CURL_FOUND: ${CURL_FOUND}")
+message("CURL_VERSION_STRING: ${CURL_VERSION_STRING}")
```
ruslo: This variables will not be defined if `find_package(CURL CONFIG REQUIRED)` used. Small overview:
* include directories and path to library stored in imported target CURL::libcurl
* CURL_FOUND not needed since we use REQUIRED (will crash if not found)
* version can be set by CMake modules, see: https://github.com/forexample/package-example/blob/cf2ea1d6a209fb9eca2ab83fdd0ac15fe4d3e807/Foo/CMakeLists.txt#L86

s-brain:
Arrgh, I have been trying to track down why it was running on one machine and not the other and that CONFIG was it. makes so much more sense now i read up on it,

ruslo:
Testing:
* https://travis-ci.org/ingenue/hunter/builds/145320620
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.675

However Linux tests will fail, see for details: https://github.com/hunter-packages/libssh2/issues/2

ruslo:
> However Linux tests will fail

Well they not. Because Libssh2 not found and is not used:
```
CMake Warning at CMakeLists.txt:496 (find_package):
  By not providing "FindLibSSH2.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "LibSSH2", but
  CMake did not find one.

  Could not find a package configuration file provided by "LibSSH2" with any
  of the following names:

    LibSSH2Config.cmake
    libssh2-config.cmake
```

I think this is a problem with case, should be `find_package(Libssh2 CONFIG REQUIRED)`.

ruslo:
> Testing: https://travis-ci.org/ingenue/hunter/builds/145320620

* `analyze` failed because of libssh2 dependency
* `sanitize-thread` fails for some strange reason...
* `ios` and `android` fails too

feel free to exclude them (send patch to [pkg.curl](https://github.com/ingenue/hunter/tree/pkg.curl) branch, comment them and leave a links to build, [example](https://github.com/ingenue/hunter/pull/22/files)) or investigate fix

s-brain:
Had been looking at that, and the android build seems to be failing because of incorrect libssh captilisation in CURL, ( they provided their own findLibSSH2.cmake module instead of using the libssh2-Config module) 
I am attempting to fix that now.

My windows box hides these errors.

s-brain:
The more i look at the CURL CMakeLists.txt file the more errors i see.
do you know if  CMAKE_REQUIRED_LIBRARIES is a standard value, as i see it is treated as a list in places then overwritten in others.

ruslo:
> do you know if CMAKE_REQUIRED_LIBRARIES is a standard value, as i see it is treated as a list in places then overwritten in others.

Not sure I understand what you mean. This variable is a list of libraries used in tests like [CheckCSourceCompiles](https://cmake.org/cmake/help/latest/module/CheckCSourceCompiles.html).

s-brain:
i have removed the libssh2 from the default config, as it would not be built if no directive given anyway.
having said that i improved the curl code to be better at handling it anyway, so if asked for it would be built

ruslo:
Please take a look at CURLTargets-*.cmake files. They should not contain full paths to OpenSSL library. It means `${OPENSSL_LIBRARIES}` used instead of `OpenSSL::Crypto`. We will have problems with binary cache. I've updated wiki page: https://github.com/ruslo/hunter/wiki/pkg.openssl

ruslo:
> It means ${OPENSSL_LIBRARIES} used instead of OpenSSL::Crypto

Actually I figured out that it might be a Hunter flaw since we must have [backward compatibility](https://docs.hunter.sh/en/latest/overview/compatibility.html). I'll take a look.

s-brain:
I have updated the CURL CMAKELists.txt to reference the OpenSSL::Crypto library if found, and fallback otherwise, I will get to updating hunter later today to point to it

ruslo:
> I'll take a look

My update breaks `try_compile` function. I think it was not desiged to use imported targets. I'll keep current behaviour.

xsacha:
What's the status on this @ruslo ?
What needs to be done to make this merge-worthy?

ruslo:
> What's the status on this @ruslo ?

I'm not working on it, you should ask @s-brain.

> What needs to be done to make this merge-worthy?

Investigate and fix issues CURL have while building with CMake. If at least one platform will work I'm okay with merging it. We can drop other platforms and fix them one-by-one in future.

xsacha:
I think the history of lowercase folder names was causing issues that made it fail to build (at least on Windows).
I squashed the history and pushed here again.
It is building fine on Linux and Windows. Can you please test again @ruslo ?

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.900
* https://travis-ci.org/ingenue/hunter/builds/157550165

xsacha:
Looks like issues with clang (sanitize and analyze) and cross-compile (TRY_RUN).

Thanks.

ruslo:
A lot of toolchains are working, if you exclude broken (pull request to [pkg.curl](https://github.com/ingenue/hunter/tree/pkg.curl)) toolchains I'm okay with merge.

xsacha:
Great! Done that.
Thanks.

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.908
* https://travis-ci.org/ingenue/hunter/builds/157586146

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.909
* https://travis-ci.org/ingenue/hunter/builds/157616611

ruslo:
> Uploading

Broken.

CURL is not relocatable in fact:

```
> wget https://github.com/ingenue/hunter-cache/releases/download/cache/98245bd8fc295548de236f059bf3a6257d5e2a04.tar.bz2
> tar xf 98245bd8fc295548de236f059bf3a6257d5e2a04.tar.bz2
> grep ssl lib/cmake/CURL/CURLTargets-debug.cmake
  IMPORTED_LINK_INTERFACE_LIBRARIES_DEBUG "winmm;ws2_32;C:/projects/hunter/_testing/Hunter/_Base/0e4d657/fb5f104/b66d985/Install/lib/ssleay32.lib;C:/projects/hunter/_testing/Hunter/_Base/0e4d657/fb5f104/b66d985/Install/lib/libeay32.lib;ZLIB::zlib"
```
As you can see there is a full path to OpenSSL library. It should be OpenSSL::SSL instead (like ZLIB::zlib). See for details: https://github.com/ruslo/hunter/wiki/example.find_package.config

xsacha:
Sorry, the commit that adds this is in hunter-packages/curl:master but wasn't in the tag that is being used:

Issue #462, opened 2016-07-18
autotools default paths 


pretyman:
https://github.com/ruslo/hunter/issues/411#issuecomment-220546303

> I remember that I was against modifying PATH and PKG_CONFIG_PATH by default, that's why we have MODIFY_PATH and MODIFY_PKG_CONFIG. I guess I'm ready to change my mind now :) Adding:
>

> - -I@HUNTER_INSTALL_PREFIX@/include

>- -L@HUNTER_INSTALL_PREFIX@/lib
> - PKG_CONFIG_PATH=@HUNTER_INSTALL_PREFIX@/lib/pkgconfig:@HUNTER_INSTALL_PREFIX@/share/pkgconfig
> - PATH=@HUNTER_INSTALL_PREFIX@/bin:${default_path}

>seems consistent to me, feel free to remove MODIFY_PATH/MODIFY_PKG_CONFIG. We can add options for disabling of modifications of anything in case there will be some issues about it.

pretyman:
> We can add options for disabling of modifications of anything in case there will be some issues about it.

I've been thinking, would it make any sense to have the option to disable this?

ruslo:
> would it make any sense to have the option to disable this?

We can have defaults and **if** there will be problems in future **then** we can add such options. I think we can move without them for now.

PR #464, opened 2016-07-20
Autotools defaults with <root-id>

pretyman:
Treats issue https://github.com/ruslo/hunter/issues/462

Tested `examples/x11` and `examples/odb-sqlite` under Linux successfully.

Test failed for me for `examples/Avahi`, could you please try it on your end? It fails even before the change of this pull request:

```
configure: WARNING: unrecognized options: --disable-shared, --enable-static
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /usr/bin/mkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking for perl... /usr/bin/perl
checking for perl >= 5.8.1... 5.22.1
configure: error: XML::Parser perl module is required for intltool
checking for XML::Parser... CMakeFiles/intltool.dir/build.make:111: recipe for target 'intltool-prefix/src/intltool-stamp/intltool-configure' failed
gmake[2]: *** [intltool-prefix/src/intltool-stamp/intltool-configure] Error 2
gmake[1]: *** [CMakeFiles/intltool.dir/all] Error 2
gmake: *** [all] Error 2
gmake[2]: Leaving directory '/home/builder/.hunter/_Base/xxxxxxx/811c9df/d17a7ab/Build/intltool/Build'
CMakeFiles/Makefile2:70: recipe for target 'CMakeFiles/intltool.dir/all' failed
gmake[1]: Leaving directory '/home/builder/.hunter/_Base/xxxxxxx/811c9df/d17a7ab/Build/intltool/Build'
Makefile:86: recipe for target 'all' failed
```


pretyman:
There is quite a lot of boilerplate for  `hunter_autotools_project("@HUNTER_EP_NAME@" ...` in custom schemes that exist just to express transitive dependencies.

Scheme x11.cmake.in mainly exists just because of https://github.com/pretyman/hunter/blob/2f5aa70797e10396cb64467d547751af38524fea/cmake/projects/x11/schemes/x11.cmake.in#L20

```CMake
hunter_add_package(xproto)
hunter_add_package(xextproto)
hunter_add_package(xtrans)
hunter_add_package(xcb)
hunter_add_package(kbproto)
hunter_add_package(inputproto)
```

I was wondering if these could be passed by args through

```CMake
hunter_pick_scheme(DEFAULT url_sha1_autotools
    ARGS
      DEPENDS_ON xproto xextproto xtrans xcb kbproto inputproto
)
```

Then I could read them in a foreach in the default autotools scheme `url_sha1_autotools` and `hunter_add_package` each of them.

We could kill a lot of files to maintain :-)

ruslo:
> It fails even before the change of this pull request

Yes, I know: https://github.com/ruslo/hunter/pull/237#issuecomment-193709117. You can ignore it. Feel free to investigate and fix of course :)

cmake/projects/drm/hunter.cmake
```diff
@@ -26,7 +26,7 @@ hunter_pick_scheme(DEFAULT drm)
 hunter_cacheable(drm)
 hunter_download(
     PACKAGE_NAME drm
-    PACKAGE_INTERNAL_DEPS_ID "1"
+    PACKAGE_INTERNAL_DEPS_ID "2"
```
ruslo: Actually I think we don't need this one here. Your patch doesn't fix broken packages, doesn't broke working packages, in general it doesn't affect how packages built - hence we don't need to change PACKAGE_INTERNAL_DEPS_ID.

ruslo:
> I was wondering if these could be passed by args through

What about CMAKE_ARGS mechanism?

pretyman:
> hence we don't need to change PACKAGE_INTERNAL_DEPS_ID.

Ok, updated ;-)

> What about CMAKE_ARGS mechanism?

Fine with me, however might now be very intuitive for others because they are not really "CMake"  arguments. I believe this declarative way for informing dependency of packages can go into some philosophical issues: what if the dependencies are different for different versions of a same package? 

But we can treat them as the problem arises.

> Yes, I know: #237 (comment). You can ignore it. Feel free to investigate and fix of course :)

If I ever need this package, I'll try ;-)

ruslo:
> however might now be very intuitive for others because they are not really "CMake" arguments

They are CMake arguments for the build scheme project. We are already using them for controlling Boost dependencies: https://github.com/ruslo/hunter/wiki/pkg.boost#cmake-options.

> what if the dependencies are different for different versions of a same package?

Yes this doesn't scale well. What does scale is CMake-based project with controlling info about dependencies internally. Luckily this is exactly how Hunter works :) I'm just saying that autotools projects will have such issue **with any package manager**. It's not about Hunter or CMake. Also it doesn't scale with scheme approach anyway.

ruslo:
Testing SQLite3:
* https://travis-ci.org/ingenue/hunter/builds/146200068

ruslo:
> Yes this doesn't scale well

Just to clarify it doesn't scale but it's possible. You have to maintain this manually by using `if(<version>)` in `hunter.cmake`:
```cmake
if(<version> VERSION_LESS 1.0)
  # cmake args for < 1.0
else()
  # cmake args for 1.0+
endif()
```

pretyman:
> They are CMake arguments for the build scheme project. We are already using them for controlling Boost dependencies: https://github.com/ruslo/hunter/wiki/pkg.boost#cmake-options.

Yeah, that's cool, I guess we can treat that in another issue to not mix things up. 

It is just that package `glfw` has all these transitive dependencies... I wonder how does transitive dependencies work with caching, since we `hunter_add_package` in the scheme, if the package is cached, then the scheme won't run, correct? Does this mean that the transitive dependencies won't be available? I'm not sure if it is worth clarifying this in wiki/docs because it is such internal behavior, but I could not find them in the wiki by searching for "cache".

ruslo:
> I wonder how does transitive dependencies work with caching

Should work fine, let me know if you have any issues. We do save dependencies information in cache.

pretyman:
> We do save dependencies information in cache.

Thanks, I was just curious 👍

PR #465, opened 2016-07-22
Autotools: use CMAKE_ARGS DEPENDS_ON_PACKAGES to pass dependencies to url_sha1_autotools default scheme

pretyman:
Use CMAKE_ARGS from `package/hunter.cmake` to pass a list of DEPENDS_ON_PACKAGES to default scheme`scheme/url_sha1_autotools`. Default scheme reads the list and `hunter_add_package` each of them.

I also updated package `xproto`, adding a dependency for `xorg-macros`, since it was looking for it during `./configure` stage [although it still compiled successfully when not found]. So I had to bump the `PACKAGE_INTERNAL_DEPS_ID` for that one.

Tested packages `x11` and `odb-sqlite`.

Sorry for so many packages changed, I understand it is a little painful to review and test :-)


pretyman:
Oh, I also had some problems with using syntax such as:
```CMake
hunter_cmake_args(... 
    CMAKE_ARGS
      LIST="value1;value2;value4"
)
```

Because https://github.com/ruslo/hunter/blob/master/cmake/modules/hunter_create_args_file.cmake would write an `args.cmake` file with something like:
```
set("LIST" ""value1;value2;value3"" ...)
```

which would make the `${LIST}` be set to an empty string `""`. Is this the intended behavior?

cmake/projects/drm/hunter.cmake
```diff
@@ -22,7 +23,17 @@ hunter_add_version(
 )
 
 hunter_configuration_types(drm CONFIGURATION_TYPES Release)
-hunter_pick_scheme(DEFAULT drm)
+hunter_pick_scheme(DEFAULT url_sha1_autotools)
+list(APPEND drm_dependencies
```
ruslo: This file can be processed several times so it should be `set(drm_dependencies ...)`. Also use names like `_drm_dependencies` since such module included into user's space.

ruslo:
> LIST="value1;value2;value4"

in such argument `"` will be treated as part of the string, just use `"LIST=value1;value2;value4"`. Also you don't need to use quotes here, can be simply `LIST=value1;value2;value4`.

cmake/projects/drm/hunter.cmake
pretyman: > This file can be processed several times

Oh, ok, I thought you removed include guards because this file would only be processed once. I guess I misinterpreted that change :-)

ruslo:
http://cgold.readthedocs.io/en/latest/tutorials/variables/regular.html#quotes

cmake/projects/drm/hunter.cmake
ruslo: Actually it's vice versa: I did remove guards because it can be processed several times. It's a part of so-called "CACHE RUN" when we don't download package but looking up for cached entry. We have to calculate version and arguments before.

pretyman:
> http://cgold.readthedocs.io/en/latest/tutorials/variables/regular.html#quotes

Cool, thanks

pretyman:
Update: changed from `list(APPEND ` to `set(` as per your comments

ruslo:
Brilliant idea, thanks!

pretyman:
> Brilliant idea, thanks!

I'm thinking if it would be a good choice to expand this idea of passing the arguments to `hunter_autotools_project(... EXTRA_FLAGS ..` from the `package/hunter.cmake`.

Right now I need pass arguments like `EXTRA_FLAGS --disable-feature --enable-other-feature` and I'm afraid of the boiler plate once again....

What do you think?

ruslo:
> What do you think?

Make sense. Also this will allow user to control the features. For now they are "freezed" in schemes.

pretyman:
> For now they are "freezed" in schemes

"frozen" ;-)

Ok, I'll look into it

Issue #218, opened 2015-08-10
geos ios, relocatable .la files


ruslo:
Currently failing with error:

```
CMake Error at autotools-merge-lipo.cmake:234 (message):
  lipo execution failed: fatal error: lipo: can't figure out the architecture
  type of:
  /.../Build/multi-arch/arm64/bin/geos-config
```
##

ruslo:
@pretyman About this particular issue: I think this script is not needed - Findgeos module doing exactly the same but in CMake-friendly fashion. I guess we can ignore it like `*.la` files.

In general: I think we can add `EXCLUDE` parameter here, e.g. `hunter_autotools_project(... EXCLUDE_FILES "bin/geos-config" ...)`.

pretyman:
I don't think it is wise to temper with the build like this, removing files from it. I think we should make an effort to reproduce the full build inside `<root-id>/Install`.

The `geos-config` is in fact needed by `spatialite`, and won't compile without it, I have tried.

The `*.la` files in `Qt` are actually needed by the `odb` ORM framework library `odb-qt` at `./configure` execution time - this means it runs libtool during the ./configure execution. This is because the `libQt5Core.la` file has a field `dependency_libs=...` which contains its dependencies.

By ripping files out of the build, we increase the chance that projects which usually compiled together, will end up not compiling. This could bring frustration to users who want to integrate new packages: "I always compiled package `Y` with package `X` in my system, but Hunter can't compile the two together for some reason...".  

The more the build is untouched inside the `<root-id>/Install`, the higher the chances that other dependents projects will succeed building as well.

I think you have to re-evaluate this tactic of removing files from the build in the overall strategy of your project Hunter, based on who you want to target and how you want their interaction with your package manager should be. 

From the cache point of view, I re-iterate my suggestion that we could look into all text files in the resulting build searching the `@HUNTER_PACKAGE_INSTALL_DIR@`, keeping the files into a list which will be saved and replace the path with the placeholder `@HUNTER_PACKAGE_INSTALL_DIR@`. 

When hitting the cache, after uncompression is done, we look at the files in the saved list and replace the placeholder with its final destination. It would be nice to have a way of telling if a file is actually text, so we don't run the risk of breaking any binary files.

At the end, it is a design decision, and it is your call :-)

ruslo:
> I think we should make an effort to reproduce the full build inside `<root-id>/Install`

I guess we have quite a few choices here

[...]

PR #469, opened 2016-07-24
Enable cache on odb packages

pretyman:


ruslo:
`odb-boost` doesn't looks good:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.288
* https://travis-ci.org/ingenue/hunter/builds/116995531

We need to make it "green" first. Feel free to exclude broken tests.

pretyman:
All AppVeyor tests failed because `odb` has only `autotools` integration and VisualStudio does not integrate with `autotools`.

Package `odb` has also VisualStudio solution/projects which I would guess it can be compiled by `msbuild`. 

Would be cool if we had some generic `msbuild` integration, following for example the scheme contributed by @Cyberunner23 (under https://github.com/ruslo/hunter/blob/master/cmake/projects/libsodium/schemes/url_sha1_libsodium_msbuild.cmake.in). Something like we have for `url_sha1_autotools`.

I'll leave this pull request open for a bit, so when I setup a windows box, I'll see if this default `msbuild` scheme can be done. I just don't want to disable the windows tests and have to enable them again later on as I have limited time.

Or... we could open an issue and wait for @Cyberunner23 to volunteer ;-)

Cyberunner23:
> I'll see if this default `msbuild` scheme can be done.

It could be done, however we'll need to do something like the `hunter_autotools_project` and add a bunch more options given how msbuild is much and perhaps annoyingly more configurable. For example we'll need to add an `INSTALL_COMMAND` argument as some msbuild buildsystems dont copy the headers to the install directory... Perhaps have arguments for the `Release`, `ReleaseDLL`, `Debug` and `DebugDLL` because apparently they can be renamed, removed, have other targets and so on. I mean its possible but we have to think of all the possible tweaks we need to be able to handle.

ruslo:
> I mean its possible but we have to think of all the possible tweaks we need to be able to handle.

We can start with what we need now and add features step-by-step.

pretyman:
To be solved first: https://github.com/ruslo/hunter/issues/218

pretyman:
Actually, can we merge this? It will help to do some public relations here: [[odb-users] Custom ODB compiler build rule on OSX](http://codesynthesis.com/pipermail/odb-users/2016-August/003418.html)

ruslo:
> Actually, can we merge this?

-

> Solve #218 before merging this

Bug is still open, I guess some toolchains will not work. Is it okay?

pretyman:
I've already commented out ios toolchain, all tests should pass 👍

ruslo:
odb-boost:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.848
* https://travis-ci.org/ingenue/hunter/builds/155248453

ruslo:
odb.mysql:
* https://travis-ci.org/ingenue/hunter/builds/155248453
* https://travis-ci.org/ingenue/hunter/builds/155249479

ruslo:
odb.pgsql:
* https://travis-ci.org/ingenue/hunter/builds/155250502

ruslo:
odb.sqlite:
* https://travis-ci.org/ingenue/hunter/builds/155250855

ruslo:
odb:
* https://travis-ci.org/ingenue/hunter/builds/155251917

ruslo:
Let me know if you think that we should do any other tests.

ruslo:
Uploading: https://github.com/ingenue/hunter/branches/all?utf8=%E2%9C%93&query=upload.odb

ruslo:
> Uploading

Failed: https://travis-ci.org/ingenue/hunter/jobs/155255076

There is a test after upload that all packages can be downloaded from server on next build. Such test is not passed because `odb-compiler` is missing. Need to make it cacheable.

pretyman:
> Need to make it cacheable.

Hum, in the past I recall reading through the code somewhere that when using scheme `url_sha1_unpack_install` it would be implicitly cacheable, but obviously something has changed.

Do I just add `hunter_cacheable(odb-compiler)` and *that's it*?

ruslo:
> Do I just add hunter_cacheable(odb-compiler) and that's it?

No no need, I think this is a Hunter issue. I'll take a look.

ruslo:
While testing I've noticed that generated files put into source-tree instead of build-tree. Consider following [out-of-source](http://cgold.readthedocs.io/en/latest/tutorials/out-of-source.html#other-notes) approach and keep all autogenerated stuff in temporary folders.

ruslo:
> I'll take a look

Fix: https://github.com/ruslo/hunter/commit/a28656387c056c160ba411718e27d2ce8c64ecb1?diff=unified

Testing:
* https://github.com/ingenue/hunter/branches/all?utf8=%E2%9C%93&query=upload.odb

Details:
* https://docs.hunter.sh/en/latest/reference/internal-variables/hunter_package_scheme.html

ruslo:
`odb-mysql` failed:
* https://travis-ci.org/ingenue/hunter/builds/155893928

Package `MySQL-client` should be cacheable.

ruslo:
> odb-mysql failed

Testing new updates: https://travis-ci.org/ingenue/hunter/builds/156127016

ruslo:
Now OSX is broken: https://travis-ci.org/ingenue/hunter/builds/156129547

```
configure: error: in `/Users/travis/build/ingenue/hunter/_testing/Hunter/_Base/dc1cb1f/cef4daf/6217056/Build/odb-mysql/Source':
configure: error: cannot run C compiled programs.
If you meant to cross compile, use `--host'.
See `config.log' for more details
make[2]: *** [odb-mysql-prefix/src/odb-mysql-stamp/odb-mysql-configure] Error 1
make[1]: *** [CMakeFiles/odb-mysql.dir/all] Error 2
make: *** [all] Error 2
```

since the last change was about cacheable version I can guess full path stuck somewhere (RPATH? *.pc?)

PR #480, opened 2016-08-03
Added package glfw

pretyman:
Not ready yet. 

Facing problem when building `glfw-Release`. Извините, голова плохо работает, can't figure it out myself.

Building with CMake 3.5 from your docker environment gives: 
```
-- Using X11 for window creation
-- [hunter *** DEBUG *** 2016-07-26T05:11:14] Single-configuration generator
-- [hunter *** DEBUG *** 2016-07-26T05:11:14] List of cache servers:
-- [hunter *** DEBUG *** 2016-07-26T05:11:14]   * https://github.com/ingenue/hunter-cache
-- [hunter *** DEBUG *** 2016-07-26T05:11:14] Settings (finalize):
-- [hunter *** DEBUG *** 2016-07-26T05:11:14]   HunterGate done (YES)
-- [hunter *** DEBUG *** 2016-07-26T05:11:14]   Cache init (YES)
-- [hunter *** DEBUG *** 2016-07-26T05:11:14] Reuse cached values
-- [hunter] HUNTER_ROOT: /home/travis/hunter/_testing/Hunter
-- [hunter *** DEBUG *** 2016-07-26T05:11:14] HUNTER_TOOLCHAIN_ID_PATH: /home/travis/hunter/_testing/Hunter/_Base/d76160c/6c9a4aa/8928885
-- [hunter *** DEBUG *** 2016-07-26T05:11:14] HUNTER_CONFIGURATION_TYPES: Release;Debug
-- [hunter] [ Hunter-ID: d76160c | Config-ID: 6c9a4aa | Toolchain-ID: 8928885 ]
CMake Error at /home/travis/hunter/_testing/Hunter/_Base/d76160c/6c9a4aa/config.cmake:1 (include):
  include could not find load file:

    hunter_config
Call Stack (most recent call first):
  /home/travis/hunter/_testing/Hunter/_Base/Download/Hunter/unknown/d76160c/Unpacked/cmake/modules/hunter_finalize.cmake:128 (include)
  /home/travis/hunter/_testing/Hunter/_Base/Download/Hunter/unknown/d76160c/Unpacked/cmake/modules/hunter_add_package.cmake:23 (hunter_finalize)
  CMakeLists.txt:196 (hunter_add_package)


CMake Error at /home/travis/hunter/_testing/Hunter/_Base/d76160c/6c9a4aa/config.cmake:2 (hunter_config):
  Unknown CMake command "hunter_config".
Call Stack (most recent call first):
[...]

Issue #293, opened 2016-01-02
Add GLFW3 Package


fire:
http://www.glfw.org/docs/latest/

ruslo:
GitHub official repository: https://github.com/glfw/glfw (note: CMake based)

ruslo:
Fixed https://github.com/ruslo/hunter/pull/487

PR #481, opened 2016-08-05
Added package glfw

pretyman:
Re-opened pull request https://github.com/ruslo/hunter/pull/480 against the develop branch


ruslo:
Testing here: https://github.com/ruslo/hunter/pull/480#issuecomment-237862975

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.765
* https://travis-ci.org/ingenue/hunter/builds/150082953

pretyman:
Ура!! Спасибо!!

ruslo:
> Ура!! Спасибо!!

You're welcome :)

Released: https://github.com/ruslo/hunter/releases/tag/v0.15.9
Feel free to add wiki page and close the https://github.com/ruslo/hunter/issues/293.

ruslo:
I think we have problems with full paths in cache:
* https://travis-ci.org/ingenue/hunter/jobs/150091811

```
make[2]: *** No rule to make target `/home/travis/build/ingenue/hunter/_testing/Hunter/_Base/9ed9e77/304b278/c07028d/Install/lib/libX11.a', needed by `glfw-example'.  Stop.
```
But ids: `-- [hunter] [ Hunter-ID: b4315bf | Config-ID: 304b278 | Toolchain-ID: c07028d ]`

pretyman:
The references to the full paths are in `<root-id>/lib/cmake/glfw3/glfw3Targets.cmake`:

```CMake
# Create imported target glfw
add_library(glfw STATIC IMPORTED)

set_target_properties(glfw PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "/usr/lib64/librt.so;/usr/lib64/libm.so;dl;/usr/lib64/libGL.so;/home/builder/.hunter/_Base/xxxxxxx/6c9a4aa/d17a7ab/Install/lib/libX11.a;-lpthread;/home/builder/.hunter/_Base/xxxxxxx/6c9a4aa/d17a7ab/Install/lib/libXrandr.a;/home/builder/.hunter/_Base/xxxxxxx/6c9a4aa/d17a7ab/Install/lib/libXinerama.a;/home/builder/.hunter/_Base/xxxxxxx/6c9a4aa/d17a7ab/Install/lib/libXxf86vm.a;/home/builder/.hunter/_Base/xxxxxxx/6c9a4aa/d17a7ab/Install/lib/libXcursor.a;/home/builder/.hunter/_Base/xxxxxxx/6c9a4aa/d17a7ab/Install/lib/libxcb.a;/home/builder/.hunter/_Base/xxxxxxx/6c9a4aa/d17a7ab/Install/lib/libXext.a;/home/builder/.hunter/_Base/xxxxxxx/6c9a4aa/d17a7ab/Install/lib/libXrender.a;/home/builder/.hunter/_Base/xxxxxxx/6c9a4aa/d17a7ab/Install/lib/libX11.a;/home/builder/.hunter/_Base/xxxxxxx/6c9a4aa/d17a7ab/Install/lib/libXfixes.a;/home/builder/.hunter/_Base/xxxxxxx/6c9a4aa/d17a7ab/Install/lib/libXau.a"
)
```

Should this file be added to `PACKAGE_UNRELOCATABLE_TEXT_FILES` or should `*.cmake` files be treated differently? I wrongly assumed `*.cmake` files were treated by the current implementation of `hunter_cacheable`

ruslo:
> Should this file be added to PACKAGE_UNRELOCATABLE_TEXT_FILES

I think no. CMake designed to do it nicely.

> it could be that the implementation of glfw's CMakeLists is not correct

Kind of. This is very similar to what is described here: https://github.com/ruslo/hunter/wiki/example.find_package.config#relocation-challenge

> Сэнсэй, что делать?

The final goal is not to have full path in `glfw*.cmake` installed files. It means we must not have commands like `target_link_libraries(glfw PUBLIC "<full-path-to-some-library>")`. I guess this is what happenning now by `target_link_libraries(glfw PUBLIC "${X11_LIBRARIES}")`. I need to think, usually introducing imported targets helps.

pretyman:
> This is very similar to what is described here: https://github.com/ruslo/hunter/wiki/example.find_package.config#relocation-challenge

Thanks

> It means we must not have commands like target_link_libraries(glfw PUBLIC "<full-path-to-some-library>")

Can you think of a way of making those paths relative? It is a hack it seems, but could (?) be valid. 

>  I need to think, usually introducing imported targets helps.

glfw is using CMake's `FindX11.cmake` which does not provide those targets, I guess we could either

1. Do like how you did with boost and have a `X11Config.cmake` in hunter which creates those targets. Downside is that there is quite a lot of targets to create in X11, but could start with the targets used by `glfw`
2. create those targets in a modified `FindX11.cmake` and try to push it upstream, but I believe it wouldn't help in practical terms because of the minimum required version for CMake being 3.0 for hunter. I just mention it because it breaks my heart that the CMake modules that come by default use old style CMake.

pretyman:
> Do like how you did with boost and have a X11Config.cmake in hunter which creates those targets. Downside is that there is quite a lot of targets to create in X11, but could start with the targets used by glfw

This one is best, I guess. Also having the same problem with the `glew` package, which needs to link to `x11`.

ruslo:
I like `X11Config.cmake` approach. Also since everything looks quite pretty for autotools projects we can keep high standards and make generation of `*Config.cmake` automatic. May be read info from `*.pc` files? Might be helpful: https://cmake.org/cmake/help/latest/module/FindPkgConfig.html

pretyman:
> Also since everything looks quite pretty for autotools projects we can keep high standards and make generation of *Config.cmake automatic. May be read info from *.pc files?

Ok, so let's say we want a function to create`xrandrConfig.cmake` for the package `xrandr` which depends on hunter/pkgconfig package `x11`.

```CMake


if(NOT TARGET xrandr)
    find_package(X11 CONFIG REQUIRED) 
    # should X11 be renamed to x11? 
    #   If so, we would have to patch the packages CMake scripts to find_package(x11 CONFIG ...)
    #   If not, we will have to somehow associate the uppercase X11 of the package with lower case x11 - any CMake tricks for that?
    add_library(xrandr IMPORTED ...)
    pkg_search_module(XRANDR xrandr REQUIRED)
    set_target_properties(xrandr PROPERTIES
        IMPORTED_LINK_INTERFACE_LANGUAGES_RELEASE "C"
        INTERFACE_LINK_LIBRARIES "${XRANDR_LIBRARIES}"
    )
    target_link_libraries(randr INTERFACE x11)   
endif()
```

I wonder how the arguments to the function will be for  difference between in cases between `X11` and `x11` makes this not so trivial :|

Any thoughts?

ruslo:
> should X11 be renamed to x11

Yes, I guess if we are planning to do generation automatically it make sense to rename package so it fits `*.pc` layout: `lib/pkgconfig/x11.pc`.

> pkg_search_module(XRANDR xrandr REQUIRED)

I guess this will not work as is, at least correctly. There must be our custom root variables somewhere, no?

pretyman:
> I guess this will not work as is, at least correctly. There must be our custom root variables somewhere, no?

Well, spotted, there seems to be  [PKG_CONFIG_USE_CMAKE_PREFIX_PATH](https://cmake.org/cmake/help/v3.1/module/FindPkgConfig.html#variable:PKG_CONFIG_USE_CMAKE_PREFIX_PATH), but I can only find it on cmake's 3.1 onwards documentation.

Maybe in this particular case we can have `glfw3Targets.cmake` + `glew`'s files  as exceptions in `PACKAGE_UNRELOCATABLE_TEXT_FILES` to not have a broken package? I'm sure the CMake best practice police won't be knocking at your door for it.

I would rather first spend time in https://github.com/ruslo/hunter/issues/218 [ + https://github.com/ruslo/hunter/pull/469 ] as I believe it is blocking your project to have some visibility: [[odb-users] Need Help for compiling ODB for iOS](
http://www.codesynthesis.com/pipermail/odb-users/2016-August/003384.html). I remember that with hunter, odb was working in iOS when `*.la` and `*.pc` files were removed from packages.

But I have to finish pushing these 3D graphics related packages before https://github.com/ruslo/hunter/issues/218 , so I give the students a hunter based build of Computer Graphics course practical assignments. This should also give some visibility plus will help students, as they have Win/Linux/OSX builds. Right now they are only provided Visual Studio projects with the libraries included [ some are not able to setup an Xcode project with the libraries from brew by themselves, so hunter would do the library dependency management ].

ruslo:
> Maybe in this particular case we can have glfw3Targets.cmake + glew's files as exceptions in PACKAGE_UNRELOCATABLE_TEXT_FILES to not have a broken package?

You can try it. I don't know how glfw `*.cmake` modules generated but usually there are build specific configs like `*-Debug.cmake` or `*-Release.cmake` so it may be tricky.

Also if we are postponing this enhancement let's open separate issue with description and applied workarounds.

> I would rather first spend time in #218 [ + #469 ]

Yep, I remember we do hit this at least 3 times :) and @lucmichalski reported this one once. It would be awesome to have it fixed.

PR #484, opened 2016-08-10
Add Protocolor Buffers 3

Osse:


ruslo:
Testing:
* https://travis-ci.org/ingenue/hunter/builds/151415877

ruslo:
> Testing

Failed. @Osse please investigate.

Osse:
To make use of the convenience macros, such as `protobuf_generate_cpp` which is the cause of the errors here, you have to [`set(protobuf_MODULE_COMPATIBLE TRUE)`](https://github.com/hunter-packages/protobuf/blob/hunter/cmake/protobuf-config.cmake.in#L10-L13).  It works if you set it before `find_package(Protobuf)`.

I don't know if it can be made part of the Hunter package configuration somehow. I suspect it can't since the file will be read when configuring the main project. Alternatively we can remove the `if()` as part of the hunterization, but that breaks more with upstream. I [asked](https://github.com/google/protobuf/issues/1931) about what the intention of this guard is, but I didn't get a clear answer. It can also be documented in the Wiki, but I decided to not update it until the PR was accepted, but I can update it immediately if you wish.

ruslo:
I don't know details but [Protobuf example](https://github.com/ruslo/hunter/blob/master/examples/Protobuf/CMakeLists.txt) has worked before: https://travis-ci.org/ingenue/hunter/builds/121754452 and now it stops. You will break old code with this update, that's my point :) At first glance this looks like a bug somewhere, either in upstream or in hunterization.

> To make use of the convenience macros, such as protobuf_generate_cpp which is the cause of the errors here,

Example is not using such module

Osse:
I agree this will break old code if nothing is changed, which is bad. However, I thought it would break old code anyway, because of the change of the namespace name (`Protobuf::`to `protobuf::`). 

> Example is not using such module

[Yes it does](https://github.com/ruslo/hunter/blob/master/examples/Protobuf/messages/CMakeLists.txt#L4). This function is not available by default unless the variable I mentioned above is set. This is a change in the upstream CMake files between beta2 and the relase of v3.0.0.

ruslo:
> Yes it does.

Indeed, sorry missed that.

> This is a change in the upstream CMake files between beta2 and the relase of v3.0.0

Okay I see. What is the way to use this code without `set(protobuf_MODULE_COMPATIBLE TRUE)`? You can add one example `Protobuf-compat` with this variable and new example for `Protobuf`. We can test both.

ruslo:
Implemented here: https://github.com/ruslo/hunter/pull/496

Issue #494, opened 2016-08-19
Cross compiling with Hunter


fwinnen:
I have a cmake crossplatform c++ project which uses Protobuf and has to support at least Linux, iOS and Android. Cross compiling protobuf is not the issue, but the project setup gets kind of messy when cross compiling libprotobuf for target (i.e iOS) and having protoc on host (i.e macOS) available to generate source files from .proto in cmake.

I had a custom setup with a hierarchy of ExternalProject's where the outer-most project uses default toolchain and creates protoc for host and the actual project for target with custom toolchains for iOS
and Android. 

As you see, this is hard to maintain, so i am investigating polly's toolchains with hunter for dependencies (Protobuf).

I was able to adapt the hunter package of Protobuf to skip compiling protoc when cross compiling and import a user specified host-protoc so everything works fine. Only problem is, that i need to supply the instance of protoc which i would like to also have as a hunter dependency.

Do you guys have some idea, how i can manage hunter dependencies for different toolchains?
If there is some mechanic in hunter that i could extend for that purpose or if there is some other way to achieve this with hunter, please let me know.
##

ruslo:
I haven't worked with protobuf but as far as I understand you need to have protobuf library and some protoc executable tool, right? I.e. you need some kind of mix:
* Linux host + Android target: protoc tool compiled with Linux host toolchain + protobuf lib compiled with Android toolchain
* OSX host + Android target: protoc tool compiled with OSX host toolchain + protobuf lib compiled with Android toolchain
* OSX host + iOS target: protoc tool compiled with OSX host toolchain + protobuf lib compiled with iOS toolchain
* OSX host + OSX target: protoc tool and protobuf lib compiled both with OSX host toolchain

Is my vision correct?

fwinnen:
Thx for the fast response!

Your vision is totally right.
Also the hunter package already builds the library and the executable (obviously for the same target). I would imagine, that i need to switch toolchains in code, or supply a separate host toolchain for hunter_add_package...

ruslo:
> Your vision is totally right

It is good :) Another good news is that it's important feature and I have a plan to implement it.
However the bad news is that it's quite tricky and I can't say in will be done in near future. This is something that is not designed in CMake itself: https://cgold.readthedocs.io/en/latest/overview/cmake-can-not.html#incomplete-functionality-coverage

> supply a separate host toolchain for hunter_add_package

[...]

Issue #495, opened 2016-08-19
Controlling multiple toolchains with Hunter


ruslo:
For now Hunter do manage only one toolchain (target). In some scenarios we need to manage host toolchain too (even several number of toolchains):

* Managing version of compiler (GCC, [Clang](https://github.com/ruslo/hunter/wiki/pkg.llvm.clang))
* Managing version of used tools for scripting (Doxygen, [Python](https://github.com/ruslo/hunter/issues/388))
* Managing set of toolchain ([Polly](https://github.com/ruslo/hunter/issues/150))
* Managing CMake itself (?) (bootstrap with one CMake and use it to build another version, then start all 3rd party builds with new version)
* Custom user's tool ([Protobuf](https://github.com/ruslo/hunter/issues/494))

Some problems I see so far:
* Dependency between them, e.g. toolchain file depends on the compiler. There should be some mechanism so toolchain can say: "I need Clang", then we should look at user's settings (config.cmake) for version of Clang to download and launch build/install. Then we should process toolchain again so it can save all compiler related values and finally go to the user's code.
* Distinguishing host and target toolchains in user's code. E.g. there should be different commands for finding tools for target and tools for host. Something like OpenCV/Android toolchain do: https://github.com/ruslo/polly/blob/1615c8126a5d34941612e91530a2cd2e6e4f5872/os/android.toolchain.cmake#L1541-L1559
* We might need several host toolchain since some builds are not portable. E.g. we can't build Git with Visual Studio, hence we need to have MinGW-based host toolchain and Visual Studio-based host toolchain. This will be tricky at usage level since if we want to have both roots in PATH it will be similar to having one root with mix of binaries, lead to all kind of problems.

PR #496, opened 2016-08-20
Protobuf cross compilation example

fwinnen:
Proposed solution for #494 

Quick Overview:
* use updated Protobuf package (pull request pending)
* added cross compilation support to the Protobuf example
* added a macro to build on host (this was only tested osx,iOS and osx,Android)

I reworked the Protobuf example to work for iOS and Android and added a helper macro which i thought would be helpful for other situations.

Tell me what you think!


cmake/modules/hunter_add_host_subdirectory.cmake
```diff
@@ -0,0 +1,69 @@
...
+#   after the host project build finished
+
+include(CMakeParseArguments) # cmake_parse_arguments
+
+include(hunter_internal_error)
+include(hunter_status_print)
+include(hunter_status_debug)
+
+# internal variables: _hunter_ahs_*
+macro(hunter_add_host_subdirectory)
```
ruslo: Do we need to use `macro` here? Can you rewrite it in terms of `function`? In this case we can drop all this `_hunter_ahs_*` variables.

cmake/modules/hunter_add_host_subdirectory.cmake
```diff
@@ -0,0 +1,69 @@
...
+
+# Usage:
+#   hunter_add_host_subdirectory(host)
+#   hunter_add_host_subdirectory(host
+#       BINARY_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/host-build
+#       ENV_VARS CC CXX
+#   )
+
+# Note:
+#   With polly Android toolchain this fails if you do not unset CC and CXX ENV{} variables.
```
ruslo: What kind of error do you have? To clean environment you can use https://github.com/ruslo/hunter/blob/master/scripts/clear-all.cmake module. Also I personally think that `CC`/`CXX` is completely "autotoolish" stuff and users should use CMAKE_CXX_COMPILER instead. Note that if you want to change compiler for all packages using toolchain is a requirement for Hunter.

cmake/modules/hunter_add_host_subdirectory.cmake
```diff
@@ -0,0 +1,69 @@
...
+  # TODO: Check why this is needed only for Android
+  # save and unset specified ENV variables
+  foreach(var ${_hunter_ahs_arg_ENV_VARS})
+      hunter_status_debug("save ${var}: $ENV{${var}}")
+      set(_hunter_ahs_saved_${var} $ENV{${var}})
+      unset(ENV{${var}})
+  endforeach()
+
+  # invoke cmake for host project
+  execute_process(COMMAND "${CMAKE_COMMAND}" -E make_directory "${_hunter_ahs_arg_BINARY_DIRECTORY}")
```
ruslo: No need to create directory, it will be created automatically in `-B`.

cmake/modules/hunter_add_host_subdirectory.cmake
```diff
@@ -0,0 +1,69 @@
...
+  execute_process(COMMAND "${CMAKE_COMMAND}" "-H${_hunter_ahs_HOME_DIRECTORY}" "-B${_hunter_ahs_arg_BINARY_DIRECTORY}")
```
ruslo: Please check exit code of this command.

cmake/modules/hunter_add_host_subdirectory.cmake
```diff
@@ -0,0 +1,69 @@
...
+  execute_process(COMMAND "${CMAKE_COMMAND}" --build "${_hunter_ahs_arg_BINARY_DIRECTORY}")
```
ruslo: All dependencies will be installed on generate step, no need to run build. See https://docs.hunter.sh/en/latest/user-guides/regular.html#hunter-in-cmake-environment

examples/Protobuf/CMakeLists.txt
```diff
@@ -15,8 +35,10 @@ find_package(Protobuf CONFIG REQUIRED)
 
 add_subdirectory(messages)
 
-# path to generated headers
-include_directories(${CMAKE_CURRENT_BINARY_DIR})
 add_executable(hello main.cpp)
-target_link_libraries(hello messages Protobuf::libprotobuf)
-                                     
+target_link_libraries(hello messages)
+install(TARGETS hello
```
ruslo: No installation will be triggered for tests, just build.

examples/Protobuf/CMakeLists.txt
```diff
@@ -15,8 +35,10 @@ find_package(Protobuf CONFIG REQUIRED)
...
+        RUNTIME DESTINATION bin
+        BUNDLE DESTINATION bin)
+
+# TODO add tests
```
ruslo: No tests will be triggered too.

examples/Protobuf/cmake/Hunter/config.cmake
```diff
@@ -0,0 +1,14 @@
...
+                protobuf_BUILD_PROTOC=OFF
+                protobuf_BUILD_LIBPROTOBUF_LITE=OFF
+    )
+else()
+    hunter_config(Protobuf
+            VERSION ${HUNTER_Protobuf_VERSION}
+            CMAKE_ARGS
+                protobuf_BUILD_LIBPROTOBUF_LITE=OFF
+    )
+endif()
```
ruslo: http://0.readthedocs.io/en/latest/git.html#check-there-is-a-new-line-at-end-of-file

cmake/modules/hunter_add_host_subdirectory.cmake
fwinnen: -- [hunter] >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
Including HunterGate: /***/hunter/gate/cmake/HunterGate.cmake
-- The C compiler identification is GNU 4.9.0
-- The CXX compiler identification is GNU 4.9.0
-- Checking whether C compiler has -isysroot
-- Checking whether C compiler has -isysroot - yes
-- Checking whether C compiler supports OSX deployment target flag
-- Checking whether C compiler supports OSX deployment target flag - no
-- Check for working C compiler: /***/ndk/toolchains/aarch64-linux-android-4.9/prebuilt/darwin-x86_64/bin/aarch64-linux-android-gcc
-- Check for working C compiler: /***/ndk/toolchains/aarch64-linux-android-4.9/prebuilt/darwin-x86_64/bin/aarch64-linux-android-gcc -- broken
CMake Error at /path/to/cmake/3.5.2/share/cmake/Modules/CMakeTestCCompiler.cmake:61 (message):
  The C compiler
  "/***/ndk/toolchains/aarch64-linux-android-4.9/prebuilt/darwin-x86_64/bin/aarch64-linux-android-gcc"
  is not able to compile a simple test program.

examples/Protobuf/messages/CMakeLists.txt
```diff
@@ -1,7 +1,50 @@
...
+    target_link_libraries(messages protobuf::libprotobuf)
+    target_include_directories(messages PUBLIC ${CMAKE_CURRENT_BINARY_DIR}/..)
+endif()
+
+#install(FILES "${MESSAGES_HDRS}"
+#        DESTINATION "include/messages"
+#        COMPONENT messages-headers)
+#install(TARGETS messages
+#        LIBRARY DESTINATION lib COMPONENT libmessages
+#        ARCHIVE DESTINATION lib COMPONENT libmessages)
```
ruslo: Please clean this stuff

examples/common.cmake
```diff
@@ -30,5 +30,12 @@ if(NOT url_is_empty AND NOT sha1_is_empty AND NOT hunter_root_is_empty)
...
 else()
   get_filename_component(HUNTER_ROOT "${CMAKE_CURRENT_LIST_DIR}/.." ABSOLUTE)
-  HunterGate(URL "x" SHA1 "xxxxxxxx" ${TESTING_CONFIG_OPT})
+
+  # allow examples to have custom config.cmake files
+  string(COMPARE EQUAL "${HUNTER_TESTING_CONFIG_FILE}" "" has_no_custom_config)
+  if (has_no_custom_config)
+    HunterGate(URL "x" SHA1 "xxxxxxxx" ${TESTING_CONFIG_OPT})
+  else(has_no_custom_config)
+    HunterGate(URL "x" SHA1 "xxxxxxxx" ${TESTING_CONFIG_OPT} FILEPATH ${HUNTER_TESTING_CONFIG_FILE})
```
ruslo: You can use `set(TESTING_CONFIG_OPT FILEPATH ${HUNTER_TESTING_CONFIG_FILE})` with the previous version to do the same.

cmake/modules/hunter_add_host_subdirectory.cmake
fwinnen: I thought this might be helpful for other use cases, since the 'host-project' is a regular CMake project and can be used for other things needed on host (i.e build docs...)

ruslo:
> Tell me what you think!

Well I have a workaround proposal, I haven't thinking about adding a new module (https://github.com/ruslo/hunter/issues/495 should be fixed nicely instead). Though I'm okay with new module but it should be generalized and follow some conventions:

* rename this module to `hunter_experimental_*` so it will be clear that you're using something that not designed well and may be removed anytime
* put a huge warning message "you are using experimental stuff bla-bla..."
* use `${CMAKE_BINARY_DIR}/_3rdParty/Hunter` directory for work, e.g. `${CMAKE_BINARY_DIR}/_3rdParty/Hunter/host-build-binary-tree`
* Save file with host root directory to `${CMAKE_BINARY_DIR}/_3rdParty/Hunter/install-root-dir`.

cmake/modules/hunter_add_host_subdirectory.cmake
ruslo: Are we trying to build host package here? Well that's the problem with global environment vars, looks like CC Android compiler was used for both target and host. User should use toolchains instead.

cmake/modules/hunter_add_host_subdirectory.cmake
ruslo: I think it should be very trivial file. For complex stuff things became complex, e.g. what about Debug/Release builds?

fwinnen: Yes I can agree on that. We have to consider a new name for the module then I guess, because  my intention was to have this as a 'add_subdiretcory' replacement with the addition to be run on host.

cmake/modules/hunter_add_host_subdirectory.cmake
fwinnen: Should have included the start:

`build.py --clear --toolchain android-ndk-r11c-api-21-arm64-v8a
Python version: 3.5
Build dir: /.../hunter/examples/Protobuf/_builds/android-ndk-r11c-api-21-arm64-v8a
Remove directory: /.../hunter/examples/Protobuf/_builds/android-ndk-r11c-api-21-arm64-v8a
Execute command: [
  `which`
  `cmake`
]

[/.../hunter/examples/Protobuf]> "which" "cmake"

/usr/local/bin/cmake
Execute command: [
  `cmake`
  `--version`
]

[/.../hunter/examples/Protobuf]> "cmake" "--version"

cmake version 3.5.2

CMake suite maintained and supported by Kitware (kitware.com/cmake).
Execute command: [
  `cmake`
  `-H.`
  `-B/.../hunter/examples/Protobuf/_builds/android-ndk-r11c-api-21-arm64-v8a`
  `-GUnix Makefiles`
  `-DCMAKE_TOOLCHAIN_FILE=/.../polly/android-ndk-r11c-api-21-arm64-v8a.cmake`
]

[/.../hunter/examples/Protobuf]> "cmake" "-H." "-B/.../hunter/examples/Protobuf/_builds/android-ndk-r11c-api-21-arm64-v8a" "-GUnix Makefiles" "-DCMAKE_TOOLCHAIN_FILE=/Users/Fabian/Development/kinemic/polly/android-ndk-r11c-api-21-arm64-v8a.cmake"

Including HunterGate: /.../hunter/gate/cmake/HunterGate.cmake
-- [polly] Used toolchain: Android NDK r11c / API 21 / arm64-v8a / c++11 support
-- The C compiler identification is GNU 4.9.0
-- The CXX compiler identification is GNU 4.9.0
-- Check for working C compiler: /.../ndk/toolchains/aarch64-linux-android-4.9/prebuilt/darwin-x86_64/bin/aarch64-linux-android-gcc
-- Check for working C compiler: /.../ndk/toolchains/aarch64-linux-android-4.9/prebuilt/darwin-x86_64/bin/aarch64-linux-android-gcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /.../ndk/toolchains/aarch64-linux-android-4.9/prebuilt/darwin-x86_64/bin/aarch64-linux-android-g++
-- Check for working CXX compiler: /.../ndk/toolchains/aarch64-linux-android-4.9/prebuilt/darwin-x86_64/bin/aarch64-linux-android-g++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- [hunter] invoke cmake subdirectory 'host' build for host...`

I used only polly android toolchain, also i did not find anything in it which could set CC and CXX. Just figured out when inspecting the issue, that those were causing the issue.

fwinnen:
> put a huge warning message "you are using experimental stuff bla-bla..."

Is there some hunter_print for warnings that i missed? Or is a `message(WARNING ...)` Ok for this?

cmake/modules/hunter_add_host_subdirectory.cmake
ruslo: Can you please print values of `CC`/`CXX` environment variable before `project` command (before loading toolchain), after `project` command and before running `execute_process` with host build command.

ruslo:
> Or is a message(WARNING ...) Ok for this?

I hope such tricks will not became something regular so `message(WARNING ...)` is okay.

cmake/modules/hunter_add_host_subdirectory.cmake
fwinnen: -- _______CC before_project:
-- _______CXX before_project:
-- [polly] Used toolchain: Android NDK r11c / API 21 / arm64-v8a / c++11 support
...
-- _______CC after_project: /.../ndk/toolchains/aarch64-linux-android-4.9/prebuilt/darwin-x86_64/bin/aarch64-linux-android-gcc
-- _______CXX after_project: /.../ndk/toolchains/aarch64-linux-android-4.9/prebuilt/darwin-x86_64/bin/aarch64-linux-android-g++
-- [hunter] invoke cmake subdirectory 'host' build for host...


-- [hunter] >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
-- _______CC before execute_process: /.../ndk/toolchains/aarch64-linux-android-4.9/prebuilt/darwin-x86_64/bin/aarch64-linux-android-gcc
-- _______CXX before execute_process: /.../ndk/toolchains/aarch64-linux-android-4.9/prebuilt/darwin-x86_64/bin/aarch64-linux-android-g++
Including HunterGate: /.../hunter/gate/cmake/HunterGate.cmake
-- _______CC before_host_project: /.../ndk/toolchains/aarch64-linux-android-4.9/prebuilt/darwin-x86_64/bin/aarch64-linux-android-gcc
-- _______CXX before_host_project: /.../ndk/toolchains/aarch64-linux-android-4.9/prebuilt/darwin-x86_64/bin/aarch64-linux-android-g++
...

fwinnen:
Thank you for the feedback! I will change the suggestions and ping you once i pushed it.

cmake/modules/hunter_add_host_subdirectory.cmake
ruslo: I just tried simple GCC host toolchain and it looks like those variables set by CMake itself :)

cmake/modules/hunter_experimental_add_host_subdirectory.cmake
```diff
@@ -0,0 +1,73 @@
+# Copyright (c) 2016 Fabian Winnen
+# All rights reserved.
+
+# Adds a subdirectory with CMakeList.txt as seperate build with default (host) toolchain
+
+# Usage:
+#   hunter_add_host_subdirectory(host)
+#   hunter_add_host_subdirectory(host
+#       BINARY_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/host-build
+#       ENV_VARS CC CXX
```
ruslo: I don't think that we should provide interface for user so he can set list of environment variables. I guess using list internally will work fine. Also I'm not sure about using [clear-all.cmake](https://github.com/ruslo/hunter/blob/master/scripts/clear-all.cmake) module since usually this module used in a separate cmake process so it doesn't affect user's environment. You can try it and if there will be no problems do include it.

Same with binary directory, see the notes below. By the way it will be convenient if we do return `HUNTER_HOST_ROOT` variable to the parent scope so user don't have to read our `_3rdParty/Hunter/...` files.

fwinnen: > I don't think that we should provide interface for user so he can set list of environment variables. 

This was meant as a experimental feature to fix problems for users which I did not encounter, since I do not fully understand whats causing the problem

> Also I'm not sure about using clear-all.cmake module since usually this module used in a separate cmake process so it doesn't affect user's environment. You can try it and if there will be no problems do include it.

Do you mean including clear-all.cmake on client side in host/CMakeList.txt before project (I think that is too late for cmake not to pick that false CC up) or in hunter_experimental_add_host_subdirectory before `execute_process`. The Problem I see with the later is, that it may clear some variables purposefully set by some toolchain and not restore them afterwards.

> Same with binary directory, see the notes below.

I will set a fixed binary_directory for the host project as you recommended. But I don't understand what you mean with: 

> By the way it will be convenient if we do return HUNTER_HOST_ROOT variable to the parent scope so user don't have to read our _3rdParty/Hunter/... files.

ruslo: > I do not fully understand whats causing the problem

I hope it is clear now (?) Toolchain set CMAKE_CXX_COMPILER and CMake do set `CC` environment variable using value of CMAKE_CXX_COMPILER. When we do run `execute_process` environment variable `CC` inherited and used as a compiler (CMake read `CC` environment variable if it's set).

> Do you mean including clear-all.cmake on client side in host/CMakeList.txt before project

Well in this case there must be no problems with using `clear-all.cmake`.

> in hunter_experimental_add_host_subdirectory before execute_process

Yes I mean this approach. Can you please try it and see what happens?

> The Problem I see with the later is, that it may clear some variables purposefully set by some toolchain and not restore them afterwards

We should not use those variables in build. Environment variables customization doesn't scale well. May be it shows us some hidden problems.

> But I don't understand what you mean with

I mean something like this:

```cmake
file(READ "${_args_BINARY_DIRECTORY}/_3rdParty/Hunter/install-root-dir" HUNTER_HOST_ROOT)
set(HUNTER_HOST_ROOT "${HUNTER_HOST_ROOT}" PARENT_SCOPE)
```

so we can use `HUNTER_HOST_ROOT` in target's CMakeLists.txt without reading Hunter internal stuff:

```
hunter_add_host_project(host) # set HUNTER_HOST_ROOT variable
add_executable(protobuf::protoc IMPORTED)
set_property(TARGET protobuf::protoc APPEND PROPERTY IMPORTED_CONFIGURATIONS RELEASE)
set_target_properties(protobuf::protoc PROPERTIES IMPORTED_LOCATION_RELEASE "${HUNTER_HOST_ROOT}/bin/protoc")
```

fwinnen:
Ok, i like the approach with HUNTER_HOST_ROOT and the binary directory. I implemented all suggestions i guess.

Some thoughts:

* what's the safest way to access ../../scripts/clear-all.cmake (see TODO in code)
* We now rely on the client to call 
`file(WRITE "${CMAKE_BINARY_DIR}/_3rdParty/Hunter/install-root-dir" "${PROTOBUF_ROOT}")` in the host project CMakeList.txt. I don't know if I am happy with that, I implemented instructions in the error message though.
* If we install more then one dependency in the host project, is there any chance that `PROTOBUF_ROOT` and `GTEST_ROOT` will be different? Then  we had a problem with `HUNTER_HOST_ROOT`...

cmake/modules/hunter_experimental_add_host_project.cmake
```diff
@@ -0,0 +1,66 @@
...
+  endif()
+  list(GET _arg_UNPARSED_ARGUMENTS 0 _home_directory)
+
+  set(_binary_directory "${CMAKE_BINARY_DIR}/_3rdParty/Hunter/host-build-binary-tree")
+
+  hunter_status_print("invoke cmake subdirectory '${_home_directory}' build for host...\n\n")
+  hunter_status_print(">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>")
+
+  # invoke cmake for host project
+  # TODO: I guess this is not portable ?! how do access HUNTER_SELF?
```
ruslo: I think `HUNTER_SELF` should be available here. So it will be `${HUNTER_SELF}/scripts/clear-all.cmake`. Check file exists just to be sure.

cmake/modules/hunter_experimental_add_host_project.cmake
```diff
@@ -0,0 +1,66 @@
...
+include(hunter_status_print)
+include(hunter_status_debug)
+
+function(hunter_experimental_add_host_project)
+  cmake_parse_arguments(_arg "" "" "" ${ARGV})
+
+  # Get subdirectory home
+  list(LENGTH _arg_UNPARSED_ARGUMENTS _len)
+  if(NOT ${_len} EQUAL 1)
+      hunter_internal_error(
```
ruslo: It's not internal error, it's user error.

cmake/modules/hunter_experimental_add_host_project.cmake
```diff
@@ -0,0 +1,66 @@
...
+
+include(hunter_internal_error)
+include(hunter_status_print)
+include(hunter_status_debug)
+
+function(hunter_experimental_add_host_project)
+  cmake_parse_arguments(_arg "" "" "" ${ARGV})
+
+  # Get subdirectory home
+  list(LENGTH _arg_UNPARSED_ARGUMENTS _len)
```
ruslo: We are in function now so can use pretty names for variables. Let it be `arg` and `len`.

cmake/modules/hunter_experimental_add_host_project.cmake
```diff
@@ -0,0 +1,66 @@
...
+#   TODO: more disclamers
+#   TODO: notice that we clear some env variables?
+#
+# Adds a project with CMakeList.txt as seperate build with default (host) toolchain
+#
+# Usage:
+#   hunter_experimental_add_host_project(host)
+#
+# sets HUNTER_HOST_ROOT variable to point at host project's binary directory
+message(WARNING "You are using a hunter module in experimental stage:\nhunter_experimental_add_host_project\n")
```
ruslo: Let's put it into function, not when user do include this module.

Also I was thinking about more scary message like:

```
*****************************************************************
                                 !!! WARNING !!!

You're using experimental feature. This may not work well
and can be removed anytime. It's a workaround for existsing
feature request:

    https://github.com/ruslo/hunter/issues/495

*****************************************************************
```

cmake/modules/hunter_experimental_add_host_project.cmake
```diff
@@ -0,0 +1,66 @@
...
+      hunter_internal_error(
+              "'hunter_experimental_add_host_project' incorrect usage,"
+              " expected one project argument"
+      )
+  endif()
+  list(GET _arg_UNPARSED_ARGUMENTS 0 _home_directory)
+
+  set(_binary_directory "${CMAKE_BINARY_DIR}/_3rdParty/Hunter/host-build-binary-tree")
+
+  hunter_status_print("invoke cmake subdirectory '${_home_directory}' build for host...\n\n")
```
ruslo: Let's make it `Building host directory: ${_home_directory}`

cmake/modules/hunter_experimental_add_host_project.cmake
```diff
@@ -0,0 +1,66 @@
...
+  hunter_status_print(">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>")
```
ruslo: Not look pretty :) let's remove it

cmake/modules/hunter_experimental_add_host_project.cmake
```diff
@@ -0,0 +1,66 @@
...
+      hunter_internal_error(
+              "Could not read file 'install-root-dir'."
+              "Create this file at the end of your host project's CMakeList.txt with:"
+              "  'file(WRITE \"\${CMAKE_BINARY_DIR}/_3rdParty/Hunter/install-root-dir\" \"\${<Package>_ROOT}\")'"
+      )
+  endif()
+  file(READ "${_binary_directory}/_3rdParty/Hunter/install-root-dir" HUNTER_HOST_ROOT)
+  set(HUNTER_HOST_ROOT "${HUNTER_HOST_ROOT}" PARENT_SCOPE)
+
+  hunter_status_print("<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<\n\n")
```
ruslo: same

cmake/modules/hunter_experimental_add_host_project.cmake
```diff
@@ -0,0 +1,66 @@
...
+  # invoke cmake for host project
+  # TODO: I guess this is not portable ?! how do access HUNTER_SELF?
+  include(../../scripts/clear-all.cmake)
+  execute_process(
+          COMMAND "${CMAKE_COMMAND}" "-H${_home_directory}" "-B${_binary_directory}"
+          RESULT_VARIABLE _exit_code
+  )
+  string(COMPARE EQUAL "${_exit_code}" "0" _success)
+  if(NOT ${_success})
+      hunter_internal_error("Failed to generate host project, exit status ${_exit_code}.")
```
ruslo: Not a Hunter issue too. It's a user error.

cmake/modules/hunter_experimental_add_host_project.cmake
```diff
@@ -0,0 +1,66 @@
...
+  endif()
+
+  # read host projects install root
+  if(NOT EXISTS "${_binary_directory}/_3rdParty/Hunter/install-root-dir")
```
ruslo: That's the magic of Hunter: user don't have to create such file, we can do it ourselfs in `hunter_finalize`: https://github.com/ruslo/hunter/blob/a74fc84308fb16a3c93dc4207c1df52cfe1b1cd6/cmake/modules/hunter_finalize.cmake path to root is `HUNTER_INSTALL_PREFIX`.

examples/Protobuf/host/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+set(TESTING_CONFIG_OPT FILEPATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/Hunter/config.cmake)
+include("../../common.cmake")
+
+project(download-Protobuf-host)
+
+# only build protoc
+hunter_add_package(Protobuf)
+
+# write the install-root-dir into file for target project
+file(WRITE "${CMAKE_BINARY_DIR}/_3rdParty/Hunter/install-root-dir" "${PROTOBUF_ROOT}")
```
ruslo: Can be removed, see notes above

examples/Protobuf/messages/CMakeLists.txt
```diff
@@ -1,7 +1,41 @@
...
-# required since no linking
-include_directories(${PROTOBUF_INCLUDE_DIR})
-PROTOBUF_GENERATE_CPP(MESSAGES_SRC MESSAGES_HEADER messages.proto)
-add_library(messages ${MESSAGES_HEADER} ${MESSAGES_SRC})
+set(MESSAGES_PROTOS messages.proto)
+set(MESSAGES_SRCS)
+set(MESSAGES_HDRS)
 
+#Code Generation
+if(protobuf_MODULE_COMPATIBLE) #Legacy Support
```
ruslo: Let's create two examples. It will be easier to understand. Also users following new trends may not be interested in this legacy stuff at all. We already did this for OpenSSL: https://github.com/ruslo/hunter/blob/master/examples/OpenSSL/CMakeLists.txt https://github.com/ruslo/hunter/blob/master/examples/OpenSSL-imported/CMakeLists.txt

cmake/modules/hunter_experimental_add_host_project.cmake
fwinnen: `HUNTER_SELF` is not set.
I checked your code and all i found is `_hunter_self` which is ok to use inside the hunter project i guess

ruslo: Pull new commits from branch `develop`. Should be available.

ruslo:
> If we install more then one dependency in the host project, is there any chance that PROTOBUF_ROOT and GTEST_ROOT will be different? Then we had a problem with HUNTER_HOST_ROOT...

No, HUNTER_HOST_ROOT is the same for all packages. It's value equal to `HUNTER_ROOT` + id-triple (hunter-id, config-id, toolchain-id) + `Install`.

cmake/modules/hunter_experimental_add_host_project.cmake
fwinnen: Yeah, that was just my first try for some visual indication that between >> and << its the host project. Any replacement for visual clues or just drop it?

ruslo: I think we can drop that. We already have warning that will help find the place where this function called :) I guess `Building host project: ...` + `Host project build finished: ...` is enough.

cmake/modules/hunter_experimental_add_host_project.cmake
fwinnen: Hmm, can you explain that a little bit more? When does hunter_finalize gets executed and how can we inject the file creation in it?

examples/Protobuf/messages/CMakeLists.txt
fwinnen: Is there anything behind the -imported postfix? 

So should i name the ${PROTOBUF_INCLUDE_DIR} Version Protobuf
and the protobuf::libprotobuf Version Protobuf-imported?

cmake/modules/hunter_experimental_add_host_project.cmake
ruslo: > When does hunter_finalize gets executed

After project (this is where toolchain loaded) on first `hunter_add_package` command. At this place we know toolchain-id and config-id. I've pushed update to branch `develop`: https://github.com/ruslo/hunter/commit/b5c3654f09906ca39c222d24eadbd707deea1ed4

examples/Protobuf/messages/CMakeLists.txt
ruslo: You can name example as you wish. It may be `protobuf-legacy` or `protobuf-compatible` or anything. The point is to have two examples instead of one hairy `if(..._COMPATIBLE)` version.

cmake/modules/hunter_experimental_add_host_project.cmake
fwinnen: Perfect! Will report back when i get this working. 
The optional build got me some problems with the module mode, which i noticed when
i separated the examples :D

fwinnen:
Ok

* examples split
* pretty output and warning
* very simple host/CMakeList.txt

cmake/modules/hunter_experimental_add_host_project.cmake
```diff
@@ -0,0 +1,78 @@
...
+      RESULT_VARIABLE exit_code
+  )
+  string(COMPARE EQUAL "${exit_code}" "0" success)
+  if(NOT ${success})
+    hunter_user_error("Failed to generate host project: ${home_directory}, exit status ${exit_code}.")
+  endif()
+
+  # read host projects install root
+  if(NOT EXISTS "${binary_directory}/_3rdParty/Hunter/install-root-dir")
+    hunter_user_error(
```
ruslo: Well now it's a fatal error since we are responsible for it, not user.

cmake/projects/Protobuf/hunter.cmake
```diff
@@ -7,29 +7,51 @@ include(hunter_download)
...
+    "3.0.0-beta-2-p1"
+    URL
+    "https://github.com/hunter-packages/protobuf/archive/v3.0.0-beta-2-hunter-2.tar.gz"
+    SHA1
+    dbae4995254f351918d0462b308b561dbab71b3f
+)
 
 
-hunter_cmake_args(Protobuf CMAKE_ARGS  CMAKE_ARGS protobuf_BUILD_TESTS=OFF)
+hunter_cmake_args(Protobuf CMAKE_ARGS CMAKE_ARGS protobuf_BUILD_TESTS=OFF)
```
ruslo: `hunter_cmake_args(Protobuf CMAKE_ARGS CMAKE_ARGS protobuf_BUILD_TESTS=OFF)` should be just `hunter_cmake_args(Protobuf CMAKE_ARGS protobuf_BUILD_TESTS=OFF)`, I guess I missed that on previous pull request.

cmake/modules/hunter_experimental_add_host_project.cmake
fwinnen: Hehe, i thought to add that message since the error will be encountered if the host project does not use hunter. But yeah, if we note that it only works with projects using hunter this would be fatal error

examples/Protobuf-legacy/cmake/Hunter/config.cmake
```diff
@@ -0,0 +1,14 @@
+if(ANDROID OR IOS)
+  hunter_config(Protobuf
+      VERSION ${HUNTER_Protobuf_VERSION}
+      CMAKE_ARGS
+      protobuf_BUILD_PROTOC=OFF
+      protobuf_BUILD_LIBPROTOBUF_LITE=OFF
```
ruslo: We can set this by default by the way: https://github.com/ruslo/hunter/wiki/usr.adding.new.package#defaults

examples/Protobuf-legacy/cmake/Hunter/config.cmake
```diff
@@ -0,0 +1,14 @@
...
+  )
+else()
+  hunter_config(Protobuf
+      VERSION ${HUNTER_Protobuf_VERSION}
+      CMAKE_ARGS
+      protobuf_BUILD_LIBPROTOBUF_LITE=OFF
```
ruslo: So `protobuf_BUILD_LIBPROTOBUF_LITE` is always `OFF`?

examples/Protobuf/messages/CMakeLists.txt
```diff
@@ -1,7 +1,25 @@
...
+  get_filename_component(basename ${proto_file} NAME_WE)
+  set(generated_files ${basename}.pb.cc ${basename}.pb.h)
 
+  list(APPEND MESSAGES_SRCS ${basename}.pb.cc)
+  list(APPEND MESSAGES_HDRS ${CMAKE_CURRENT_BINARY_DIR}/${basename}.pb.h)
 
+  add_custom_command(
+      OUTPUT ${generated_files}
+      COMMAND protobuf::protoc
+      ARGS --cpp_out ${CMAKE_CURRENT_BINARY_DIR} -I ${CMAKE_CURRENT_SOURCE_DIR} ${proto_file_abs}
```
ruslo: We don't need `ARGS`. From [documentation](https://cmake.org/cmake/help/latest/command/add_custom_command.html):

```
The optional ARGS argument is for backward compatibility and will be ignored
```

examples/Protobuf-legacy/cmake/Hunter/config.cmake
fwinnen: We don't need it in the example, usually you only need either the full or lite version.
I guess hunter defaults should still be to build both, so there won't be 2 different installs. 
I skipped the lite library here just to improve build time and show this feature in the example.

ruslo:
I've tried to run some tests, there are some mess with the project location:
```
-- [hunter] Building host project: cmake/host

CMake Error: The source directory "/home/travis/build/ingenue/hunter/_testing/cmake/host" does not exist.
```

fwinnen:
Thats a problem with relative paths. I can fix this by either:

* change `cmake/host` to `${CMAKE_CURRENT_SOURCE_DIR}/cmake/host` in user code
* have some magic in hunter_experimental_add_host_project that dies this for relative paths

Do you think the first one is enough?

ruslo:
Use `CMAKE_CURRENT_LIST_DIR` to create full path. See http://cgold.readthedocs.io/en/latest/tutorials/control-structures.html#cmake-current-list-dir. Check that path [is absolute](https://cmake.org/cmake/help/latest/command/if.html?highlight=is_absolute) and add directory if it's not.

fwinnen:
The more I think about it, i think that we should be as close as `add_subdirectory`, to not confuse users.

From the [source](https://github.com/Kitware/CMake/blob/master/Source/cmAddSubDirectoryCommand.cxx) you see, that they reinterpret relative paths as relative to ${CMAKE_CURRENT_SOURCE_DIR}. 

Also we could consider adding the relative part (cmake/host i.e) to our predefined binary directory, so it is possible to add two `hunter_experimental_add_host_project`...

ruslo:
> EDIT: I guess thats the same like you suggest

Yes, in our situation it's the same.

> Also we could consider adding the relative part (cmake/host i.e) to our predefined binary directory

Okay, in this case check that path is relative. Otherwise we will end with `binary_dir` sub-option :) Want to keep things simple.

fwinnen:
> Okay, in this case check that path is relative. Otherwise we will end with binary_dir sub-option :) 

You want out of source path to be an error or default to our binary directory? 
For me it would be fine to not allow those and thus guarantee a subdirectory

ruslo:
I mean allow only relative paths, so we can do:
* `host1`: read `host1/CMakeLists.txt`, build tree `${CMAKE_BINARY_DIR}/_3rdParty/Hunter/host-build-binary-tree/host1`
* `host2`: read `host2/CMakeLists.txt`, build tree `${CMAKE_BINARY_DIR}/_3rdParty/Hunter/host-build-binary-tree/host2`

> set(binary_directory "${CMAKE_BINARY_DIR}/_3rdParty/Hunter/host-build-binary-tree")

Should be `CMAKE_CURRENT_BINARY_DIR` if we have two `hunter_experimental_add_host_project(foo)` in different directories.

fwinnen:
> > set(binary_directory "${CMAKE_BINARY_DIR}/_3rdParty/Hunter/host-build-binary-tree")

> Should be CMAKE_CURRENT_BINARY_DIR if we have two hunter_experimental_add_host_project(foo) in different directories.

Yeah, that would clash.

> I mean allow only relative paths, so we can do:

Hehe, thats a lot easier and fixes the problem i was trying to solve the last 10 minutes, which was to figure out if a absolute path is in the current subtree. Only thing i found was checking if computed relative path contains `..`:D

fwinnen:
Well `hunter_experimental_add_host_project(../../OtherProject/foo)`
would be relative too, but we would not want to add that to our binary dir, so i guess we have to check for `..` anyway.

Or did i miss something?

ruslo:
> i guess we have to check for `..` anyway

I think [get_filename_component](https://cmake.org/cmake/help/latest/command/get_filename_component.html) can be useful here.

fwinnen:
That should deal with:

* `${CMAKE_CURRENT_SOURCE_DIR}/foo`
* `foo`
* `${CMAKE_CURRENT_SOURCE_DIR}/Dir/../Dir/foo` (may be some concatenation from variables)

And reject everything which is out of source

fwinnen:
Last thought in that regard:

lets assume the project:

* foo/CMakeList.txt                     [Entry point with `project(..)`]
* foo/host/CMakeList.txt          
* foo/sub/CMakeList.txt             [NO `project(...)` simply for organization]
* foo/sub/host/CMakeList.txt 
* foo/bar/CMakeList.txt             [WITH `project(...)` like separate component]
* foo/bar/host/CMakeList.txt   

With the structure now we would get:
* foo-bin/_3rdParty/Hunter/host-build-binary-tree/host
* foo-bin/**sub**/_3rdParty/Hunter/host-build-binary-tree/host
* foo-bin/bar/_3rdParty/Hunter/host-build-binary-tree/host

But maybe we want:
* foo-bin/_3rdParty/Hunter/host-build-binary-tree/host
* foo-bin/_3rdParty/Hunter/host-build-binary-tree/**sub**/host
* foo-bin/bar/_3rdParty/Hunter/host-build-binary-tree/host

Hope its not too hard to read.
Maybe i put too much thoughts into this and nobody will care :D

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.834
* https://travis-ci.org/ingenue/hunter/builds/154440008

ruslo:
> Maybe i put too much thoughts into this and nobody will care

Kind of :) The main point is that directories differ, the rest is of no concern.

fwinnen:
>  error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MT_StaticRelease' doesn't match value 'MD_DynamicRelease'

Looks like libprotobuf and the project are using different runtime libraries (static vs. dynamic).
I never compiled anything with cmake on windows, but ill try to figure out how to set the library type...

ruslo:
Somebody force `/MD` instead of default `/MT`. Windows testing was disabled before. If you can make a fix it would be nice, otherwise I will exclude toolchains again.

fwinnen:
Should work now, have no windows machine to test this though

ruslo:
> have no windows machine to test this though

Works on my local machine. Let's wait for Travis to finish before next iteration.

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.835
* https://travis-ci.org/ingenue/hunter/builds/154513683

ruslo:
Here comes more problems :)

`HUNTER_ROOT` is not populated so two projects use different roots:

```
       dst='/home/travis/build/ingenue/hunter/_Base/Download/Protobuf/3.0.0-p1/e3d5ab6/v3.0.0-hunter-3.tar.gz'
```

```
       dst='/home/travis/build/ingenue/hunter/_testing/Hunter/_Base/Download/Protobuf/3.0.0-p1/e3d5ab6/v3.0.0-hunter-3.tar.gz'
```

We are polluting user's directories with testing packages.

fwinnen:
> HUNTER_ROOT is not populated so two projects use different roots

Can you point me where you found the first log? I saw you set `HUNTER_ROOT` in cmake, i guess i have to check some Hunter related variables and pass them through to the host`s `execute_process`.

ruslo:
> Can you point me where you found the first log? 

It's Android + Linux build: https://travis-ci.org/ingenue/hunter/jobs/154513689

> I saw you set HUNTER_ROOT in cmake, i guess i have to check some Hunter related variables and pass them through to the hostsexecute_process`.

It's a tricky part, let's have just `HUNTER_ROOT` for now. I.e. `execute_process(... -DHUNTER_ROOT=${HUNTER_ROOT} ...)`

fwinnen:
That show another problem, which is that we ignore *all* arguments passed to cmake in the first place. For Hunter it's easy to forward its variables, but there may be some user defined args which need to be passed through..
* Add `ARGS` to hunter_add_host_project
* Add some var which forwards that to host `-DHUNTER_FWD_HOST="protobuf_BUILD_TEST=ON"`

But yeah, for now i fwd:
* `-DHUNTER_STATUS_DEBUG`
* `-DHUNTER_ROOT`
*  `-DTESTING_URL`
 * `-DTESTING_SHA1`
 * `-DHUNTER_RUN_INSTALL` ?

ruslo:
> we ignore all arguments passed to cmake in the first place

If we pass all arguments we will have toolchain build :)

> Add some var which forwards that to host -DHUNTER_FWD_HOST="protobuf_BUILD_TEST=ON"

Job for the host's `HunterGate`, not the target.

> -DTESTING_URL -DTESTING_SHA1

Used only on testing, make no sense. However such variables need to be forwarded to host project. Let it be HUNTER_HOST_URL and HUNTER_HOST_SHA1 for now.

```cmake
# host/CMakeLists.txt
HunterGate(URL "${HUNTER_HOST_URL}" SHA1 "${HUNTER_HOST_SHA1}" ...)
```

```cmake
# hunter_experimental_...
execute_process(... -DHUNTER_HOST_URL=${HUNTER_GATE_URL} -DHUNTER_HOST_SHA1=${HUNTER_GATE_SHA1})
```

> -DHUNTER_STATUS_DEBUG -DHUNTER_ROOT -DHUNTER_RUN_INSTALL

Yes. Well actually all user's settings. Let's have some small set for now. In future this should be common cache file like `config.cmake`, `cache.cmake`, `args.cmake`, etc. `HunterGate` should be refactored so it can understand that file.

fwinnen:
Ok, thanks. Will implement that.

Tests look good so far, except some windows stuff...
Do you know whats wrong with [this](https://ci.appveyor.com/project/ingenue/hunter/build/1.0.835/job/1l16t0dya2mny9co#L6781)? Is this  a valid command on windows? Those paths look weird.

ruslo:
> Those paths look weird

Yes, it's MSYS - kind of Unix on Windows :) Most of the tools is not designed to be used on MSYS. I will exclude it from testing.

ruslo:
> https://travis-ci.org/ingenue/hunter/builds/154513683
> https://ci.appveyor.com/project/ingenue/hunter/build/1.0.837

Shiny green! Once HUNTER_ROOT is fixed I'm ready to merge this.

fwinnen:
Hope this is it :)

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.838
* https://travis-ci.org/ingenue/hunter/builds/154661575

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.839
* https://travis-ci.org/ingenue/hunter/builds/154704278

Osse:
Hi, guys.

Other stuff got in the way so I haven't been able to work on this. Does this PR include (or replace) the work  I did to add protobuf 3.0.0?

Also a big thanks for working on dealing with cross compilation! This was something I've thought about but couldn't find a solution for.

ruslo:
> Does this PR include (or replace) the work I did to add protobuf 3.0.0?

Release is here https://github.com/hunter-packages/protobuf/commit/c3baf804170a1c9902a470a8ebb8a2532a4f1996

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.15.19

@fwinnen I guess it would be nice to have a stand-alone example without all this `../../../HunterGate` magic, can be added to wiki and related issue.

fwinnen:
Perfect! Thank you very much for you support and suggestions! 

I was actually thinking about a bigger standalone example to showcase the benefits of the cross platform, cross compiling support with ease. Perfect proof of concept would be a minimal application for ios/android/osx/linux/(windows) all build with hunter and polly toolchains. [+Djinni for java, objc interfaces].

That would be some work though and I have not much time atm.

What would your standalone example cover?

ruslo:
> Perfect proof of concept would be a minimal application for ios/android/osx/linux/(windows) all build with hunter and polly toolchains. [+Djinni for java, objc interfaces].

Just for your info:
* https://github.com/ruslo/weather
* https://github.com/skypjack/djinni-cmake

> What would your standalone example cover?

What I mean is to have standalone Protobuf example, just what we have now in `example/Protobuf` but without all testing stuff.

fwinnen:
Alright, if you create a Protobuf repository with your forexample account, i will push the example as standalone version.

ruslo:
Well it's not necessarily must be `forexample` but okay, done: https://github.com/forexample/protobuf-toolchains-mix

fwinnen:
Done, hope that is how you meant it.

ruslo:
Thanks, wiki page updated: https://github.com/ruslo/hunter/wiki/pkg.protobuf

Issue #503, opened 2016-08-29
Build fail with lib in lib64 instead of lib/


dsarrut:
Hello, 

I tried to use Hunter for a project that require odb (https://github.com/ruslo/hunter/tree/master/examples/odb-sqlite), but it fails after compilation. I'm on Linux OpenSuse 13.2 (3.16.7-42-desktop)

Here is my CmakeLists.txt:
```
cmake_minimum_required(VERSION 3.0)
include("cmake/HunterGate.cmake")
HunterGate(
    URL "https://github.com/ruslo/hunter/archive/v0.15.20.tar.gz"
    SHA1 "1771b5509cf8645bdb1c7ac12bc5d29a35465d12"
)
project(myproject)
hunter_add_package(Boost COMPONENTS date_time system filesystem)
hunter_add_package(odb-compiler)
hunter_add_package(odb-sqlite)
```

And the last part of the error message:
```
-- [hunter] Build step successful (dir: /home/dsarrut/.hunter/_Base/1771b55/cef4daf/c658a14/Build/odb)
[hunter ** FATAL ERROR **] File not exists:
[hunter ** FATAL ERROR **]   /home/dsarrut/.hunter/_Base/1771b55/cef4daf/c658a14/Build/odb/Install/lib/libodb.la
[hunter ** FATAL ERROR **] (check HUNTER_PACKAGE_UNRELOCATABLE_TEXT_FILES)
[hunter ** FATAL ERROR **] [Directory:/home/dsarrut/.hunter/_Base/Download/Hunter/0.15.20/1771b55/Unpacked/cmake/projects/odb]
```

I checked and libodb.la exist but in lib64/ not in lib/. Do you have an idea to correct that ? 

thanks !
David

pretyman:
Seems relevant: http://stackoverflow.com/questions/9247769/lib-directory-on-64-bit-opensuse-distribution

Looks like an OpenSuse issue, I guess we could add `--libdir` to the autotools scheme

pretyman:
Or even better could be clear the variable LIBDIR which *might* be present in OpenSuse in the clear vars script. I have to download an Image and run it in a VM to test it out

ruslo:
@pretyman Is https://github.com/ruslo/hunter/pull/507 fixing this one?

pretyman:
Yep

ruslo:
> Yep

[...]

PR #507, opened 2016-08-31
Autotools: clear CONFIG_SITE environment variable as Linux distros can override it

pretyman:
An example of a distribution that overrides it is OpenSuse:
https://github.com/ruslo/hunter/issues/503

Tested this change in OSX 10.11.2 and OpenSuse. Could be good idea to run tests for pkg.glfw as it compiles a lot of x11 packages which rely on autotools.


pretyman:
Meh, I think this will (further) break iOS Autotools integration, need to test more, don't merge :|

Sorry, will have a look at this better over the week.

pretyman:
Best solution would be to package a `config.site` file together with hunter: https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/Site-Defaults.html#Site-Defaults and set the environment variable CONFIG_SITE pointing to it.

ruslo:
> Could be good idea to run tests for pkg.glfw as it compiles a lot of x11 packages which rely on autotools.

We will test nothing until you change INTERNAL_DEPS_ID of affected packages :)

pretyman:
> We will test nothing until you change INTERNAL_DEPS_ID of affected packages :)

I expect that this change does not alter the contents of the compiled packages, change anyway?

ruslo:
> I expect that this change won't alter the contents of the compiled packages

No need if you're 100% sure. But in this case there make no sense to run tests as you suggested before: "pkg.glfw as it compiles a lot of x11 packages which rely on autotools" - they all will be taken from cache.

pretyman:
Updated this pull request.

For now the less intrusive fix is to clear `CONFIG_SITE` that is set in OpenSuse. If such issue pops up again in the future, then we can take more drastic measures and create hunter's down `config.site`.

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.15.23

Issue #525, opened 2016-09-19
Hunter project flatbuffers not found


ahundt:
I tried running hunter with flatbuffers and got the following error:

```
-- [hunter] Initializing Hunter workspace (53b198e364dc7bc8360fc545f798563229bd7e20)
-- [hunter]   https://github.com/ruslo/hunter/archive/v0.10.9.tar.gz
-- [hunter]   -> /Users/athundt/.hunter/_Base/Download/Hunter/0.10.9/53b198e
-- [hunter] Calculating Config-SHA1
-- [hunter] Calculating Toolchain-SHA1
-- [hunter] HUNTER_ROOT: /Users/athundt/.hunter
-- [hunter] [ Hunter-ID: 53b198e | Config-ID: 1c72480 | Toolchain-ID: a651279 ]

[hunter ** INTERNAL **] Project 'flatbuffers' not found
[hunter ** INTERNAL **] [Directory:/Users/athundt/source/distributed_systems_cs437]

------------------------------ WIKI -------------------------------
    https://github.com/ruslo/hunter/wiki/error.internal
-------------------------------------------------------------------

CMake Error at /Users/athundt/.hunter/_Base/Download/Hunter/0.10.9/53b198e/Unpacked/cmake/modules/hunter_wiki.cmake:12 (message):
Call Stack (most recent call first):
  /Users/athundt/.hunter/_Base/Download/Hunter/0.10.9/53b198e/Unpacked/cmake/modules/hunter_internal_error.cmake:13 (hunter_wiki)
  /Users/athundt/.hunter/_Base/Download/Hunter/0.10.9/53b198e/Unpacked/cmake/modules/hunter_add_package.cmake:30 (hunter_internal_error)
  CMakeLists.txt:31 (hunter_add_package)


-- Configuring incomplete, errors occurred!
See also "/Users/athundt/source/distributed_systems_cs437/build/CMakeFiles/CMakeOutput.log".
```

I based this off of the [example script](https://github.com/ruslo/hunter/blob/develop/examples/flatbuffers/CMakeLists.txt), and the repository it appears it should clone seems to be a bit out of date compared to the mainline flatbuffers repository.

This is on OS X 10.11.6.

ruslo:
> I tried running hunter with flatbuffers and got the following error

Use newer version of archive: https://github.com/ruslo/hunter/releases

> the repository it appears it should clone seems to be a bit out of date compared to the mainline flatbuffers repository

[...]

PR #529, opened 2016-09-25
Update flatbuffers hunter.cmake for 1.4.0-p1

ahundt:


ahundt:
Should resolve:
https://github.com/ruslo/hunter/issues/525

Until the following is merged upstream:
https://github.com/google/flatbuffers/pull/4035

When that happens the upstream version should be used.

cmake/projects/flatbuffers/hunter.cmake
```diff
@@ -13,19 +13,19 @@ hunter_add_version(
     PACKAGE_NAME
     flatbuffers
     VERSION
-    1.3.0-p3
+    1.4.0-p1
```
ruslo: Please add new version instead of rewriting. Old version still can be specified by custom user config: https://github.com/ruslo/hunter/wiki/example.custom.config.id#custom-package-version

Set new version as default here: https://github.com/ruslo/hunter/blob/master/cmake/configs/default.cmake

ahundt: should be fixed now

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.999
* https://travis-ci.org/ingenue/hunter/builds/162630889

ruslo:
> Testing

Some tests failed. On Linux clang-libstdcxx, sanitize-* tests failed because flatbuffers decide to add `-std=c++0x -stdlib=libc++` even if toolchain set `-std=c++11 -stdlib=libstdc++`, this is an issue in upstream. Static analyzer test log: https://travis-ci.org/ingenue/hunter/jobs/162630915 need to investigate is it a real issues or false positives.

@ahundt You can report/fix issues or exclude these tests by sending pull request to [ingenue/hunter@pkg.flatbuffers](https://github.com/ingenue/hunter/tree/pkg.flatbuffers) branch. Please comment out broken tests and add link to the logs so if anybody who decide to fix them will have easy access. See for example Boost: https://github.com/ingenue/hunter/blob/pkg.boost/.travis.yml

ahundt:
ah, just came back to this. I've been using the for used for the pull request without problems in my situation, but I can see how that is a genuine issue. I reported it upstream https://github.com/google/flatbuffers/issues/4038

ruslo:
> I reported it upstream google/flatbuffers#4038

If you're not planning to fix it yourself then I think it's better to exclude those tests. Otherwise this pull request will be openned for an infinite amount of time :) Also the fix is trivial:

```cmake
if(EXISTS "${CMAKE_TOOLCHAIN_FILE}")
  # do not apply any global settings
else()
  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x")
endif()
```

ahundt:
Thanks, added a pull request for that change https://github.com/google/flatbuffers/pull/4045

ahundt:
The pull request has been merged upstream. Should finishing this (1) wait for the next point release, (2) go to some commit in fb master, (3) apply a patch externally?

ruslo:
> The pull request has been merged upstream. Should finishing this (1) wait for the next point release, (2) go to some commit in fb master, (3) apply a patch externally?

@ahundt You can proceed with anything that you think is better.

ahundt:
Waiting for 1.5.0 will be the most straightforward option, let's go with that. Wish it was possible to set alerts for this sort of thing :-)

ruslo:
> Wish it was possible to set alerts for this sort of thing :-)

You can, just subscribe to atom feed: https://github.com/google/flatbuffers/releases.atom

ahundt:
> You can, just subscribe to atom feed: https://github.com/google/flatbuffers/releases.atom
@ruslo amazing! Thanks!

ahundt:
hmm seems 1.5 still is not out yet...

ruslo:
> ruslo closed this 10 days ago

I've removed branch `develop` (now we can just use `master`) and it was closed automatically. Reopenning.

> hmm seems 1.5 still is not out yet...

We can always return to the fork if you want.

ruslo:
https://github.com/google/flatbuffers/releases/tag/v1.6.0

ruslo:

PR #554, opened 2016-10-19
Fix Android/iOS compile for CURL

xsacha:
There are a few tests that the CMakeLists attempts to run on the host system. By setting the variables to what the result would be, it avoids this attempt and successfully compiles.
Tested on Android.

### DIFF
--- a/cmake/projects/CURL/hunter.cmake
+++ b/cmake/projects/CURL/hunter.cmake
@@ -27,6 +27,17 @@ hunter_add_version(
 )
 
 
+if (ANDROID OR IOS)
+  set(_curl_cmake_args
+      HAVE_GLIBC_STRERROR_R=1
+      HAVE_GLIBC_STRERROR_R__TRYRUN_OUTPUT=""
+      HAVE_POSIX_STRERROR_R=0
+      HAVE_POSIX_STRERROR_R__TRYRUN_OUTPUT=""
+      HAVE_POLL_FINE_EXITCODE=0
+  )
+else()
+  set(_curl_cmake_args "")
+endif()
 
 hunter_cmake_args(
     CURL
@@ -35,7 +46,8 @@ hunter_cmake_args(
         BUILD_CURL_EXE=OFF
         CMAKE_USE_OPENSSL=ON
         CMAKE_USE_LIBSSH2=OFF
- )
+        ${_curl_cmake_args}
+)
 
 
 


docs/rtfd-css
```diff
@@ -1 +1 @@
-Subproject commit ea8d83e77ef1c07ebd31f2e6dbf7c081f55ce06b
```
ruslo: Not related I guess.

cmake/projects/CURL/hunter.cmake
```diff
@@ -27,15 +27,30 @@ hunter_add_version(
...
+            HAVE_GLIBC_STRERROR_R__TRYRUN_OUTPUT=""
+            HAVE_POSIX_STRERROR_R=0
+            HAVE_POSIX_STRERROR_R__TRYRUN_OUTPUT=""
+            HAVE_POLL_FINE_EXITCODE=0
+     )
+else()
+    hunter_cmake_args(
+        CURL
+        CMAKE_ARGS
+            BUILD_CURL_TESTS=OFF
```
ruslo: This part is common for both configuration. Use separate variable like `_curl_cmake_args`:

```cmake
set(_curl_cmake_args BUILD_CURL_TESTS=OFF BUILD_CURL_EXE=OFF ...)
if(ANDROID OR IOS)
  list(APPEND _curl_cmake_args HAVE_GLIBC_STRERROR_R=1 HAVE_GLIBC_STRERROR_R__TRYRUN_OUTPUT="" ...)
endif()
```

xsacha:
Thanks. Updated.

cmake/projects/CURL/hunter.cmake
```diff
@@ -27,6 +27,15 @@ hunter_add_version(
 )
 
 
+if (ANDROID OR IOS)
+  set(_curl_cmake_args
```
ruslo: If somebody set this variable outside and platform is not Android/iOS we can carry junk to the option. Just add `set(_curl_cmake_args "")` for `else` condition.

xsacha: Fixed that. Thanks.

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1086
* https://travis-ci.org/ingenue/hunter/builds/169129812

Note that there is no iOS/Android toolchain in matrix, we do test only the fact that change doesn't broke old code. After this pull request will be merged feel free to send pull request to [pkg.curl](https://github.com/ingenue/hunter/tree/pkg.curl) branch with uncommented iOS/Android entries.

ruslo:
Will be released with https://github.com/ruslo/hunter/pull/552

PR #565, opened 2016-10-24
Add ceres-solver

tbaudier:
Add ceres-solver package from https://github.com/hunter-packages/ceres-solver in Hunter
thx
Thomas


ruslo:
Can you please add an example so we can run automated tests, see https://github.com/ruslo/hunter/blob/master/examples/GTest/CMakeLists.txt for the hints

cmake/projects/ceres-solver/hunter.cmake
```diff
@@ -0,0 +1,29 @@
...
+include(hunter_pick_scheme)
+
+# List of versions here...
+hunter_add_version(
+    PACKAGE_NAME
+    ceres-solver
+    VERSION
+    "1.12.0"
+    URL
+    "https://github.com/hunter-packages/ceres-solver/archive/hunter.zip"
```
ruslo: This archive is not "stable", after you push updates to branch `hunter` it will change. Use `tar.gz` archive from releases: https://github.com/hunter-packages/ceres-solver/releases/tag/1.12.0rc1

cmake/projects/ceres-solver/hunter.cmake
```diff
@@ -0,0 +1,29 @@
...
+    SHA1
+    8554129d92cdddaead42274a47322709f82943e0
+)
+
+# Probably more versions for real packages...
+
+# Pick a download scheme
+hunter_pick_scheme(DEFAULT url_sha1_cmake) # use scheme for cmake projects
```
ruslo: Is this package cacheable/relocatable? See details: https://github.com/ruslo/hunter/wiki/usr.adding.new.package#important

examples/ceres-solver/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+cmake_minimum_required(VERSION 3.0)
+
+include("../common.cmake")
+
+project(download-ceres-solver)
+
+hunter_add_package(ceres-solver)
+
+find_package(Ceres REQUIRED)
```
ruslo: I think it's better to explicitly mark CeresConfig.cmake usage by `find_package(Ceres CONFIG REQUIRED)`

cmake/projects/ceres-solver/hunter.cmake
tbaudier: I added hunter_cacheable

cmake/configs/default.cmake
```diff
@@ -92,6 +92,7 @@ hunter_config(ZeroMQ VERSION 4.1.4-p2)
 hunter_config(caffe VERSION rc3-p2)
 hunter_config(ccv VERSION 0.7-p6)
 hunter_config(cereal VERSION 1.1.2-p5)
+hunter_config(ceres-solver VERSION 1.12.0)
```
ruslo: Version not found in `hunter.cmake`

ruslo: Use [jenkins.py](https://github.com/ruslo/hunter/wiki/dev.hunter.root#jenkinspy) testing to avoid such issues.

ruslo:
@tbaudier I see you have pushed new commits. Please let me know explicitly that new change are ready to be tested and is not work-in-progress status. Thanks.

tbaudier:
@ruslo I change the version of ceres to the correct one and I did the modifications you proposed. You can test it.
thx
Thomas

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1138
* https://travis-ci.org/ingenue/hunter/builds/170696205

tbaudier:
I saw it fails, I'll look at that

ruslo:
> I saw it fails, I'll look at that

Okay, thanks. Note that you don't have to fix everything, you can just exclude toolchain from testing so everything became "green". Then you can test the fixes one by one. Send pull request with notes/excludes to branch https://github.com/ingenue/hunter/tree/pkg.ceres-solver . For example see Boost FIXME comments: https://github.com/ingenue/hunter/blob/pkg.boost/.travis.yml

NeroBurner:
I was able to build ceres-solver from your fork with minor tweaks. I needed to add the following line to `internal/ceres/CMakeLists.txt`
```
if(HUNTER_ENABLED)
	target_link_libraries(ceres Eigen::eigen)
endif()
```
`target_link_libraries` is neccessary because of the custom hunter-FindEigen.cmake file, which only defines the imported target `Eigen::eigen` (see !623)

it seems, that ceres-solver has a hard dependency on Eigen3, but has no check error when Eigen3 is not found. After setting it build with
```
cmake -H. -B_build -DBUILD_EXAMPLES=OFF -DBUILD_SHARED_LIBS=OFF -DBUILD_TESTING=OFF
cmake --build _build
```

PR #567, opened 2016-10-24
Hunter side of "Setup static lib build" for dlib

caseymcc:
see https://github.com/hunter-packages/dlib/pull/5

Setup static lib build for dlib, currently referencing my branch of dlib for testing. Will change for final.


cmake/projects/dlib/hunter.cmake
```diff
@@ -5,9 +5,22 @@
...
+#temp redirect till branch is accpeted
+hunter_add_version(
+    PACKAGE_NAME
+    dlib
+    VERSION
+    "19.0-p2"
+    URL
+    "https://github.com/caseymcc/dlib/archive/v19.0-p2.tar.gz"
+    SHA1
+    0f88871c06292521ba3f27617c80620d00b4dd8b
```
ruslo: doesn't match, from my experience it's 60f57478a639625c429d092e2d8dc6176d2d35f1

cmake/projects/Assimp/hunter.cmake
```diff
@@ -1,45 +0,0 @@
-# Copyright (c) 2016, Ruslan Baratov
```
ruslo: Please revert this file back :)

caseymcc:
This should be good to go now.

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1164
* https://travis-ci.org/ingenue/hunter/builds/175090530

caseymcc:
hmm, seems VS2012 does not support intializer list (https://msdn.microsoft.com/en-us/library/hh567368.aspx) and likely the reason for the fail as they are keying off of rvalue support which 2012 does have. The oddity though is that VS2010 succeeded and it has the same language features. I can add an extra flag to check for initializes but I an not sure how to test the results as this is my first time to ever use travis/appveyor. Still looking into the clang failure.

caseymcc:
I fixed the issue on the windows side (https://github.com/hunter-packages/dlib/pull/6) and ran it through appveyor with no issues (seems the guys at dlib also hit this with the latest version). The clang issue however is a little harder to resolve, I hoped that the same change for windows would work but no go. Apparently the issue is osx/clang only and because clang on osx links with an older libstdc++ that does not include initializer_lists (http://stackoverflow.com/questions/35599150/apple-clang-using-c11-with-libstdc) but it still reports initializer lists support. I don't believe there is a way around it so it maybe better to remove that test as the libcxx one works fine and using libcxx is the fix for the issue.

ruslo:
> I can add an extra flag to check for initializes but I an not sure how to test the results as this is my first time to ever use travis/appveyor

> I fixed the issue on the windows side

So you figured out how to deal with AppVeyor?

> I don't believe there is a way around it so it maybe better to remove that test as the libcxx one works fine and using libcxx is the fix for the issue

In this case the upstream documentation should help because if they do support old compilers/libs then it should be fixed (or at least reported). If they do not support such compilers/libs then you should leave a link to this info to the `.travis.yml`. As a general note I think there is a bug in feature detection logic because `std::initializer_list` is about library and not about compiler (and you're checking only compiler's version). Probably Boost can help?

caseymcc:
Yep (on Appveyor/Travis) figured out enough to test it. Looks pretty good will have to look into using it in future projects.

I can't find documentation on dlib's supported compilers. I will look into boost's checks and see if there is one for this situation. They look to be using one with the intel compiler.

caseymcc:
assuming this will past the tests, should be good to go.

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1178
* https://travis-ci.org/ingenue/hunter/builds/176073261

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1181
* https://travis-ci.org/ingenue/hunter/builds/176201452

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.16.30

Issue #566, opened 2016-10-24
include HunterGate fail with VS2015 x64 generator


Ubiquite:
[hunter *\* INTERNAL **] vcvarsall.bat not found in `/VC`
[hunter *\* INTERNAL **] [Directory:C:/.hunter/_Base/Download/Hunter/0.10.9/53b198e/Unpacked/cmake]

Ubiquite:
OK it's just because the documentation in out of date : https://docs.hunter.sh/en/latest/quick-start/boost-components.html
should be update with
HunterGate(
    URL "https://github.com/ruslo/hunter/archive/v0.16.15.tar.gz"
    SHA1 "6974c2150fc0d3b09de3ad1efcbf15d360647ffa"
)
I guess

ruslo:
> OK it's just because the documentation in out of date : https://docs.hunter.sh/en/latest/quick-start/boost-components.html
should be update with

Feel free to send the patch.

ruslo:
Closing inactive discussion.

PR #568, opened 2016-10-25
update doc with new hunter archive

Ubiquite:
following issue 566
https://github.com/ruslo/hunter/issues/566


ruslo:
Applied: https://docs.hunter.sh/en/latest/quick-start/boost-components.html

Issue #230, opened 2015-09-14
hunter_config: specify package version by Git


ruslo:
Allow SHA1 of commit of git repo:

``` cmake
hunter_config(Foo GIT_SHA1 xxx GIT_REPO https://.../Foo.git)
```

Allow getting GIT_SHA1 from submodule:

``` cmake
hunter_config(Foo GIT_SUBMODULE)
```

_Originally from here: https://github.com/ruslo/hunter/issues/56_
##

ruslo:
Can be used as a workaround for https://github.com/ruslo/CGold/issues/8 (?)

ruslo:
> hunter_config(Foo GIT_SUBMODULE)

See https://docs.hunter.sh/en/latest/user-guides/hunter-user/git-submodule.html

> hunter_config(Foo GIT_SHA1 xxx GIT_REPO https://.../Foo.git)

I don't think it's very convenient to update GIT_SHA1 manually each time, submodule seems easier to manage.

PR #573, opened 2016-10-28
Add scheme to use local cmake source directory as package

caseymcc:
I mainly used it for testing a package to prep it for hunter inclusion. Currently it copies the source using the value of HUNTER_PACKAGE_URL into HUNTER_PACKAGE_DOWNLOAD_DIR while filtering the .git folder (as it was causing issues deleting the folder through cmake). From there the ExternalProject copies it to HUNTER_PACKAGE_SOURCE_DIR as needed.

It ignores any timestamp/SHA/etc... so the source should be copied each time from the original folder. Including this for test purposes only as for production it really should include timestamps/SHA/etc...

### DIFF
--- a/cmake/modules/hunter_download.cmake
+++ b/cmake/modules/hunter_download.cmake
@@ -117,6 +117,7 @@ function(hunter_download)
   # Check that only one scheme is set to 1
   set(all_schemes "")
   set(all_schemes "${all_schemes}${HUNTER_PACKAGE_SCHEME_DOWNLOAD}")
+  set(all_schemes "${all_schemes}${HUNTER_PACKAGE_SCHEME_LOCAL_CMAKE}")
   set(all_schemes "${all_schemes}${HUNTER_PACKAGE_SCHEME_UNPACK}")
   set(all_schemes "${all_schemes}${HUNTER_PACKAGE_SCHEME_UNPACK_INSTALL}")
   set(all_schemes "${all_schemes}${HUNTER_PACKAGE_SCHEME_INSTALL}")
@@ -126,6 +127,7 @@ function(hunter_download)
     hunter_internal_error(
         "Incorrect schemes:"
         "  HUNTER_PACKAGE_SCHEME_DOWNLOAD = ${HUNTER_PACKAGE_SCHEME_DOWNLOAD}"
+        "  HUNTER_PACKAGE_SCHEME_LOCAL_CMAKE = ${HUNTER_PACKAGE_SCHEME_LOCAL_CMAKE}"
         "  HUNTER_PACKAGE_SCHEME_UNPACK = ${HUNTER_PACKAGE_SCHEME_UNPACK}"
         "  HUNTER_PACKAGE_SCHEME_UNPACK_INSTALL = ${HUNTER_PACKAGE_SCHEME_UNPACK_INSTALL}"
         "  HUNTER_PACKAGE_SCHEME_INSTALL = ${HUNTER_PACKAGE_SCHEME_INSTALL}"
@@ -193,6 +195,9 @@ function(hunter_download)
     if(HUNTER_PACKAGE_SCHEME_DOWNLOAD)
       set(${root_name} "${HUNTER_PACKAGE_SOURCE_DIR}")
       hunter_status_debug("Download to: ${HUNTER_PACKAGE_SOURCE_DIR}")
+    elseif(HUNTER_PACKAGE_SCHEME_LOCAL_CMAKE)
+      set(${root_name} "${HUNTER_PACKAGE_SOURCE_DIR}")
+      hunter_status_debug("Copy to: ${HUNTER_PACKAGE_SOURCE_DIR}")
     elseif(HUNTER_PACKAGE_SCHEME_UNPACK)
       set(${root_name} "${HUNTER_PACKAGE_SOURCE_DIR}")
       hunter_status_debug("Unpack to: ${HUNTER_PACKAGE_SOURCE_DIR}")
@@ -502,7 +507,7 @@ function(hunter_download)
     )
   endif()
 
-  if(HUNTER_PACKAGE_SCHEME_DOWNLOAD)
+  if(HUNTER_PACKAGE_SCHEME_DOWNLOAD OR HUNTER_PACKAGE_SCHEME_LOCAL_CMAKE)
     # This scheme not using ExternalProject_Add so there will be no stamps
   else()
     hunter_find_stamps("${HUNTER_PACKAGE_BUILD_DIR}")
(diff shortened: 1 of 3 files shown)


ruslo:
I'm not sure what/why is happenning here. Can you describe the scenario of usage?

caseymcc:
I used it for testing. With dlib I kept the dlib source in a local directory where I made changes to the cmake setup. I setup the hunter.cmake for dlib as follows:

```
hunter_add_version(
    PACKAGE_NAME
    dlib
    VERSION
    "19.0-p2"
    URL
     "D:/projects/hunter_cmake/packages/dlibStatic"
   SHA1
    ANYTHINGWORKSITSIGNORED
)
...
hunter_pick_scheme(DEFAULT url_local_cmake)
```

this way I could make changes to dlib's cmake setup and test the results without pushing changes to github (create release, build sha, etc...). Before release the changes I would change the URL to an actual release and return to the original scheme. I am not sure it would every be used in "production" but works really nice for testing.

ruslo:
You can achieve the same by creating archive manually, like:
```
> git archive HEAD -o archive.tar
> openssl sha1 archive.tar
SHA1(archive.tar)= c5bc3cca2437885110fcae637b4ec9f2c25acb62
```

then put this info to `hunter.cmake`:
```
hunter_add_version(
    ...
    URL "file://C:/path/to/archive.tar"
    SHA1 c5bc3cca2437885110fcae637b4ec9f2c25acb62
)
```

May be even more easier way to follow is allowing SHA1 commit:
* https://github.com/ruslo/hunter/issues/230

As a general note testing of the integration of installed package is the feature that missing in CMake itself. This is something that should be implemented outside of the Hunter.

caseymcc:
Yes, you could do extra work and build a tar and compute the SHA1 every time you make a small change to the package's cmake configuration, but why? With this scheme you can temporarily change the package to use a source directory where you can test configuration changes and test the full hunter build with a single cmake call. Once you get everything working build the tar (github release) compute the SHA1 and change the scheme back.

I am not overly familiar with cmake (especially the config stuff) so it takes a little bit of tinkering for me to understand what cmake is doing and get the config working. Building a tar and computing a sha1 is a bit more todo (especially on windows) just to test something.

ruslo:
> but why?

Because it's not Hunter specific but do add some extra logic to Hunter. This scenario is not something that will be used in production widely and by accepting this I will must accept other not very popular scenarios. Number of schemes should decrease, not increase.

> With this scheme you can temporarily change the package to use a source directory where you can test configuration changes and test the full hunter build with a single cmake call

> I am not overly familiar with cmake (especially the config stuff) so it takes a little bit of tinkering for me to understand what cmake is doing and get the config working

This is understandable, I do myself experience some inconvenience while testing but like I said before this is something that should be done on package level, not in global Hunter level. Current Hunter way of testing should be a sanity tests of ready to be used packages, not the main platform for testing.

I can accept something that will be an orthogonal to the current functionality so I don't have to think about weird `url_sha1_local` build scheme every time I hack `hunter_download` function. What do you think about introducing function like `hunter_add_test_version` which will create archive from source directory so for the `hunter_download` it will be like building from regular archive with URL and SHA1?

```cmake
hunter_add_test_version(
    PACKAGE_NAME Foo
    SOURCE_DIR "C:/source/directory/Foo"
)

hunter_download(Foo)
```
)

caseymcc:
Ah, I see the objection now and it is actually one I was a bit confused by when I first looked at adding a scheme. I was surprised to see that the adding a scheme required changes in any other files. I am not sure that looking at the scheme as reducing is the right idea, I would assume overtime people would add more schemes for handling different types of build systems other than currently supported ones.

I think a better solution here would be that the references to scheme in `hunter_download.cmake` and `hunter_pick_scheme.cmake` be abstracted out to the scheme file. For example `hunter_download` should open the cmake/schemes folder and get the name of each scheme rather than having it hard coded in the file. And instead of hard coding the scheme to use `hunter_find_stamps` we call a macro in the scheme file `${SchemeName}_find_stamps` that does/does not call hunter_find_stamps as needed. Or we just call generic macros in the scheme like `${SchemeName}_download`, `${SchemeName}_build` that duplicate some of the code in `hunter_download` for each scheme. I think similar things could be done for each of the currently hard coded cases.

I am not fond of `hunter_add_test_version` as it seems like you are adding a new version rather than altering how the version is downloaded and included/built. Where as it seems the scheme is suppose to be how the package is downloaded and included/built.

ruslo:
> I was surprised to see that adding a scheme required changes in any other files

It does not if the scenario of usage is build/install. For example next schemes fit:
* [Autotools](https://github.com/ruslo/hunter/blob/master/cmake/schemes/url_sha1_autotools.cmake.in)
* [Boost](https://github.com/ruslo/hunter/tree/master/cmake/projects/Boost/schemes)
* [Qt](https://github.com/ruslo/hunter/tree/master/cmake/projects/Qt/schemes)
* [OpenSSL](https://github.com/ruslo/hunter/tree/master/cmake/projects/OpenSSL/schemes)

in general that's what packages do inside package manager, right? 😎 

> I would assume overtime people would add more schemes for handling different types of build systems other than currently supported ones

Yes and no. Yes - if there will be package with build system different from supported one then we have to introduce new scheme. By "no" I mean that even existent schemes doesn't support all functionality that CMake do support from the box. CMake is quite powerful and easy extendable and achieving the same with other build system is really hard. The plan is to wait when CMake will be used as a build system for such packages and remove custom build scheme when it's done (I'm not saying it's doable or will happen in observable future, just saying that we should follow this direction).

> For example hunter_download should open the cmake/schemes folder and get the name of each scheme rather than having it hard coded in the file

The name of the scheme may depeds on the host, platforms and even [version of the package](https://github.com/ruslo/hunter/blob/592ffe04425bf38c08b25f77571c01b2ab77b796/cmake/projects/OpenSSL/hunter.cmake#L254-L258) so what the point of moving it **inside** the `hunter_download` instead of using **outside** and works perfectly so far?

> And instead of hard coding the scheme to use hunter_find_stamps we call a macro in the scheme file ${SchemeName}_find_stamps that does/does not call hunter_find_stamps as needed

`hunter_find_stamps` function called for every scheme that use ExternalProject_Add. That's the most popular way of building. So why instead of calling it in one place in `hunter_download` we have to add absolutely similar duplicate code to the 10+ schemes?

> I am not fond of hunter_add_test_version as it seems like you are adding a new version rather than altering how the version is downloaded and included/built.

I'm not saying it's good, I'm saying that it's better than introducing new scheme because:
1. new scheme will never be used in production
2. new scheme is intrusive and for no strong reason change the internal logic

> Where as it seems the scheme is suppose to be how the package is downloaded and included/built.

That's the point, it is not. The production code will still use `url_sha1_cmake` build scheme even you're using `url_sha1_local` for testing. Now imagine you want to test OpenSSL locally. What you have to do? Introduce new scheme `url_sha1_openssl_local`? `hunter_add_test_version` will work fine in this case.

caseymcc:
url_local_cmake should work for all projects that use cmake and with source in a local directory, which I believe makes is generic enough to be usable by multiple projects. Assuming openssl needs a custom scheme I would assume its scheme needs to be with the package.

ruslo:
> url_local_cmake should work for all projects that use cmake and with source in a local directory, which I believe makes is generic enough to be usable by multiple projects.

`hunter_add_test_version` will work with all CMake projects and with all non-CMake projects :) It's even more generic.

> Assuming openssl needs a custom scheme I would assume its scheme needs to be with the package.

Not sure I understand this phrase.

PR #577, opened 2016-11-08
register OpenCV version 2.4.13.1-p1

Ubiquite:
version 2.4.13.1-p1 supports VS2015

### DIFF
--- a/cmake/projects/OpenCV/hunter.cmake
+++ b/cmake/projects/OpenCV/hunter.cmake
@@ -171,6 +171,17 @@ hunter_add_version(
     PACKAGE_NAME
     OpenCV
     VERSION
+    "2.4.13.1-p1"
+    URL
+    "https://github.com/hunter-packages/opencv/archive/v2.4.13.1-p1.tar.gz"
+    SHA1
+    c7da4588a1037743f8328007edc309087f6ce8b2
+)
+
+hunter_add_version(
+    PACKAGE_NAME
+    OpenCV
+    VERSION
     "2.4.11-p1"
     URL
     "https://github.com/hunter-packages/opencv/archive/2.4.11-p1.tar.gz"
--- a/examples/OpenCV-2.4/CMakeLists.txt
+++ b/examples/OpenCV-2.4/CMakeLists.txt
@@ -1,0 +1,19 @@
+cmake_minimum_required(VERSION 3.0)
+
+set(TESTING_CONFIG_OPT FILEPATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/Hunter/config.cmake)
+
+# Emulate HunterGate:
+# * https://github.com/hunter-packages/gate
+include("../common.cmake")
+
+project(download-opencv)
+
+hunter_add_package(OpenCV)
+find_package(OpenCV REQUIRED)
+
+message("OpenCV_DIR: ${OpenCV_DIR}")
+message("OpenCV_CONFIG: ${OpenCV_CONFIG}")
+message("OpenCV_LIBS: ${OpenCV_LIBS}")
+
+add_executable(foo foo.cpp)
+target_link_libraries(foo PRIVATE ${OpenCV_LIBS})
--- a/examples/OpenCV-2.4/cmake/Hunter/config.cmake
+++ b/examples/OpenCV-2.4/cmake/Hunter/config.cmake
@@ -1,0 +1,1 @@
+hunter_config(OpenCV VERSION 2.4.13.1-p1)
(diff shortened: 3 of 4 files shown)


ruslo:
> version 2.4.13.1-p1 supports VS2015

It would be nice if we can test it (for now tested only version from `default.cmake`, which is 3+). Please add a **new** example which will use this version, we can call it `OpenCV-2.4`. Version can be set explicitly be `config.cmake`, see example of customization: https://github.com/ruslo/hunter/blob/0edd03ed68b7a56d6f6ef0a4aa8912498daf715a/examples/Protobuf-optimized/CMakeLists.txt#L5

ruslo:
> version 2.4.13.1-p1 supports VS2015

Have you test the example yourself? It doesn't work for me and still produce error:
```
-- OpenCV ARCH: x86
-- OpenCV RUNTIME: vc14
-- OpenCV STATIC: OFF
CMake Warning at C:/.../Hunter/_Base/e0f9166/95b4ad6/7670235/Install/OpenCVConfig.cmake:163 (message):
  Found OpenCV Windows Pack but it has not binaries compatible with your
  configuration.

  You should manually point CMake variable OpenCV_DIR to your build of OpenCV
  library.
Call Stack (most recent call first):
  CMakeLists.txt:12 (find_package)


CMake Error at CMakeLists.txt:12 (find_package):
  Found package configuration file:

    C:/.../Hunter/_Base/e0f9166/95b4ad6/7670235/Install/OpenCVConfig.cmake

  but it set OpenCV_FOUND to FALSE so package "OpenCV" is considered to be
  NOT FOUND.
```

Ubiquite:
no didn't tested, I was sure it worked. 
I have that issue also. I have not idea why, it works for my project

ruslo:
> I have that issue also. I have not idea why, it works for my project

You have that issue but at the same time it works for your project? 😎

Ubiquite:
Yep I was able to compile with VS2015 and OpenCV 2.4, removing plenty of errors.
I'm investigating though

Ubiquite:
I see a difference : 
```
[hunter] OPENCV_ROOT: D:/hunter/_Base/6974c21/2288afc/78eefd3/Install (ver.: 2.4.13.1-p1)
OpenCV ARCH: x86
OpenCV RUNTIME: vc14
OpenCV STATIC: **OFF**
CMake Warning at D:/hunter/_Base/6974c21/2288afc/78eefd3/Install/OpenCVConfig.cmake:163 (message):
  Found OpenCV Windows Pack but it has not binaries compatible with your
  configuration.
```
```
[hunter] OPENCV_ROOT: D:/hunter/_Base/6974c21/4adca17/78eefd3/Install (ver.: 2.4.13.1-p1)
OpenCV ARCH: x86
OpenCV RUNTIME: vc14
OpenCV STATIC: **ON**
Found OpenCV 2.4.13.1 in D:/hunter/_Base/6974c21/4adca17/78eefd3/Install/x86/vc14/staticlib
```

ruslo:
> OpenCV STATIC: **OFF**
> OpenCV STATIC: **ON**

How this variable is set? It should depend on the fact how OpenCV is built.

Ubiquite:
Ok it works with BUILD_SHARED_LIBS ON but not OFF

For information BUILD_SHARED_LIBS OFF on my project is a side effect of integrating GTest with add_subdirectory(gtest..) which define option(BUILD_SHARED_LIBS ... OFF)

ruslo:
> Ok it works with BUILD_SHARED_LIBS ON but not OFF

With `BUILD_SHARED_LIBS` on the user's side or on the OpenCV side? What is the way to detect static library in installed OpenCVConfig.cmake? This should be based on the OpenCV's value of BUILD_SHARED_LIBS, something like this:
```cmake
# OpenCVConfig.cmake.in template
...
if(@BUILD_SHARED_LIBS@) # the value of BUILD_SHARED_LIBS on configuration step of OpenCV
  # searching for shared libraries
else()
  # searching for static libraries
endif()
```

Ubiquite:
User's side

ruslo:
> User's side

That's wrong I think. For example if I want to have OpenCV as shared and link it to my static library.

Ubiquite:
But that's the way OpenCV seems to work

ruslo:
> But that's the way OpenCV seems to work

As you can see it doesn't work correctly, that's what we hit here. I think the fix is trivial.

Ubiquite:
In the installed OpenCVConfig.cmake I have ; 
```
if(NOT DEFINED OpenCV_STATIC)
  # look for global setting
  if(BUILD_SHARED_LIBS)
    set(OpenCV_STATIC OFF)
  else()
    set(OpenCV_STATIC ON)
  endif()
endif()
```
 
Whereas in the template there is only : 
```
# Some additional settings are required if OpenCV is built as static libs
set(OpenCV_SHARED @BUILD_SHARED_LIBS@)
```

I'm confused

PR #580, opened 2016-11-12
Fix an issue in OpenSSL build when cross compiling

edtanous:
The commit 44c681e1a551e0599a8ca170cb0cfcc19c4a5d54 seems to have
broken cross compiling builds, as openssl ./config seems to require
both the CC and MACHINE variables to be set, otherwise it seems to
default to build host compiler and machine variables.  Someone appears
to have already seen this issue when cross compiling for android, and
fixed it in commit bb657ff41c874ca3bba1ca394ff1ea09ee8049b3.  This
patch moves the MACHINE argument and adds it to ALL build types, not
just android.

I'm unclear on the architecture of the PACKAGE_INTERNAL_DEPS_ID, but I saw it being incremented for other similar changes, so I assume this will require an update as well.  If I'm wrong, let me know and I will update the PR.

### DIFF
--- a/cmake/projects/OpenSSL/hunter.cmake
+++ b/cmake/projects/OpenSSL/hunter.cmake
@@ -263,4 +263,4 @@ else()
 endif()
 
 hunter_cacheable(OpenSSL)
-hunter_download(PACKAGE_NAME OpenSSL PACKAGE_INTERNAL_DEPS_ID 4)
+hunter_download(PACKAGE_NAME OpenSSL PACKAGE_INTERNAL_DEPS_ID 5)
--- a/cmake/projects/OpenSSL/schemes/url_sha1_openssl.cmake.in
+++ b/cmake/projects/OpenSSL/schemes/url_sha1_openssl.cmake.in
@@ -45,7 +45,6 @@ if(ANDROID)
     hunter_user_error("Could not find android architecture. Please set ANDROID_ARCH_NAME in your toolchain or use CMake 3.7+")
   endif()
   set(configure_command
-      MACHINE=${CMAKE_SYSTEM_PROCESSOR}
       # Ignored. Prevents script from checking host uname
       RELEASE=2.6.37
       SYSTEM=android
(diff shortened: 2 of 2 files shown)


ruslo:
> I'm unclear on the architecture of the PACKAGE_INTERNAL_DEPS_ID

I've added some details here:
* https://docs.hunter.sh/en/latest/reference/user-modules/hunter_download.html

> The commit 44c681e seems to have broken cross compiling builds

So what platforms do we fix by this change?

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1165
* https://travis-ci.org/ingenue/hunter/builds/175285201

edtanous:
This fixes cross compiling from Linux x86_64 to Armv6 32 bit.  My CMAKE_TOOLCHAIN_FILE looks like the following

``` cmake
SET(CMAKE_SYSTEM_NAME Linux)
SET(CMAKE_SYSTEM_VERSION 1)
set(CMAKE_SYSTEM_PROCESSOR armv6)

# specify the cross compiler
SET(CMAKE_C_COMPILER   arm-linux-gnueabi-gcc)
SET(CMAKE_CXX_COMPILER arm-linux-gnueabi-g++)

# where is the target environment 
SET(CMAKE_FIND_ROOT_PATH  /usr/arm-linux-gnueabi)

# search for programs in the build host directories
SET(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
# for libraries and headers in the target directories
SET(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
SET(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
```

This is on an Ubuntu 16.04 machine with a 5.4 gcc toolchain standard, although it's not clear if that matters.

ruslo:
> it's not clear if that matters

I will merge it because it doesn't break old behavior and seems to fix the issue for you. However I can't guarantee that there will be no problem with your toolchain in future. If you want to be sure you have to extend testing.

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1171
* https://travis-ci.org/ingenue/hunter/builds/175429374

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.16.27

PR #584, opened 2016-11-15
Add opencl headers as package

caseymcc:
If you open up a hunter_packages/OpenCL_Headers repo, I will put in a PR and update this PR to the appropriate release. Planing to also add another package "opencl" that includes the ICD (library) compile which is the reason I called this one opencl_headers.


ruslo:
Packages `OpenCL_Headers` and `opencl` have common files? Why there are two packages?

caseymcc:
No common files, they are in 2 separate repos, the headers can be used without the lib as long as you link with the lib on your platform. The OpenCL package will include the ICD which is a stand in for the native platform lib that will find the native one at runtime. It depends on the headers so I am setting up the package to use hunter to fetch the OpenCL_Header package.

ruslo:
> the headers can be used without the lib as long as you link with the lib on your platform

You mean system library that is not controlled by Hunter?

caseymcc:
yes or you can include the next package (OpenCL) I am going to add that includes the ICD which will be a hunter controlled library that will also use the OpenCL_Header package as a dependency.

ruslo:
> yes

In this case it conflicts with the Hunter approach, see this page:
* https://docs.hunter.sh/en/latest/faq/interaction-with-other-package-managers.html

> you can include the next package (OpenCL) I am going to add that includes the ICD which will be a hunter controlled library that will also use the OpenCL_Header package as a dependency

There should be introduced new feature that allow to have different packages to implement same "interface" (like different implementation of MPI), but in all such approaches the headers will not be used separately from the different implementation of sources. Note that headers and API are not the same.

API, library should provide method `boost A::a()` to user:

```cpp
struct A {
  bool a();
};
```

Real header for implementation-1:

```cpp
struct A {
  bool a() { return true; }
};
```

Real header for implementation-2:

```cpp
struct A {
  bool a() { return false; }
};
```

Issue #546, opened 2016-10-12
Add 'rapidxml' package


roalz:
http://rapidxml.sourceforge.net/index.htm
A clone on GitHub: https://github.com/dwd/rapidxml

Note: headers-only, few header files.

(My personal doubt: does it even make sense to have it as Hunter package?)
##

ruslo:
> My personal doubt: does it even make sense to have it as Hunter package?

Why not?

roalz:
Another fork on github, this one with tags for official versions and added CMake build, that may be easier to include in Hunter.
https://github.com/timniederhausen/rapidxml

ruslo:
Fixed by https://github.com/ruslo/hunter/pull/589

PR #589, opened 2016-11-17
Added RapidXML package

ErniBrown:
I created the RapidXML package, as proposed by @roalz in #546

### DIFF
--- a/cmake/projects/RapidXML/hunter.cmake
+++ b/cmake/projects/RapidXML/hunter.cmake
@@ -1,0 +1,21 @@
+# !!! DO NOT PLACE HEADER GUARDS HERE !!!
+
+include(hunter_add_version)
+include(hunter_cacheable)
+include(hunter_download)
+include(hunter_pick_scheme)
+
+hunter_add_version(
+    PACKAGE_NAME
+    RapidXML
+    VERSION
+    "1.13"
+    URL
+    "https://github.com/hunter-packages/rapidxml/archive/v1.13-hunter.tar.gz"
+    SHA1
+    d7f9b263a4c1e90dca488d3cf040364339ff09e2
+)
+
+hunter_pick_scheme(DEFAULT url_sha1_cmake)
+hunter_cacheable(RapidXML)
+hunter_download(PACKAGE_NAME RapidXML)
--- a/examples/RapidXML/CMakeLists.txt
+++ b/examples/RapidXML/CMakeLists.txt
@@ -1,0 +1,20 @@
+
+# Copyright (c) 2016, Ruslan Baratov
+# All rights reserved.
+
+cmake_minimum_required(VERSION 3.0)
+
+# Emulate HunterGate:
+# * https://github.com/hunter-packages/gate
+include("${CMAKE_CURRENT_SOURCE_DIR}/../common.cmake")
+
+project(download-RapidXML)
+
+# download RapidXML
+hunter_add_package(RapidXML)
+
+# RapidXML available
+find_package(RapidXML REQUIRED CONFIG)
+
+add_executable(foo foo.cpp)
+target_link_libraries(foo RapidXML::RapidXML)
--- a/cmake/configs/default.cmake
+++ b/cmake/configs/default.cmake
@@ -78,6 +78,7 @@ hunter_config(QtAndroidCMake VERSION 1.0.6)
 hunter_config(QtCMakeExtra VERSION 1.0.20)
 hunter_config(QtQmlManager VERSION 1.0.0)
 hunter_config(RapidJSON VERSION 1.0.2-p2)
+hunter_config(RapidXML VERSION 1.13)
 hunter_config(SDL2 VERSION 2.0.4-p2)
 hunter_config(SQLite3 VERSION autoconf-3080803) #R-Tree enabled
 hunter_config(Sober VERSION 0.1.3)
--- a/examples/RapidXML/foo.cpp
+++ b/examples/RapidXML/foo.cpp
@@ -1,0 +1,7 @@
+#include "rapidxml/rapidxml.hpp"
+
+int main()
+{
+	return 0;
+}
+


ruslo:
This library is header-only as far as I understand. This will be a trivial task to add CMake configs for it. It's much better than introducing new scheme and keeping the package unrelocatable. Few forks mentioned in https://github.com/ruslo/hunter/issues/546 @ErniBrown have you looked at them?

ErniBrown:
I'm not sure I understand, should we use a cmake project with an integrated config?
In #546 a cmake project is proposed, but no config is available with it.
What's your suggestion about this package?
Thanks

ruslo:
> I'm not sure I understand, should we use a cmake project with an integrated config?

We should patch the project and add CMake build. Since it's a header-only library that will be pretty simple step. Here is the boilerplate code to copy:
* https://github.com/ruslo/leathers/blob/1cf363e8abb440973cb400f421da8a3706a6b030/CMakeLists.txt#L91-L130

> In #546 a cmake project is proposed, but no config is available with it

I haven't look at those projects, it may be so.

> What's your suggestion about this package?

As far as I understand there are a few forks on GitHub already. Please investigate what is the best for us and I will add new fork for the hunter-packages. If on your opinion they are not good for any reason I can create new repository `hunter-packages/rapidxml`.

ErniBrown:
Ok, my proposal is [timniederhausen/rapidxml](https://github.com/timniederhausen/rapidxml). It has a little bit more feature than required, test, documentation, but is one of the simplest one with cmake support.
I will add a config finder.
Can you please add a hunter-packages fork?

ruslo:
> Ok, my proposal is timniederhausen/rapidxml

Thanks, created: https://github.com/hunter-packages/rapidxml

I've added you to the team, you will be able to push commits after your accept an invitation (check your email). However it make sense to create pull requests for the review reasons. Also please read https://docs.hunter.sh/en/latest/creating-new/cmake-dependencies.html#maintenance

ErniBrown:
I switched to [hunter-packages/rapidxml](https://github.com/hunter-packages/rapidxml) and added a very simple RapidXMLConfig.cmake.
There's an issue: both release and debug configuration are "built", which is useless since it's a header only library: I was not able to correctly set **HUNTER_PACKAGE_CONFIGURATION_TYPES**, where should i change it?

ruslo:
> added a very simple RapidXMLConfig.cmake

It doesn't add an imported target. Just copy/paste the code I've mentioned earlier:
* https://github.com/ruslo/leathers/blob/1cf363e8abb440973cb400f421da8a3706a6b030/CMakeLists.txt#L91-L130

> both release and debug configuration are "built"

Since it's header only library I guess the performance is not an issue. We are just copying set of files twice.

> which is useless since it's a header only library

In this case probably yes, however in general even library is header only we may want to add debug macro for example, like RAPIDXML_DEBUG.

ErniBrown:
> It doesn't add an imported target. Just copy/paste the code I've mentioned earlier:

You are right, i fixed it.

cmake/find/FindRapidXML.cmake
```diff
@@ -0,0 +1,41 @@
+# ---------------------------------------------------------------------------
```
ruslo: We don't need this file if RapidXMLConfig.cmake installed

cmake/projects/RapidXML/hunter.cmake
```diff
@@ -0,0 +1,19 @@
...
+    VERSION
+    "1.13"
+    URL
+    "https://github.com/hunter-packages/rapidxml/archive/v1.13-hunter.tar.gz"
+    SHA1
+    95164691ae21bd64b4dc0608a4e6d2d27c487dea
+)
+
+hunter_pick_scheme(DEFAULT url_sha1_cmake)
+hunter_download(PACKAGE_NAME RapidXML)
```
ruslo: http://0.readthedocs.io/en/latest/git.html#check-there-is-a-new-line-at-end-of-file

cmake/projects/RapidXML/hunter.cmake
```diff
@@ -0,0 +1,19 @@
...
+    RapidXML
+    VERSION
+    "1.13"
+    URL
+    "https://github.com/hunter-packages/rapidxml/archive/v1.13-hunter.tar.gz"
+    SHA1
+    95164691ae21bd64b4dc0608a4e6d2d27c487dea
+)
+
+hunter_pick_scheme(DEFAULT url_sha1_cmake)
```
ruslo: You should add `hunter_cacheable`, see https://docs.hunter.sh/en/latest/reference/user-modules/hunter_cacheable.html

examples/RapidXML/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+
+project(download-RapidXML)
+
+# download RapidXML
+hunter_add_package(RapidXML)
+
+# RapidXML available
+find_package(RapidXML REQUIRED CONFIG)
+
+add_executable(foo foo.cpp)
```
ruslo: We should link `RapidXML::RapidXML` to the `foo` target.

ErniBrown:
Everything fixed. Thanks for your patience, indeed I need a newline check script, i always forgot one somewhere

examples/RapidXML/CMakeLists.txt
ruslo: The name of the target was my guess. I think the real name is `RapidXML::rapidxml`. To avoid such errors you can test changes with [jenkins.py](https://github.com/ruslo/hunter/wiki/dev.hunter.root#jenkinspy) script.

ErniBrown:
I decided to maintain RapidXML::RapidXML as target name, and fix it in hunter-packages/rapidxml, since there was a previous error in it.

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1188
* https://travis-ci.org/ingenue/hunter/builds/178850127

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1192
* https://travis-ci.org/ingenue/hunter/builds/178854011

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.16.32

@ErniBrown feel free to create new wiki page

PR #591, opened 2016-11-17
[glog] remove entry with dead link

NeroBurner:


ruslo:
@NeroBurner My bad, thanks for fixing it.

PR #593, opened 2016-11-17
Update cmake-dependencies: fix typos, clarify

NeroBurner:
I don't understand what you want to say with the following paragraph

```
Note that the SHA1 of the Hunter archive in ``HunterGate`` commands of ``Bar`` and
 ``Roo`` can be different. Same location will be set automatically internally
 you don't have to manage it manually.
```

Do you mean, that the huntergate of the top project will be used for the submodule?

### DIFF
--- a/docs/creating-new/cmake-dependencies.rst
+++ b/docs/creating-new/cmake-dependencies.rst
@@ -4,9 +4,9 @@
 CMake (with dependencies)
 -------------------------
 
-If your project use external packages (i.e. has command ``find_package(Foo)``)
-you need to patch it first so this packages can be found in Hunter root
-directory instead of standard one:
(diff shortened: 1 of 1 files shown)


ruslo:
> Do you mean, that the huntergate of the top project will be used for the submodule?

Kind of. But why submodule? There are few scenarios when HunterGate called several times: project dependency in Hunter, dependencies of dependency, subproject added by `add_subdirectory`. In all cases only one HunterGate will be used, the parent one. User don't have to manage them manually, i.e. user don't have to use same URL/SHA1 in all involved HunterGates.

NeroBurner:
thanks for the clarification. I've tried to condense it into a paragraph. See #597

PR #594, opened 2016-11-17
OpenSSL :  fix windows compilation error NASM not found

Ubiquite:
I guess the continus integration computer has NASM, but not every computer : I fixed it like with 1.0 OpenSSL scheme


ruslo:
As I mentined in gitter the 1.1.0 version doesn't work for me on Windows. Take a look at https://github.com/ruslo/hunter/tree/openssl.update branch. `no-asm` is already there but it doesn't help as far as I remember.

PR #597, opened 2016-11-18
Update cmake-dependencies: clarify subproject SHA1

NeroBurner:


docs/creating-new/cmake-dependencies.rst
```diff
@@ -70,8 +70,11 @@ In terms of CMake code it means adding ``HunterGate`` and ``hunter_add_package``
   :align: center
 
 Note that the SHA1 of the Hunter archive in ``HunterGate`` commands of ``Bar`` and
-``Roo`` can be different. Same location will be set automatically internally
-you don't have to manage it manually.
+``Roo`` can be different. The main project may contain subprojects also using the
+``HunterGate`` command. The subprojects may be introduced as hunter dependencies,
+git submodules, or just the whole project added as subfolder.
```
ruslo: > The subprojects may be introduced as hunter dependencies

If we are calling "subprojects" the source subtree that have `project` command and added by `add_subdirectory` then Hunter dependencies are not subprojects.

> git submodules

This is a little bit confusing. CMake source tree and VCS are orthogonal. You can have subprojects without git submodules and submodules which is not used as a subprojects.

docs/creating-new/cmake-dependencies.rst
```diff
@@ -70,8 +70,11 @@ In terms of CMake code it means adding ``HunterGate`` and ``hunter_add_package``
...
+For all these submodules the URL and SHA1 of the main project are used. The user
```
ruslo: > submodules

I think it's confusing.

docs/creating-new/cmake-dependencies.rst
NeroBurner: you are right. Hunter uses something like SuperBuilds (done right :) )

Just for better understanding. Hunter dependencies are downloaded, extracted an build (ignoring the existence of cache servers). Only in the case of hunter dependencies, which are cmake-projects it is possible that the project has a `HunterGate` command.

The second case is a subproject added by `add_subdirectory` and this subproject uses the `HunterGate` command.

Do you know how we could name both cases with a single identifier?

And VCS (thus the mention of git submodules) should not be mentioned

docs/creating-new/cmake-dependencies.rst
NeroBurner: I'd say to replace `submodules` with `subprojects` or any other more fitting name we can come up with

NeroBurner:
@ruslo Had another go, hope it is more clearer now

docs/creating-new/cmake-dependencies.rst
```diff
@@ -69,9 +69,11 @@ In terms of CMake code it means adding ``HunterGate`` and ``hunter_add_package``
 .. image:: /images/package-conflict-resolved.png
   :align: center
 
-Note that the SHA1 of the Hunter archive in ``HunterGate`` commands of ``Bar`` and
-``Roo`` can be different. Same location will be set automatically internally
-you don't have to manage it manually.
+Note that now the main project ``Bar`` and the hunter depencency ``Roo`` contain
```
ruslo: Spell check [failed](https://travis-ci.org/ruslo/hunter/jobs/176962588):
* `depencency` -> `dependency`
* `subprojects` -> `sub-projects`

NeroBurner: Fixed typos

docs/creating-new/cmake-dependencies.rst
```diff
@@ -69,9 +69,11 @@ In terms of CMake code it means adding ``HunterGate`` and ``hunter_add_package``
...
   :align: center
 
-Note that the SHA1 of the Hunter archive in ``HunterGate`` commands of ``Bar`` and
-``Roo`` can be different. Same location will be set automatically internally
-you don't have to manage it manually.
+Note that now the main project ``Bar`` and the hunter dependency ``Roo`` contain
+a ``HunterGate`` command. The URL and SHA1 of the ``HunterGate`` command my not match.
+In this case the URL and SHA1 of the main project ``Bar`` are used for both 
+``HunterGate`` commands. The user doesn't need to manage them manually.
+The same is true for sub-projects added by `add_subdirectory` calls.
```
ruslo: I guess we need
```
``add_subdirectory``
```
instead of

```
`add_subdirectory`
```

ruslo: > Shall I squash the commits for this PR?

As you wish.

PR #598, opened 2016-11-24
yaml-cpp hunter package is created

tatraian:
I created the yaml-cpp package. Please check it.


cmake/projects/YAML_CPP/hunter.cmake
```diff
@@ -0,0 +1,21 @@
+# !!! DO NOT PLACE HEADER GUARDS HERE !!!
+
+include(hunter_add_version)
+include(hunter_cacheable)
+include(hunter_download)
+include(hunter_pick_scheme)
+
+hunter_add_version(
+  PACKAGE_NAME
+  YAML_CPP
```
ruslo: The C++ namespace is `YAML` so I guess we can name it just `YAML` (or even `yaml`).

tatraian: Ok, but in this case one should use package `yaml` for hunter package addition and find `YAML_CPP` by `find_package`. I think this is missleading. Anyway the project calls itself as `yaml-cpp` refering to the "official" `c++` implementation of yaml standard. So I think it would be better to rename the project to `yaml-cpp` and rewrite the package project name also to this.

ruslo: > I think it would be better to rename the project to yaml-cpp and rewrite the package project name also to this

Will you do it? Should I pause merge? I'm okay with merging it as is, though I think `YAML_CPP` is a strange name :)

cmake/configs/default.cmake
```diff
@@ -184,6 +184,7 @@ hunter_config(xrender VERSION 0.9.7)
 hunter_config(xshmfence VERSION 1.2)
 hunter_config(xtrans VERSION 1.2.7)
 hunter_config(xxf86vm VERSION 1.1.2)
+hunter_config(YAML_CPP VERSION 0.5.3)
```
pretyman: if the package is to be named with uppercase, can this be moved with the upper case packages above the lowercase?

ruslo: You mean the order in file? Yes it should be between `WinSparkle` and `ZLIB`.

cmake/projects/YAML_CPP/hunter.cmake
tatraian: Yes. I replace it in the hunter-package. Can I do it directly or you want to review it?

And I think `yaml-cpp` will be at good plae in `default.cmake`

Anyway, do you have "standard" for package names?

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1190
* https://travis-ci.org/ingenue/hunter/builds/178851359

ruslo:
Something failed. I guess the nmake, ninja problems on Windows is about Boost. On Travis there is a problem with `analyze` toolchain. @tatraian feel free to fix/exclude toolchains, branch is https://github.com/ingenue/hunter/tree/pkg.yaml-cpp

ruslo:
New testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1202
* https://travis-ci.org/ingenue/hunter/builds/179110867

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1203
* https://travis-ci.org/ingenue/hunter/builds/179121035

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.16.34

PR #601, opened 2016-11-26
Add gflags 2.2.0 and make it new default

schuhschuh:
Tested it with [gflags/example](https://github.com/gflags/example) project.


ruslo:
@schuhschuh Please update the name of the target in example. Now it's `gflags` instead of `gflags-static`: https://github.com/ruslo/hunter/blob/8cb68e7d470bb19d6bc1dbfe14391899c147b965/examples/gflags/CMakeLists.txt#L19

Also there are a few broken toolchains. For now they are excluded from testing, I can include them back if you're interested in fixing issues:
* `analyze` https://travis-ci.org/ingenue/hunter/jobs/164668384 (seems like a valid issue, `program_name` variable is not used indeed)
* `android` https://travis-ci.org/ingenue/hunter/jobs/164668381 (not sure what is happenning here, also does the Android supported platform? Probably it doesn't make sense to have gflags on Android?)

schuhschuh:
Sure, will update the example use of the Hunter package.

The unused `program_name` error is interesting. Will fix it in gflags.

Regarding Android build failure, I noticed that Hunter downloads a previous gflags version:
```
-- downloading...
       src='https://github.com/hunter-packages/gflags/archive/v2.1.2-p0.tar.gz'
       dst='/home/travis/build/ingenue/hunter/_testing/Hunter/_Base/Download/gflags/2.1.2-p0/113255c/v2.1.2-p0.tar.gz'
       timeout='none'
```

ruslo:
> Regarding Android build failure, I noticed that Hunter downloads a previous gflags version

The link to logs is about previous errors, however I've tested new version locally and `analyze`/`android` issues still present.

schuhschuh:
Ok, well it doesn't seem to be an error coming from gflags. What seems to happen is that some linking CMake script does not dereference the compiler variable:

```
CMAKE_C_COMPILER -fexceptions -frtti -Wno-psabi
  --sysroot=/home/travis/build/ingenue/hunter/_ci/android-ndk-r10e/platforms/android-19/arch-arm
  -funwind-tables -finline-limit=64 -fsigned-char -no-canonical-prefixes
  -march=armv7-a -mfloat-abi=softfp -mfpu=neon -fdata-sections
  -ffunction-sections -Wa,--noexecstack -std=c++11 -Wl,--fix-cortex-a8
  -Wl,--no-undefined -Wl,--gc-sections -Wl,-z,noexecstack -Wl,-z,relro
  -Wl,-z,now -Wl,-z,nocopyreloc -fPIE -pie
  CMakeFiles/cmTC_d22d2.dir/testCXXCompiler.cxx.o -o cmTC_d22d2
  "/home/travis/build/ingenue/hunter/_ci/android-ndk-r10e/sources/cxx-stl/gnu-libstdc++/4.9/libs/armeabi-v7a/libgnustl_static.a"
  "/home/travis/build/ingenue/hunter/_ci/android-ndk-r10e/sources/cxx-stl/gnu-libstdc++/4.9/libs/armeabi-v7a/libsupc++.a"
  -lm
  Error running link command: No such file or directory
```

The `No such file or directory` is probably `CMAKE_C_COMPILER`. Which should be the actual compiler binary path...

ruslo:
I've reverted back all excluded toolchains, testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1199
* https://travis-ci.org/ingenue/hunter/builds/179102816

ruslo:
> Ok, well it doesn't seem to be an error coming from gflags

The reason is about `project(... CXX)` but the error is in Android toolchain somewhere. I've tested new Android toolchain that came with CMake 3.7+ and it seems to work.

schuhschuh:
So would the error go away if `C` would be included as language for the Android toolchain? Could this be added as an exception for CMake <3.7 and Android toolchain only?

Similar to older CMake versions with optional `enable_language` if needed: https://github.com/gflags/gflags/blob/22ed2ce6c8e38cd390517ddf96276d8cc199df12/CMakeLists.txt#L94

ruslo:
```cmake
project (${PACKAGE_NAME} CXX)
if (CMAKE_VERSION VERSION_LESS 100)
  # C language still needed because the following required CMake modules
  # (or their dependencies, respectively) are not correctly handling
  # the case where only CXX is enabled.
  # - CheckTypeSize.cmake (fixed in CMake 3.1, cf. http://www.cmake.org/Bug/view.php?id=14056)
  # - FindThreads.cmake (--> CheckIncludeFiles.cmake <--)
  enable_language (C)
endif ()
```
I think this is equivalent to `project(${PROJECT_NAME} CXX C)` which is equivalent to just `project(${PROJECT_NAME})`?

schuhschuh:
Yes, but only when CMake is old (2.8) and the modules shipped with it would fail if the `C` language is not enabled. But gflags is a C++ only library. With CMake 3, you don't need to enable the C language. To fix the Android issue, I was just wondering to enable it in the if block also when gflags is configured for the Android toolchain using CMake <= 3.6. Then in future when CMake 3.7 is the minimum required version, this if block can be removed.

ruslo:
> So would the error go away if C would be included as language for the Android toolchain?

It seems so.

> Could this be added as an exception for CMake <3.7 and Android toolchain only?

On `gflags` side? I guess so.

> Similar to older CMake versions with optional enable_language if needed

Same part just quoted :) You have C language enabled anyway.

schuhschuh:
> Same part just quoted :) You have C language enabled anyway.

No. Only `if (CMAKE_VERSION VERSION_LESS 100)` ;-)

ruslo:
> No. Only if (CMAKE_VERSION VERSION_LESS 100) ;-)

Nope, `CMAKE_VERSION VERSION_LESS 100` means for every currently available releases. This branch will not be used only for CMake 100.0.0 version.

Try this one with CMake 3.7:
```cmake
if(CMAKE_VERSION VERSION_LESS 100)
  message("CMAKE_VERSION less 100: ${CMAKE_VERSION}")
endif()
```
Result:
```
CMAKE_VERSION less 100: 3.7.20161114
```

schuhschuh:
Ok, so I can add the condition `if (ANDROID)` to enable the `C` language.

Oh, I just wonder why `CMAKE_VERSION LESS 100`.... yep, got it :-S.

ruslo:
> Ok, so I can add the condition if (ANDROID) to enable the C language

This is a tricky part. ANDROID will be set in [this toolchain](https://github.com/taka-no-me/android-cmake) hence only **after** `project(...)` command called. And since issue is inside `project(...)` this will not help :) So the workaround is enable all languages always.

schuhschuh:
So the following will cover the Android issue, right?
`project(... CXX)` is save for Android with CMake 3.7?

```cmake
project (${PACKAGE_NAME} CXX)
if (CMAKE_VERSION VERSION_LESS 3.4 OR (ANDROID AND CMAKE_VERSION_LESS 3.7))
  # C language still needed because the following required CMake modules
  # (or their dependencies, respectively) are not correctly handling
  # the case where only CXX is enabled.
  # - CheckTypeSize.cmake (fixed in CMake 3.1, cf. http://www.cmake.org/Bug/view.php?id=14056)
  # - FindThreads.cmake   (fixed in CMake 3.4, cf. https://cmake.org/Bug/print_bug_page.php?bug_id=14905)
  enable_language (C)
endif ()
```

> And since issue is inside project(...) this will not help :) So the workaround is enable all languages always.

Can you try if calling `enable_language(C)` after `project` when `ANDROID` is already set is sufficient?

ruslo:
> So the following will cover the Android issue, right?

Nope. See above.

> Can you try if calling enable_language(C) after project when ANDROID is already set is sufficient?

Nope, doesn't work.

Also toolchain and CMake version is orthogonal here. CMake 3.7 do support new Android features but user may not use them in fact, i.e. they can create old kind of toolchain with CMake 3.7+.

ruslo:
Anyway this is an issue in toolchain, I don't think we should bother.

schuhschuh:
Alright, so this seems to have been fixed then anyway at some point in the toolchain.

I'll stick to enabling only CXX in gflags. No need for the compiler checks and tests for C when not used. (https://github.com/gflags/gflags/pull/196)

schuhschuh:
So given that I tagged gflags 2.2.0 yesterday and these are already 3 minor fixes I'll have to tag 2.2.1 before updating Hunter. Probably makes sense to wait for a couple days and see if more fixes come up. Then I'll modify this PR to use 2.2.1 instead.

ruslo:
Ok, so for now 2 questions left:
* `analyze` error confirmed on new version: https://travis-ci.org/ingenue/hunter/jobs/179102821
* `vs-9-2008` error https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1199/job/ah5vowfm7e1fxvqu (the Visual Studio version seems pretty old so it may be the compiler issue again, however it would be nice to have the minimum required version in official documenation)

ruslo:
> Then I'll modify this PR to use 2.2.1 instead

Okay, thanks.

v1bri:
Ping, any update on this PR? I saw it was closed (due to inactivity) but there hasn't been a 2.2.1 gflags release. Just wondering if we can get this ball rolling again.

ruslo:
@v1bri submit new pull request with version you need.

v1bri:
Sounds good. Quick question, is there any documentation for kicking off the CI build tests (ex: travis, appveyor, ...)? Are they automatically kicked off when a pull request is submitted to https://github.com/ingenue/hunter?

ruslo:
> Are they automatically kicked off when a pull request is submitted to https://github.com/ingenue/hunter?

You should submit pull request to `ruslo/hunter`, I will create temporary branch in `ingenue/hunter` to test package. You need to open pull request to `ingenue/hunter` only if you need to update `.travis.yml`/`appveyor.yml` of particular package.

May be this comment can be useful:
* https://github.com/ruslo/hunter/pull/742#issuecomment-293812754

PR #606, opened 2016-12-06
Package for getopt on Windows

caseymcc:
Package for getopt on Windows, most POSIX systems (Linux, Mac, etc...) already have it but Microsoft is still the hold out. Provides a package that works just like getopt on the other systems.


cmake/projects/getopt/hunter.cmake
```diff
@@ -0,0 +1,24 @@
...
+include(hunter_download)
+include(hunter_pick_scheme)
+
+hunter_add_version(
+    PACKAGE_NAME
+    getopt
+    VERSION
+    "1.0.0-p0"
+    URL
+    "https://github.com/caseymcc/getopt/archive/1.0.0-p0.tar.gz"
```
ruslo: You can use hunter-packages now:
* https://github.com/hunter-packages/wingetopt

cmake/projects/getopt/hunter.cmake
```diff
@@ -0,0 +1,24 @@
+# !!! DO NOT PLACE HEADER GUARDS HERE !!!
+if(MSVC) #only for Visual Studios, all other posix compiliers should have getopt
```
ruslo: I guess it's better to do:
```cmake
if(NOT MSVC)
  hunter_user_error("Only for MSVC")
endif()
```

Usually such checks performed right before `hunter_download`

cmake/projects/getopt/hunter.cmake
```diff
@@ -0,0 +1,24 @@
...
+    "https://github.com/caseymcc/getopt/archive/1.0.0-p0.tar.gz"
+    SHA1
+    4e9a11b5dad6ff849fbb775556df02251198c249
+)
+
+hunter_pick_scheme(DEFAULT url_sha1_cmake)
+hunter_cacheable(getopt)
+hunter_download(PACKAGE_NAME getopt)
+
+endif()
```
ruslo: http://0.readthedocs.io/en/latest/git.html#check-there-is-a-new-line-at-end-of-file

cmake/configs/default.cmake
```diff
@@ -51,6 +51,9 @@ hunter_config(CppNetlibUri VERSION 1.0.4-hunter)
 hunter_config(CsvParserCPlusPlus VERSION 1.0.1)
 hunter_config(Eigen VERSION 3.2.4-p0)
 hunter_config(Expat VERSION 2.1.1)
+if(MSVC)
+  hunter_config(getopt VERSION 1.0.0-p0)
```
ruslo: Should it be called `wingetopt`?

examples/getopt/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+# Emulate HunterGate:
+# * https://github.com/hunter-packages/gate
+include("../common.cmake")
+
+project(download-getopt)
+
+hunter_add_package(getopt)
+find_package(getopt CONFIG REQUIRED)
+
+add_executable(getopt_example main.cpp)
```
ruslo: I guess your example is designed to be used both on Windows and other platforms (?) In this case it should be:
```cmake
add_executable(getopt_example main.cpp)
if(MSVC)
  hunter_add_package(getopt)
  find_package(getopt CONFIG REQUIRED)
  target_link_libraries(getopt_example getopt::getopt)
endif()
```

cmake/configs/default.cmake
caseymcc: I dont know, I went back and forth. First coded it under wingetopt then changed it to getopt as there is nothing about it that makes it windows only. You could conceivably conpile it for linux or mac but there is no need.

examples/getopt/CMakeLists.txt
caseymcc: Had planned to just remove all the others from the test, but handling it this way should verify that getopt exist elsewhere.

cmake/configs/default.cmake
ruslo: > as there is nothing about it that makes it windows only

Well, except that this package works only on Windows, right? :)

> And the header are the same getopt.h

Okay, this is strong argument. If you want you can keep the name of the package, up to you.

caseymcc: I would like to get getopt, should we change "https://github.com/hunter-packages/wingetopt" to "https://github.com/hunter-packages/getopt"?

ruslo: @caseymcc renamed

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1239
* https://travis-ci.org/ingenue/hunter/builds/185524768

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1241
* https://travis-ci.org/ingenue/hunter/builds/185550285

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.17.4

PR #610, opened 2016-12-08
Add package - librtmp

caseymcc:
This adds librtmp package.

On Windows it requires the getopt package which has been included here till it is included with hunter. I have tested on Windows and Linux but will have to see what the results are from the other toolchains.

I had to include several fixes for compiling on VS as it appears it was never ported to VC. I have posted on the rtmpdump mailing list if/how to make the updates, so far just crickets.


ruslo:
Please rebase this one

PR #617, opened 2016-12-15
Add package flann 1.9.1

NeroBurner:
Add FLANN package. The default configuration allows building without any other dependencies

https://github.com/mariusmuja/flann


cmake/projects/flann/hunter.cmake
```diff
@@ -0,0 +1,24 @@
...
+    "1.9.1"
+    URL
+    "https://github.com/mariusmuja/flann/archive/1.9.1.tar.gz"
+    SHA1
+    ca3aee5670297f1db2eff122679ed9e87a70d830
+)
+
+# Pick a download scheme
+hunter_pick_scheme(DEFAULT url_sha1_cmake)
+hunter_download(PACKAGE_NAME flann)
```
ruslo: Is this package [cacheable](https://docs.hunter.sh/en/latest/reference/user-modules/hunter_cacheable.html)?

NeroBurner: I have no idea. How can I check it?

ruslo: We need an example for the start.

ruslo: > but I don't know how it helps to solve the question of relocatability

You can run install prodecure and take a look at installed `*Config.cmake` files. To run install procedure we need to have an example.

NeroBurner: it seems flann does provide neither a `*Config.cmake` nor `Find*.cmake` with the install target. I need to look into how to generate `*Config.cmake` and provide a patch to upstream

ruslo: > `Find*.cmake` with the install target

Project should not install `Find*.cmake` file, see for details:
* http://stackoverflow.com/a/20857070/2288008

> provide a patch to upstream

I can create fork in `hunter-packages` if you want

Issue #519, opened 2016-09-13
Create Package from precompiled tar.gz and zip files


NeroBurner:
I want to create a package, but I only have precompiled tar.gz or zip files. Is it possible to add them to a hunter package and download the right version (os, compiler)?
There is a precompiled tar-ball for each os-architecture-compiler combination

currently I'm using a SuperBuild schema. 
To define CMAKE_ARCH if use the following [script](https://github.com/axr/solar-cmake/blob/master/TargetArch.cmake).

First I define the download link as follows:

```
set(url_base ${SuperBuild_DOWNLOAD_BASE_URL})

# foolib
list(APPEND projects foo)
set(foolib_version "2.0.802")
if (CMAKE_ARCH STREQUAL "x86_64")
    set(foo_arch "x86_64")
elseif(CMAKE_ARCH STREQUAL "i386")
    set(foo_arch "x86")
elseif(CMAKE_ARCH MATCHES "(arm)|(armv5)|(armv6)|(armv7)")
    set(foo_arch "arm")
else()
    message(FATAL_ERROR "foo unknown architecture: ${CMAKE_ARCH}")
endif()
if(WIN32)
    set(foo_os "windows")
elseif(UNIX AND NOT APPLE)
    set(foo_os "linux")
else()
    message(FATAL_ERROR "foo unknown OS (not windows and not linux")
endif()
set(foo_url_base "${url_base}/foolib")
set(foo_url "${foo_url_base}/${foolib_version}/fooblib-${foolib_version}-${foo_arch}-${foo_os}-factory.tar.gz")
```

Then I define an External Project with an URL (but without sha-checksum).

```
set(_install "${SuperBuild_INSTALL_PREFIX}")

ExternalProject_Add(foo
    URL ${foo_url}
    UPDATE_COMMAND ""
    CONFIGURE_COMMAND ""
    BUILD_COMMAND ""
    INSTALL_COMMAND ${CMAKE_COMMAND} -E make_directory ${_install}
        COMMAND ${CMAKE_COMMAND} -E copy_directory <SOURCE_DIR>/interface/c ${_install}/include
        COMMAND ${CMAKE_COMMAND} -E copy_directory <SOURCE_DIR>/interface/cpp ${_install}/include
        COMMAND ${CMAKE_COMMAND} -E copy_directory <SOURCE_DIR>/library ${_install}/lib
    LOG_DOWNLOAD 1
    LOG_INSTALL 1
)

[...]

PR #624, opened 2016-12-19
Add module to check for define in toolchain.info

NeroBurner:
as discussed in #519 a function to read `toolchain.info` file and check if a specific define is set

### DIFF
--- a/cmake/modules/hunter_check_toolchain_definition.cmake
+++ b/cmake/modules/hunter_check_toolchain_definition.cmake
@@ -1,0 +1,82 @@
+# Copyright (c) 2016 NeroBurner
+# All rights reserved.
+
+include(CMakeParseArguments) # cmake_parse_arguments
+
+include(hunter_test_string_not_empty)
+
+function(hunter_check_toolchain_definition)
+  set(function_name "hunter_check_toolchain_definition")
+  set(function_synopsis "${function_name}(NAME definition [DEFINED out_is_defined] [VALUE out_value])")
+
+  # parse arguments
+  set(one_value_args NAME DEFINED VALUE)
+  cmake_parse_arguments(hunter_ct_arg "" "${one_value_args}" "" ${ARGV})
+
+  string(COMPARE NOTEQUAL "${hunter_ct_arg_NAME}"    "" arg_NAME)
+  string(COMPARE NOTEQUAL "${hunter_ct_arg_DEFINED}" "" arg_DEFINED)
+  string(COMPARE NOTEQUAL "${hunter_ct_arg_VALUE}"   "" arg_VALUE)
+  # No free arguments allowed
+  list(LENGTH hunter_ct_arg_UNPARSED_ARGUMENTS hunter_ct_len)
+  if(NOT hunter_ct_len EQUAL 0)
+    hunter_internal_error(
+      "'${function_name}' incorrect usage,"
+      " expected no free arguments '${hunter_ct_arg_UNPARSED_ARGUMENTS}'."
+      " Synopsis: ${function_synopsis}"
+    )
+  endif()
+  # option NAME is mandatory
+  if(NOT arg_NAME)
+    hunter_internal_error(
+      "'${function_name}' incorrect usage,"
+      " option 'NAME' with one argument is mandatory."
+      " Synopsis: ${function_synopsis}"
+    )
+  endif()
+  string(FIND "${hunter_ct_arg_NAME}" " " arg_NAME_whitespace_position)
+  if(NOT arg_NAME_whitespace_position EQUAL -1)
+    hunter_internal_error(
+      "'${function_name}' incorrect usage,"
+      " definition '${hunter_ct_arg_NAME}' with whitespaces not allowed."
+    )
+  endif()
+  set(definition "${hunter_ct_arg_NAME}")
+
+  # the toolchain path must be set
+  hunter_test_string_not_empty("${HUNTER_TOOLCHAIN_ID_PATH}")
+  set(TOOLCHAIN_INFO_FILE "${HUNTER_TOOLCHAIN_ID_PATH}/toolchain.info")
+
+  # read toolchain.info file into memory, each line a entry in a list
+  file(STRINGS "${TOOLCHAIN_INFO_FILE}" toolchain_contents)
+
+  # clean ouptut variables
+  set(is_defined NO)
+  set(defined_value)
+
+  foreach(line ${toolchain_contents})
+    # find the specified define
+    string(REGEX MATCH "^#define ${definition} .*$" match "${line}")
+    if(match)
+      # check for double definition
+      if(is_defined)
+        hunter_internal_error(
+          "'${function_name}' incorrect toolchain.info file,"
+          " double definition of define '${definition}'"
+          " in file '${TOOLCHAIN_INFO_FILE}'."
+        )
+      endif()
+      # definition found
+      set(is_defined YES)
+      # extract the definition
+      string(REGEX REPLACE "^#define ${definition} " "" defined_value "${match}")
+    endif()
+  endforeach()
+
+  # set output_var to found definition
+  if(arg_DEFINED)
+    set(${hunter_ct_arg_DEFINED} ${is_defined} PARENT_SCOPE)
+  endif()
+  if(arg_VALUE)
+    set(${hunter_ct_arg_VALUE} ${defined_value} PARENT_SCOPE)
+  endif()
+endfunction()


cmake/modules/hunter_check_toolchain_for_define.cmake
```diff
@@ -0,0 +1,36 @@
...
+  # the toolchain path must be set
+  hunter_test_string_not_empty("${HUNTER_TOOLCHAIN_ID_PATH}")
+  set(TOOLCHAIN_INFO_FILE "${HUNTER_TOOLCHAIN_ID_PATH}/toolchain.info")
+
+  # read toolchain.info file into memory
+  file(READ "${TOOLCHAIN_INFO_FILE}" toolchain_contents)
+
+  # Convert file contents into a CMake list (where each element in the list
+  # is one line of the file)
+  string(REGEX REPLACE ";" "\\\\;" toolchain_contents "${toolchain_contents}")
```
ruslo: You can use `file(STRINGS)`:
* https://cmake.org/cmake/help/v3.0/command/file.html

cmake/modules/hunter_check_toolchain_for_define.cmake
```diff
@@ -0,0 +1,36 @@
...
+  string(REGEX REPLACE "\n" ";"    toolchain_contents "${toolchain_contents}")
+
+  set(definition_found 0)
+  foreach(line ${toolchain_contents})
+    # trim line
+    string(REGEX REPLACE "^#define " "" line "${line}")
+    string(REGEX REPLACE " 1$" "" line "${line}")
```
ruslo: It's not always `1`, it can be anything in general like `"6.0 (clang-600.0.54)"` or `__attribute__((__blocks__(byref)))`:
* https://github.com/ingenue/hunter-cache/blob/master/049fe35/toolchain.info

cmake/modules/hunter_check_toolchain_for_define.cmake
```diff
@@ -0,0 +1,36 @@
...
+
+  # Convert file contents into a CMake list (where each element in the list
+  # is one line of the file)
+  string(REGEX REPLACE ";" "\\\\;" toolchain_contents "${toolchain_contents}")
+  string(REGEX REPLACE "\n" ";"    toolchain_contents "${toolchain_contents}")
+
+  set(definition_found 0)
+  foreach(line ${toolchain_contents})
+    # trim line
+    string(REGEX REPLACE "^#define " "" line "${line}")
```
ruslo: With the note below I think we may have some false positives here for lines which don't start with `#define`. I think next regex should be check and analyzed:
```cmake
string(REGEX MATCH "^#define ${definition} .*$" match "${line}")
```

NeroBurner:
@ruslo incorporated your changes, please have another look

cmake/modules/hunter_check_toolchain_for_define.cmake
```diff
@@ -0,0 +1,29 @@
...
+
+  # clean ouptut variable
+  set(definition_found)
+
+  foreach(line ${toolchain_contents})
+    # find the specified define
+    string(REGEX MATCH "^#define ${definition} .*$" match "${line}")
+    if(match)
+      # extract the definition
+      string(REGEX REPLACE "^#define ${definition} " "" definition_found "${match}")
```
ruslo: The name of variable `definition_found` doesn't match the meaning. I guess the value of variable `definition_found` should be `YES` or `NO`.

NeroBurner: renamed the variable. maybe the hunter function should also get a different name. For example `hunter_load_toolchain_definition`?

ruslo: > renamed the variable

I didn't said that the name of variable is wrong :) So now if we have toolchain.info:
```
#define A
```
and what to check A (defined) and B (not defined), we will have the same results:
```cmake
hunter_check_toolchain_for_define("A" a_defined)
hunter_check_toolchain_for_define("B" b_defined)
if(a_defined STREQUAL "")
  message("A not defined")
endif()
if(b_defined STREQUAL "")
  message("B not defined")
endif()
```
I guess output will be:
```
A not defined
B not defined
```
which is wrong, or at least confusing.

There are two traits of macro in C++. One is about is the macro defined or not, and the second is about the value of macro. So if you're not sure you should write `#if defined(SOMETHING) && (SOMETHING > 3)`. Same is here, I was thinking that this function should work similar to `#if defined(SOMETHING)`, looks like you're trying to get the value.

ruslo: > For example hunter_load_toolchain_definition?

What about `s,load,check,`?

NeroBurner: I was halfway through with an answer about how nice it is to get the defined value, until I've reread your post and understood your point.
`#define A` IS defined, but with the current implementation it would return the same result as if `#define A` was not in `toolchain.info`.

If I just return a boolean (1/0), then what should `#define B 0` return? Because strictly speaking B is defined, but it is defined as false.

If I return the defined value, then empty return values are equal to `#define A`. I could replace the output with true (1) if the value would be empty

ruslo: > If I just return a boolean (1/0), then what should #define B 0 return?

Will the code inside `#if defined(A)` be evaluated?

```cpp
#define A 0
#if defined(A)
  // some code
#endif
```

As far as I know yes. Hence we should return YES.

Just like I said I was thinking about function that will just return YES if **defined** and NO if **not defined**, nothing about the real value. If you want value then you should introduce two output variables, like:
```cmake
hunter_check_toolchain_definition(NAME A DEFINED a_defined VALUE a_value)
```
in this case `#defined A` will return `a_defined=YES`/`a_value=""`, `#defined A 0` will return `a_defined=YES`/`a_value=0`, etc.

NeroBurner: implemented your suggestion

```cmake
hunter_check_toolchain_definition(NAME A DEFINED a_defined VALUE a_value)
```

cmake/modules/hunter_check_toolchain_definition.cmake
```diff
@@ -0,0 +1,68 @@
+# Copyright (c) 2016 NeroBurner
+# All rights reserved.
+
+include(CMakeParseArguments) # cmake_parse_arguments
+
+include(hunter_test_string_not_empty)
+
+function(hunter_check_toolchain_definition)
+  set(_function_name "hunter_check_toolchain_definition")
```
ruslo: There is no need to use `_*` for variables - we are inside function. Function has separate variable scope.

NeroBurner: force of habbit :)

cmake/modules/hunter_check_toolchain_definition.cmake
```diff
@@ -0,0 +1,68 @@
...
+  set(one_value_args NAME DEFINED VALUE)
+  cmake_parse_arguments(hunter_ct_arg "" "${one_value_args}" "" ${ARGV})
+
+  string(COMPARE NOTEQUAL "${hunter_ct_arg_NAME}"    "" arg_NAME)
+  string(COMPARE NOTEQUAL "${hunter_ct_arg_DEFINED}" "" arg_DEFINED)
+  string(COMPARE NOTEQUAL "${hunter_ct_arg_VALUE}"   "" arg_VALUE)
+  # No free arguments allowed
+  list(LENGTH hunter_ct_arg_UNPARSED_ARGUMENTS hunter_ct_len)
+  if(NOT ${hunter_ct_len} EQUAL 0)
+    hunter_internal_error(
```
ruslo: `include(hunter_internal_error)` needed

cmake/modules/hunter_check_toolchain_definition.cmake
```diff
@@ -0,0 +1,68 @@
...
+  endif()
+  # option NAME is mandatory
+  if(NOT arg_NAME)
+    hunter_internal_error(
+      "'${function_name}' incorrect usage,"
+      " option 'NAME' with one argument is mandatory."
+      " Synopsis: ${function_synopsis}"
+    )
+  endif()
+  set(definition "${hunter_ct_arg_NAME}")
```
ruslo: We need to check that there is no spaces in `definition`.

cmake/modules/hunter_check_toolchain_definition.cmake
```diff
@@ -0,0 +1,68 @@
...
+
+  foreach(line ${toolchain_contents})
+    # find the specified define
+    string(REGEX MATCH "^#define ${definition} .*$" match "${line}")
+    if(match)
+      # definition found
+      set(is_defined YES)
+      # extract the definition
+      string(REGEX REPLACE "^#define ${definition} " "" defined_value "${match}")
+      break()
```
ruslo: I have an idea for one more sanity check: we can keep moving in loop and check if both `match` and `is_defined` is `YES` then report iternal error (double match).

cmake/modules/hunter_check_toolchain_definition.cmake
NeroBurner: updated

cmake/modules/hunter_check_toolchain_definition.cmake
NeroBurner: added

cmake/modules/hunter_check_toolchain_definition.cmake
ruslo: ping

cmake/modules/hunter_check_toolchain_definition.cmake
```diff
@@ -0,0 +1,82 @@
...
+  # option NAME is mandatory
+  if(NOT arg_NAME)
+    hunter_internal_error(
+      "'${function_name}' incorrect usage,"
+      " option 'NAME' with one argument is mandatory."
+      " Synopsis: ${function_synopsis}"
+    )
+  endif()
+  string(FIND "${hunter_ct_arg_NAME}" " " arg_NAME_whitespace_position)
+  if(arg_NAME_WHITESPACE_POSITION EQUAL -1)
```
ruslo: `arg_NAME_whitespace_position` and `arg_NAME_WHITESPACE_POSITION` are different variables.

NeroBurner: fixed

cmake/modules/hunter_check_toolchain_definition.cmake
ruslo: !

cmake/modules/hunter_check_toolchain_definition.cmake
```diff
@@ -0,0 +1,82 @@
...
+      "'${function_name}' incorrect usage,"
+      " option 'NAME' with one argument is mandatory."
+      " Synopsis: ${function_synopsis}"
+    )
+  endif()
+  string(FIND "${hunter_ct_arg_NAME}" " " arg_NAME_whitespace_position)
+  if(NOT arg_NAME_whitespace_position EQUAL -1)
+    hunter_internal_error(
+      "'${function_name}' incorrect usage,"
+      " definition with whitespaces not allowed."
```
ruslo: You can add the definition itself here: `" definition '${definition}' with whitespaces not allowed."`.

NeroBurner: > !

sorry, what do you mean? Do you mean we should allow unparsed arguments, or is the following line wrong?
```cmake
if(NOT ${hunter_ct_len} EQUAL 0)
```

NeroBurner:
I've squashed the commits

ruslo:
Documentation updated:
* https://docs.hunter.sh/en/latest/reference/user-modules/hunter_check_toolchain_definition.html

feel free to add better example :)

Issue #623, opened 2016-12-19
Understanding the Eigen Package


NeroBurner:
I'm working on Ubuntu 16.04. When installing `libeigen3-dev` the file `/usr/lib/cmake/eigen3/Eigen3Config.cmake` is installed. Therefore the eigen package can be found with `find_package(Eigen3 REQUIRED)` and the include dir is saved in `EIGEN3_INCLUDE_DIRS`.

But with hunter one needs the lines
```
hunter_add_package(Eigen)
find_package(Eigen)
```
but neither `Eigen3_INCLUDE_DIRS` nor `Eigen_INCLUDE_DIRS` is set

On both Hunter and Ubuntu a file `share/pkgconfig/eigen3.pc` is installed. With the only difference of the variable `prefix` being set to either `/usr` or  the hunter build directory.

After some more digging I've found the file `${HUNTER_ROOT}/cmake/find/FindEigen.cmake`. There a target `Eigen::eigen` is defined

To support both Hunter and Ubuntu a CMakeLists file would need to look as follows:
```
hunter_add_package(Eigen)
add_executable(eigen_test main.cpp)
if (HUNTER_ENABLED)
	find_package(Eigen REQUIRED)
	target_link_libraries(eigen_test PRIVATE Eigen::eigen)
else()
	find_package(Eigen3 REQUIRED)
	message("target_include_directories(eigen_test PRIVATE ${EIGEN3_INCLUDE_DIRS})")
	target_include_directories(eigen_test PRIVATE ${EIGEN3_INCLUDE_DIRS})
endif()
```

I'd like to unify the the approach of finding and adding the Eigen header files. A few approaches I could think of:
- supply the Upstream Eigen package with a `FindEigen3.cmake` file, which is installed just like the `eigen3.pc` file
- Rename the Hunter package to `Eigen3` (`project/Eigen` and `find/FindEigen.cmake`) and provide the `EIGEN3_INCLUDE_DIRS` variable
- somehow use the already generated `eigen3.pc`-file as hints

What do you guys think?

---

caseymcc:
Take this all with a grain of salt as I haven't been using/working with hunter long (and defiantly not a final authority). I believe the idea behind hunter is for it to ignore (as much as possible) the os packages. Mainly for the reason of cross compiling, you don't want to get the local os package pulled in during the compile for some other os. Providing fallback or look ups on os packages would likely cause more issues than it solves.

[...]

PR #626, opened 2016-12-21
Eigen: update to 3.3.1, remove old versions

NeroBurner:
Updating Eigen to 3.3.1, this version provides a `Eigen3Config.cmake` file itself as discussed in #623

Shall we also rename the hunter package from `Eigen` to `Eigen3` such that the `hunter_add_package` and `find_package` names match?


cmake/projects/Eigen/hunter.cmake
```diff
@@ -17,26 +17,13 @@ hunter_add_version(
...
-    Eigen
-    VERSION
-    "3.2.4"
-    URL
-    "https://bitbucket.org/eigen/eigen/get/3.2.4.tar.gz"
-    SHA1
-    b826f665bdfe31784795eef79c75540db71ab702
-)
-
-hunter_cmake_args(Eigen CMAKE_ARGS EIGEN_ENABLE_TESTING=OFF)
```
ruslo: is this one removed on purpose?

ruslo:
> Shall we also rename the hunter package from Eigen to Eigen3 such that the hunter_add_package and find_package names match?

I need to know what files installed by this new version of Eigen. Have you tried to build an example? Seems that it will not work and have to be changed too.

cmake/projects/Eigen/hunter.cmake
NeroBurner: yes, the variable was introduced by you with https://github.com/hunter-packages/eigen/commit/930d0a2c836b4957b114554031c596ef219e7fb2
In Eigen 3.3.1 I was not able to find the variable `EIGEN_ENABLE_TESTING`

ruslo: > yes, the variable was introduced by you with hunter-packages/eigen@930d0a2

Heh :) Okay, I see.

> In Eigen 3.3.1 I was not able to find the variable EIGEN_ENABLE_TESTING

Also as you may noticed in comment, this variable fix the Android build https://github.com/hunter-packages/eigen/commit/930d0a2c836b4957b114554031c596ef219e7fb2

NeroBurner:
I setup a small test project downloading the NeroBurner Eigen 3.3.1 branch
https://github.com/NeroBurner/hunter_test_eigen_download

cmake/projects/Eigen/hunter.cmake
NeroBurner: > Also as you may noticed in comment, this variable fix the Android build hunter-packages/eigen@930d0a2

could you test what exactly breaks the Android build. I havn't built for Android. Maybe we can send a patch upstream to the Eigen developers which disables tests

ruslo:
> I setup a small test project downloading the NeroBurner Eigen 3.3.1 branch

I haven't tried it but according to C++ code the header is still `Eigen/*` and not `Eigen3/*`. And C++ namespace is `Eigen::*` and not `Eigen3::*`. Hence package should be called `Eigen` and config file should be called `EigenConfig.cmake`. So let's keep the name.

cmake/projects/Eigen/hunter.cmake
ruslo: Please update an example and I will run all tests, including Android.

examples/Eigen/CMakeLists.txt
```diff
@@ -11,10 +11,10 @@ project(download-eigen)
 
 hunter_add_package(Eigen)
 
-find_package(Eigen REQUIRED)
+find_package(Eigen3 REQUIRED)
```
ruslo: Let it be `CONFIG REQUIRED` just to be sure

examples/Eigen/CMakeLists.txt
```diff
@@ -11,10 +11,10 @@ project(download-eigen)
...
 
 # Test double library creation
-find_package(Eigen REQUIRED)
+find_package(Eigen3 REQUIRED)
```
ruslo: `CONFIG REQUIRED`

examples/Eigen/CMakeLists.txt
NeroBurner: updated

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1250
* https://travis-ci.org/ingenue/hunter/builds/185806578

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1251
* https://travis-ci.org/ingenue/hunter/builds/185833309

ruslo:
Released:
* https://github.com/ruslo/hunter/releases/tag/v0.17.5

Please update wiki page.

cmake/projects/Eigen/hunter.cmake
ruslo: > this variable fix the Android build

Here is the case https://github.com/ruslo/hunter/issues/628

PR #631, opened 2017-01-03
include(hunter_cacheable)

headupinclouds:
Make nlohmann-json cacheable


cmake/projects/nlohmann-json/hunter.cmake
```diff
@@ -1,6 +1,7 @@
 # !!! DO NOT PLACE HEADER GUARDS HERE !!!
 
 include(hunter_add_version)
+include(hunter_cacheable)
```
ruslo: We should actually call `hunter_cacheable(nlohmann-json)`, see:
* https://github.com/ruslo/hunter/pull/633/files

PR #635, opened 2017-01-04
Add package OpenCL-clhpp

caseymcc:
Adds c++ headers for opencl, uses python to generate cl.hpp, cl2.hpp is just copied. Produces a warning if python is not found, it continues the build as cl2.hpp can be built without python.


cmake/projects/OpenCL-clhpp/hunter.cmake
```diff
@@ -0,0 +1,33 @@
...
+include(hunter_pick_scheme)
+
+# List of versions here...
+hunter_add_version(
+    PACKAGE_NAME
+    OpenCL-clhpp
+    VERSION
+    "2.0.10-p0"
+    URL
+    "https://github.com/caseymcc/OpenCL-CLHPP/archive/2.0.10-p0.tar.gz"
```
ruslo: Please use this one https://github.com/hunter-packages/OpenCL-CLHPP

examples/OpenCL-clhpp/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+
+project(download_opencl_clhpp)
+
+hunter_add_package(OpenCL)
+find_package(OpenCL CONFIG REQUIRED)
+
+hunter_add_package(OpenCL-clhpp)
+find_package(OpenCL-clhpp CONFIG REQUIRED)
+
+include_directories("${OPENCL_CLHPP_INCLUDE_DIRS}")
```
ruslo: No full paths. Since this package is cacheable (relocatable) all include paths should be propagated through imported target (like OpenCL::OpenCL).

ruslo:
> Produces a warning if python is not found

Please end with FATAL_ERROR in this case. Configuration must be predictable and should not rely on outside environment:
* https://docs.hunter.sh/en/latest/creating-new/cmake.html#environment

caseymcc:
It requires the OpenCL package (which includes the lib, not system lib), I have modified the package to set OpenCL up as a dependency. 

As far as the FATAL_ERROR, I will modified it to have an option of "generating"  cl2.hpp only as this does not require python. If the option is false (meaning cl.hpp is to be generated as well) then I will have it FATAL_ERROR if python is not available.

ruslo:
> If the option is false

So in the current default configuration if Python not found we will exit with FATAL_ERROR, right?

Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1300
* https://travis-ci.org/ingenue/hunter/builds/189377480

caseymcc:
Yeah default will exit with FATAL_ERROR. Also I see there are some failures on the testing, will look into it tomorrow.

On a side note, the package name "OpenCL-clhpp" seems a bit cumbersome. Wondering if it should be changed to something like OpenCL++ (will hunter and cmake take that?), as it is the c++ wrappers for the c api.

ruslo:
> Wondering if it should be changed to something like OpenCL++

Up to you.

> will hunter and cmake take that?

Will work, see for example [Async++](https://github.com/ruslo/hunter/blob/master/examples/Async%2B%2B/CMakeLists.txt). However I prefer `xx` or `pp` instead of `++`.

caseymcc:
Ok, so another question for you. This package requires OpenCL which includes the ICD lib, The ICD does not work on Android and is remarked out in the test. However (haven't checked yet) but these headers should work fine on the Android toolchain (without the OpenCL package) assuming the NDK is available somewhere. Should I attempt to get that working or just turn the Android test off?

ruslo:
> Should I attempt to get that working or just turn the Android test off?

How about turning Android tests off now, and trying to fix this issue later?

caseymcc:
This one should be good to go, test updates at https://github.com/ingenue/hunter/pull/47. Some issues there though as since I changed the name from OpenCL-clhpp to OpenCL-cpp test package name (pkg.opencl-clhpp) is a little off.

Here are my test
https://travis-ci.org/caseymcc/hunter/builds/190275079
https://ci.appveyor.com/project/caseymcc/hunter/build/1.0.69

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1310
* https://travis-ci.org/ingenue/hunter/builds/191888885

ruslo:
Uploading
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1312
* https://travis-ci.org/ingenue/hunter/builds/191915625

ruslo:
Released:
* https://github.com/ruslo/hunter/releases/tag/v0.17.13

PR #638, opened 2017-01-09
Add link to Supported Packages wiki

caseymcc:


ruslo:
> https://github.com/ruslo/hunter/wiki/packages

This looks nice, thank you! However I prefer readthedocs, there is already "All packages" page created:
* https://docs.hunter.sh/en/latest/packages/all-packages.html

Feel free to convert wiki page to RST format and save in to https://github.com/ruslo/hunter/blob/master/docs/packages/all-packages.rst

PR #639, opened 2017-01-10
add zookeeper package

wheybags:
Hey, not sure how adding a package works, but here's a pull request anyway. For now the packages is hosted on a fork on our organisation, I suppose it would be better if it were on the hunter-packages org?

For reference, zookeeper is a distributed systems  message passing tool written in java, and this package is for the c client library.

Also I could add an example maybe?

### DIFF
--- a/cmake/configs/default.cmake
+++ b/cmake/configs/default.cmake
@@ -233,4 +233,4 @@ if(ANDROID)
   endif()
 endif()
 
-hunter_config(zookeeper VERSION 3.4.9-p1)
+hunter_config(zookeeper VERSION 3.4.9-p2)
--- a/cmake/configs/default.cmake
+++ b/cmake/configs/default.cmake
@@ -232,3 +232,5 @@ if(ANDROID)
     )
   endif()
 endif()
+
+hunter_config(zookeeper VERSION 3.4.9-p1)
--- a/cmake/projects/zookeeper/hunter.cmake
+++ b/cmake/projects/zookeeper/hunter.cmake
@@ -14,9 +14,9 @@ hunter_add_version(
     VERSION
     "3.4.9-p1"
     URL
-    "https://github.com/Artomatix/zookeeper/archive/3.4.9-hunter-4.tar.gz"
+    "https://github.com/Artomatix/zookeeper/archive/3.4.9-hunter-5.tar.gz"
     SHA1
-    4edd9af81461f70cf63f97a153f98449246cbb21
+    f98aba57b183de9e00a58536c7a47c612e3d51fc
 )
 
 # Pick a download scheme
--- a/cmake/projects/zookeeper/hunter.cmake
+++ b/cmake/projects/zookeeper/hunter.cmake
@@ -15,9 +15,9 @@ hunter_add_version(
     VERSION
     "3.4.9-p1"
     URL
-    "https://github.com/hunter-packages/zookeeper/archive/3.4.9-hunter-7.tar.gz"
+    "https://github.com/hunter-packages/zookeeper/archive/3.4.9-p1.tar.gz"
     SHA1
-    6d73772207e55d7b614a2aaa3f5ba5203c2df045
+    477ddb59d7071e83fc1acc9f8f0425f97ccfa56a
 )
 
 # Pick a download scheme
--- a/cmake/projects/zookeeper/hunter.cmake
+++ b/cmake/projects/zookeeper/hunter.cmake
@@ -15,9 +15,9 @@ hunter_add_version(
     VERSION
     "3.4.9-p1"
     URL
-    "https://github.com/hunter-packages/zookeeper/archive/3.4.9-hunter-6.tar.gz"
+    "https://github.com/hunter-packages/zookeeper/archive/3.4.9-hunter-7.tar.gz"
     SHA1
-    fe68462b58a6eff5688a81e9031fbe32f0acfe74
+    6d73772207e55d7b614a2aaa3f5ba5203c2df045
 )
 
 # Pick a download scheme
--- a/cmake/projects/zookeeper/hunter.cmake
+++ b/cmake/projects/zookeeper/hunter.cmake
@@ -6,6 +6,7 @@
 include(hunter_add_version)
 include(hunter_download)
 include(hunter_pick_scheme)
+include(hunter_cacheable)
 
 # List of versions here...
 hunter_add_version(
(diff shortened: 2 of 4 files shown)


wheybags:
the changes from upstream: https://github.com/Artomatix/zookeeper/compare/release-3.4.9...release-3.4.9-hunter

cmake/projects/zookeeper/hunter.cmake
```diff
@@ -0,0 +1,29 @@
...
+    f98aba57b183de9e00a58536c7a47c612e3d51fc
+)
+
+# Pick a download scheme
+hunter_pick_scheme(DEFAULT url_sha1_cmake) # use scheme for cmake projects
+
+# Download package.
+# Two versions of library will be build by default:
+#     * libexample_A.a
+#     * libexample_Ad.a
```
ruslo: Is this package relocatable? https://docs.hunter.sh/en/latest/creating-new/cmake.html#important

ruslo:
> For now the packages is hosted on a fork on our organisation, I suppose it would be better if it were on the hunter-packages org?

This is up to you. If you want it to be hosted under `hunter-packages` org I can create corresponding repo and give you permissions to write.

> Also I could add an example maybe?

This is a requirement if you want package to be tested/uploaded. You can add simple one, just to check headers are found and libraries linked (use some function from package).

wheybags:
I think it is relocatable yeah? Just means that it produces files for config mode right?
Also yeah, I'd like to have it under hunter-packages.

Actually, I have a question too, can you point a hunter package at a subdir in an archive? I ask because the c binding is actually in the src/c subfolder. For noe I've just added a CMakeLists.txt at the root which uses add_subdirectory to add that folder, but its not a wildly  elegant solution :p

ruslo:
> Just means that it produces files for config mode right?

Yes, and there should be no full paths inside those configs. Please add `hunter_cacheable` in this case (see the link above).

> I'd like to have it under hunter-packages

Created: https://github.com/hunter-packages/zookeeper
Please read: https://docs.hunter.sh/en/latest/creating-new/cmake-dependencies.html#maintenance
I've sent invitation, please check your email.

> Actually, I have a question too, can you point a hunter package at a subdir in an archive?

No you can't right now. And if even such feature is present in CMake I think we should not use it. We should move forward unified project structure instead of allowing hairy customization.

> For noe I've just added a CMakeLists.txt at the root which uses add_subdirectory to add that folder

Yes this is appropriate workaround, you should send it upstream I think.

wheybags:
Ok, marked it as cacheable, moved it into the new repo + added an example.
Sorry it took so long.
As for pushing it upstream, this is the most recent stable release, upstream has changed a lot on master, they've been developing for a while on a series of alpha releases. When they move over to the new branch as stable I'll try push it upstream. I actually added the whole cmake build system, not just the CMakeLists.txt in root, they had a nasty autotools + special vs sln setup.

ruslo:
> can you point a hunter package at a subdir in an archive?

https://github.com/ruslo/hunter/issues/679

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1359
* https://travis-ci.org/ingenue/hunter/builds/203293369

ruslo:
> Testing

Some tests failed. Please exclude broken toolchains (send pull request to [pkg.zookeeper](https://github.com/ingenue/hunter/tree/pkg.zookeeper), see [README](https://github.com/ingenue/hunter)) or apply fixes.

wheybags:
What should I do about failing the "analyze" toolchain? https://travis-ci.org/ingenue/hunter/jobs/203293374

Seems like possible bugs in the library, but that's an upstream issue really.

wheybags:
Ok, it should work now.
Where can I find the results of CI builds? It's a rather odd setup you've got with builds running on separate branches of a fork.

cmake/projects/zookeeper/hunter.cmake
```diff
@@ -0,0 +1,27 @@
...
+include(hunter_download)
+include(hunter_pick_scheme)
+include(hunter_cacheable)
+
+# List of versions here...
+hunter_add_version(
+    PACKAGE_NAME
+    zookeeper
+    VERSION
+    "3.4.9-p1"
```
ruslo: I think it make sense if version and URL matches. I.e. it should be `3.4.9-hunter-7`.

ruslo:
> Where can I find the results of CI builds? It's a rather odd setup you've got with builds running on separate branches of a fork.

It's here:
* https://github.com/ingenue/hunter/branches/all

You can fetch `pkg.zookeeper` and test it with your own fork.

cmake/projects/zookeeper/hunter.cmake
wheybags: I figure I should only increment the hunter package version for releases that actually got released, whereas everything so far has just been internal testing releases.
I can delete the old releases though and make a new one that matches the name in the package file.

wheybags:
How are new CI builds triggered though?
I'd like to see the whole set again.

wheybags:
fixed the version mismatch

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1375
* https://travis-ci.org/ingenue/hunter/builds/205493068

ruslo:
> Testing

Some tests failed. Follow https://github.com/ruslo/hunter/pull/639#issuecomment-280970213

wheybags:
Should work now.
What should I do about the analyze toolchains? Just exclude them? I have made a PR for that here: https://github.com/ingenue/hunter/pull/51

cmake/configs/default.cmake
```diff
@@ -232,3 +232,5 @@ if(ANDROID)
     )
   endif()
 endif()
+
+hunter_config(zookeeper VERSION 3.4.9-p1)
```
ruslo: Should be `3.4.9-p2`

ruslo:
> What should I do about the analyze toolchains? Just exclude them?

Yes, if you have no plans to fix it right now, just exclude them.

>  I have made a PR for that here: ingenue#51

Thanks, merged

wheybags:
ok, fixed default version, sorry, that was stupid

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1381
* https://travis-ci.org/ingenue/hunter/builds/205512784

ruslo:
MinGW still not working

wheybags:
Ok, I've had enough of messing with mingw: https://github.com/ingenue/hunter/pull/52

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1384
* https://travis-ci.org/ingenue/hunter/builds/205531688

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1386
* https://travis-ci.org/ingenue/hunter/builds/205535170

ruslo:
Released:
* https://github.com/ruslo/hunter/releases/tag/v0.18.5

Issue #521, opened 2016-09-13
HUNTER_DOWNLOAD_SERVER for custom download directory location


ruslo:
Use `HUNTER_DOWNLOAD_SERVER` variable to specify custom location of directory with downloaded archives having same format as `${HUNTER_ROOT}/_Base/Download`. Example:
- default location: `https://downloads.sourceforge.net/project/boost/boost/1.61.0/boost_1_61_0.tar.bz2`
  - downloaded to `${HUNTER_ROOT}/_Base/Download/Boost/1.61.0/f84b1a1/boost_1_61_0.tar.bz2`
- if `HUNTER_DOWNLOAD_SERVER=https://example.com/hunter-downloads` then effective URL became `https://example.com/hunter-downloads/Boost/1.61.0/f84b1a1/boost_1_61_0.tar.bz2`

`HUNTER_DOWNLOAD_SERVER` with a local path can be useful for:
- testing, multiple HUNTER_ROOT directories; one git-base, another "normal" one; both have shared `Downloads` directory
- read-only network location

We need extract file name from given URL. Can have problems similar to https://github.com/ruslo/hunter/issues/455.
##

ruslo:
Fixed by https://github.com/ruslo/hunter/pull/1084

PR #646, opened 2017-01-16
Boost: use base url for official releases

NeroBurner:
I'd like to propose the use of base urls for official releases.

I have build machines not connected to the internet using hunter. I provide a LAN available web server with the official released tar balls. I have to modify hunter for every new release of an internal used package. With this change I could easily integrate new releases

### DIFF
--- a/cmake/projects/Boost/hunter.cmake
+++ b/cmake/projects/Boost/hunter.cmake
@@ -12,13 +12,16 @@ include(hunter_pick_scheme)
 # Disable searching in locations not specified by these hint variables.
 set(Boost_NO_SYSTEM_PATHS ON)
 
+# use base url for official boost releases
+set(Boost_base_url "https://downloads.sourceforge.net/project/boost/boost/")
+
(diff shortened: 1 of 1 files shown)


cmake/projects/Boost/hunter.cmake
```diff
@@ -12,13 +12,16 @@ include(hunter_pick_scheme)
 # Disable searching in locations not specified by these hint variables.
 set(Boost_NO_SYSTEM_PATHS ON)
 
+# use base url for official boost releases
+set(Boost_base_url "https://downloads.sourceforge.net/project/boost/boost/")
```
ruslo: We are in include-like file, let it be `_hunter_boost_base_url`.

ruslo:
> I provide a LAN available web server with the official released tar balls

We have already discussed it, just as a reminder:
* https://github.com/ruslo/hunter/issues/521

NeroBurner:
new commit using `_hunter_boost_base_url`

PR #647, opened 2017-01-16
Update packages in docs

caseymcc:


ruslo:
Thanks, merged. I've split packages into sections (https://github.com/ruslo/hunter/pull/649) want to have layout similar to this one in the end:
* http://docs.ansible.com/ansible/modules_by_category.html

It will looks more consistent when `pkg.*` pages will be part of RST documentation. Probably some conversion script needed for this job.

Example of new RST-based pages:
* https://docs.hunter.sh/en/latest/packages/computer-vision.html

PR #648, opened 2017-01-16
Add ceres-solver v1.12.0-p0

NeroBurner:
this version compiles without any sparse matrix support


NeroBurner:
I don't know if it is relocatable. It generates a cmake config file, but it also has hint paths from the dependencies at compile time
https://github.com/hunter-packages/ceres-solver/blob/hunter-1.12.0/cmake/CeresConfig.cmake.in#L207

cmake/projects/ceres-solver/hunter.cmake
```diff
@@ -0,0 +1,35 @@
...
+    URL
+    "${_hunter_ceres_base_url_fork}/v1.12.0-p0.tar.gz"
+    SHA1
+    c8a24d83bf4b26b99fd8fc3bed28a267e6247c85
+)
+# explicitly remove dependency on gflags (only needed for tests)
+hunter_cmake_args(ceres-solver CMAKE_ARGS
+    GFLAGS=OFF
+    BUILD_TESTING=OFF
+    CMAKE_BUILD_TYPE=Release
```
ruslo: This variable will be set by CMake build scheme

ruslo:
> It generates a cmake config file, but it also has hint paths from the dependencies at compile time

Yes, doesn't look like a relocatable, `lib/cmake/Ceres/CeresConfig.cmake` contains full path to Eigen and Glog.

NeroBurner:
removed `CMAKE_BUILD_TYPE` hope that was the variable you meant

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1328
* https://travis-ci.org/ingenue/hunter/builds/193002177

ruslo:
> Testing

Some tests failed. @NeroBurner please proceed as usual, branch [pkg.ceres-solver](https://github.com/ingenue/hunter/tree/pkg.ceres-solver).

NeroBurner:
I didn't yet need to do 'deep per-pkg testing'. Could you explain what 'proceed as usual' means?

ruslo:
> I didn't yet need to do 'deep per-pkg testing'

We need to make everything "green" so I can easily check that new version doesn't break things.

> Could you explain what 'proceed as usual' means?

Ooops, sorry. I just realized that you've never send pull request to ingenue/hunter. So we have some tests failed. It may be valid reason that signals about issue in packages or some invalid reason like missing c++11 support from compiler. Again - goal is to make everything "green", either by fixing package or by excluding broken toolchains. If you're not interested in deep testing then just exclude "red" toolchains from `.travis.yml`/`appveyor.yml` and leave at least link to the broken jobs. See for example:
* https://github.com/ingenue/hunter/pull/46/files

Please send pull request to [pkg.ceres-solver](https://github.com/ingenue/hunter/tree/pkg.ceres-solver) branch with changes only of `.travis.yml`/`appveyor.yml` (see [README](https://github.com/ingenue/hunter)).

caseymcc:
This one should be good with the changes I made to the test. I can work on the iOS version later but for now this should work.

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1484
* https://travis-ci.org/ingenue/hunter/builds/216809272

ruslo:
Released:
* https://github.com/ruslo/hunter/releases/tag/v0.18.30

Feel free to add `pkg.ceres-solver` wiki page and update list here:

caseymcc:
all the package docs refer back to the wiki, do we want to start adding pkg.<package>.rs files to the documentation and link back to that rather than the wiki?

ruslo:
> all the package docs refer back to the wiki, do we want to start adding pkg.packagename.rs files to the documentation and link back to that rather than the wiki?

I was thinking that we need the script, so for now it's still new `pkg.*` wiki pages and leaving the link in `*.rst` docs. See https://github.com/ruslo/hunter/issues/677.

PR #658, opened 2017-01-25
Get the number of CPU cores correctly.

nicber:
NUMBER_OF_LOGICAL_CORES is actually the number of logical cores per physical core.
I tested it on Windows, not on any other OS.
Excerpt from SystemInformation.cxx from CMake 3.7.1:
```
    this->NumberOfLogicalCPU = this->LogicalCPUPerPhysicalCPU();
```


cmake/modules/hunter_jobs_number.cmake
```diff
@@ -11,9 +11,14 @@ function(hunter_jobs_number jobs_options_varname toolchain_path)
...
     cmake_host_system_information(
-        RESULT jobs_number
+        RESULT l_cores
         QUERY NUMBER_OF_LOGICAL_CORES
     )
+	cmake_host_system_information(
+        RESULT p_cores
+        QUERY NUMBER_OF_PHYSICAL_CORES
+    )
+	math(EXPR jobs_number "${l_cores} * ${p_cores}")
```
ruslo: No tabs:
* http://0.readthedocs.io/en/latest/git.html#never-use-tabs

cmake/modules/hunter_jobs_number.cmake
```diff
@@ -11,9 +11,14 @@ function(hunter_jobs_number jobs_options_varname toolchain_path)
   string(COMPARE EQUAL "${HUNTER_JOBS_NUMBER}" "" use_default_jobs)
   if(use_default_jobs)
     cmake_host_system_information(
```
ruslo: > Excerpt from SystemInformation.cxx from CMake 3.7.1
> this->NumberOfLogicalCPU = this->LogicalCPUPerPhysicalCPU();

Please leave this comments here.

cmake/modules/hunter_jobs_number.cmake
ruslo: Please check that neither `l_cores` nor `p_cores` is 0 (substitute it with `1`). For example I got `l_cores=16`, `p_cores=0` on Windows + CMake 3.7.2 with [my processor](http://ark.intel.com/products/78931/Intel-Core-i7-4710MQ-Processor-6M-Cache-up-to-3_50-GHz) (which has 8 cores in fact).

cmake/modules/hunter_jobs_number.cmake
ruslo: Actually it would be nice if you can send documentation patch to [CMake](https://gitlab.kitware.com/cmake/cmake).

cmake/modules/hunter_jobs_number.cmake
```diff
@@ -10,16 +10,36 @@ function(hunter_jobs_number jobs_options_varname toolchain_path)
...
+    string(COMPARE EQUAL "${p_cores}" "0" p_cores_is_zero)
+    if(p_cores_is_zero)
+      hunter_status_debug("Number of physical cores is 0 - forcing 1")
+      set(p_cores "1")
+    else()
+      hunter_status_debug("Number of physical cores: ${p_cores}")
+    endif()
+    math(EXPR jobs_number "${l_cores} * ${p_cores}")
+    hunter_status_debug("Total number of cores: ${jobs_number}")
     endif()
```
ruslo: Syntax error

ruslo: Testing info can be found here:
* https://github.com/ruslo/hunter/wiki/dev.hunter.root

ruslo:
@nicber Looks great, merged. Thank you!

PR #672, opened 2017-02-09
Adds GBenchmark

mdavezac:


mdavezac:
This is a follow up of #665, splitting GBenchmark into its own pull-request

mdavezac:
Also, I've added an example and allowed for debug builds.

cmake/find/FindGBenchmark.cmake
```diff
@@ -0,0 +1,34 @@
+# Copyright (c) 2015-2016, University College London
```
ruslo: Instead of adding `FindGBenchmark.cmake` it's better to patch CMakeLists.txt so `GBenchmarkConfig.cmake` generated automatically. There are a few pitfalls here, for example this code doesn't work correctly with Release/Debug build types, custom build types. For Windows static library have '*.lib' suffix, not '*.a'. Also I'm not sure about `find_library` call. It probably can find system library instead of Hunter one. And components is not handled, this configuration should report error:
```cmake
find_package(GBenchmark REQUIRED COMPONENTS missing)
```

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1349
* https://travis-ci.org/ingenue/hunter/builds/200011911

cmake/find/FindGBenchmark.cmake
mdavezac: Let me see if the maintainers of the code will take a fix to their build.
This pull-request will be on hold until then.

ruslo: Okay. Alternatively we can fork upstream repository and do the fix (it will be faster). I can create [hunter-packages](https://github.com/hunter-packages) for you in this case.

headupinclouds: @mdavezac: Any updates?

mdavezac:
No, apologies. I have not gotten to it yet.

ruslo:
Closing inactive pull request. Original issue https://github.com/ruslo/hunter/issues/332.

chfast:
Does it require any upstream changes?

ruslo:
@chfast I guess so, not sure. Generating `*Config.cmake`?

mdavezac:
I think the main change would have been adding FindGBenchmark.cmake to the upstream.

chfast:
@mdavezac The FindGBenchmark.cmake is not to be in the upstream project. It probably requires exporting config scripts. I will handle this hopefully soon.

Are you fine with the name GBenchmark? Considering we have GTest and GMock already this sounds ok.

ruslo:
> Are you fine with the name GBenchmark?

From https://github.com/google/benchmark:
* `#include <benchmark/benchmark.h>`
* `benchmark::State`

I guess it should be `benchmark`.

Naming:
* https://docs.hunter.sh/en/latest/creating-new/create/cmake.html#add-package-to-hunter

PR #674, opened 2017-02-10
mpark_variant

tobimpub:
Add 'mpark_variant' package


ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1353
* https://travis-ci.org/ingenue/hunter/builds/201551045

cmake/projects/mpark_variant/hunter.cmake
```diff
@@ -0,0 +1,31 @@
...
+    "https://github.com/mpark/variant/archive/v1.0.0.tar.gz"
+    SHA1
+    59ab98b61daa335ba1fa613e591de743f72d24b4
+)
+
+# Probably more versions for real packages...
+
+# Pick a download scheme
+hunter_pick_scheme(DEFAULT url_sha1_cmake) # use scheme for cmake projects
+
```
ruslo: Is the package [cacheable](https://docs.hunter.sh/en/latest/reference/user-modules/hunter_cacheable.html)?

tobimpub: it's a header-only library.
should cacheable be added to those?

ruslo: Yes

ruslo:
> Testing

All Linux and Windows tests failed already.

tobimpub:
added CMAKE_CXX_STANDARD 14 to the examle

@ruslo if you take a look at the requirements table in [here](https://github.com/mpark/variant#requirements) you will find that gcc is only supported from 5.4 on. Testing with 4.x is probably pointless. 
The VC version is also slightly to old with version 19.0.24215 on appveyor.

Is there a precedent for such cases?

cmake/projects/mpark_variant/hunter.cmake
tobimpub: is updated

ruslo:
> The VC version is also slightly to old with version 19.0.24215 on appveyor

AppVeyor use Visual Studio 2015 with Update 3 too:
* https://www.appveyor.com/docs/installed-software/#visual-studio-2015

I think the difference is about compiler, in documentation Clang/LLVM mentioned.

> Is there a precedent for such cases?

I'm afraid there is no such toolchains in current testing. If you want to upgrade you should add one to Polly first:
* https://github.com/ruslo/polly

I'm pretty sure `.travis.yml` upgrade needed too.

ruslo:
Merged without testing. Released:
* https://github.com/ruslo/hunter/releases/tag/v0.18.1

tobimpub:
great!
I think i'll hold off adding in the test case for now.

thank you for the help.

PR #675, opened 2017-02-11
index.rst: split into paragraphs; general copyedit

waldyrious:
Breaking a large block of text into paragraphs makes it easier to read. Also, some minor English issues were fixed.

By the way: is there a guide to how people can contribute to finalize the migration of the docs from the wiki? The warning on this page provides no indication of what needs to be done.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -13,31 +13,37 @@ Hunter: organize freedom
 
 Welcome to the Hunter package manager documentation!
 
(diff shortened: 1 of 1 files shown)


ruslo:
> Breaking a large block of text into paragraphs makes it easier to read. Also, some minor English issues were fixed.

Thanks, merged.

> By the way: is there a guide to how people can contribute to finalize the migration of the docs from the wiki?

Let's start with this one: https://github.com/ruslo/hunter/issues/677

PR #692, opened 2017-03-10
Weaken CMake version check for Android

jomof:


ruslo:
It's not that simple. There are different set of variables which will provide/set information about Android build. Starting with CMake 3.7.1 there are standard set introduced, Hunter is using it since 0.17:
* https://docs.hunter.sh/en/latest/quick-start/cmake.html

Before Hunter was using `taka-no-me` toolchain:
* http://polly.readthedocs.io/en/latest/toolchains/android/old.html

In general I don't know how to support Android Studio nicely while it's using builtin CMake 3.6 with custom toolchain and one more set of Android variables. Supporting both sets will be a huge mess I think. There is an openned discussion already, please leave any notes there:
* https://github.com/ruslo/hunter/issues/618

xsacha:
I still have not been able to get the new Android 3.7.1+ codepath to work at all.
Even trying to use just gcc-4.9, it fails spectacularly.
It complains about:

> cannot find crtend_android.o: No such file or directory

and does not respect CMAKE_SYSROOT to find this.
On the second attempt to compile, it will use host compiler instead!

My (failed) attempt:
`cmake $ROOT -DCMAKE_SYSROOT=${ANDROID_NDK}/platforms/android-${ANDROID_NATIVE_API_LEVEL}/arch-arm64 -DCMAKE_BUILD_TYPE=$CONFIG -DCMAKE_SYSTEM_NAME=Android -DCMAKE_SYSTEM_VERSION=${ANDROID_NATIVE_API_LEVEL} -DCMAKE_ANDROID_NDK=$ANDROID_NDK -DCMAKE_ANDROID_NDK_TOOLCHAIN_VERSION=4.9 -DANDROID_NATIVE_API_LEVEL=${ANDROID_NATIVE_API_LEVEL} -DCMAKE_ANDROID_ARCH_ABI=${ANDROID_ABI} -DANDROID_NDK_ABI_NAME=${ANDROID_ABI}`

Note: Also doesn't work with polly scripts

It is very hard to use and I wonder if anyone has got it working (outside of a test that compiles only one library). Having the ability to use the older version is ideal so I can continue updating Hunter until the new code path is fixed.

xsacha:
Maybe if we can fix up the new Android 3.7.1 codepath to work as well as the old one, it would be fine. Right now I can't compile a lot of libraries.

Using polly or even manually setting up android, it gives me this when compiling, for example, OpenCV:

> [ 61%] Building CXX object modules/imgcodecs/CMakeFiles/opencv_imgcodecs.dir/src/loadsave.cpp.o
> In file included from /usr/include/features.h:434:0,
>                  from /usr/include/bits/libc-header-start.h:33,
>                  from /usr/include/limits.h:26,
>                  from /home/sacha/.hunter/_Base/ebea5d2/9f20262/8a35d48/Build/OpenCV/Source/modules/hal/include/opencv2/hal/defs.h:52,
>                  from /home/sacha/.hunter/_Base/ebea5d2/9f20262/8a35d48/Build/OpenCV/Source/modules/core/include/opencv2/core/cvdef.h:59,
>                  from /home/sacha/.hunter/_Base/ebea5d2/9f20262/8a35d48/Build/OpenCV/Source/modules/core/include/opencv2/core.hpp:52,
>                  from /home/sacha/.hunter/_Base/ebea5d2/9f20262/8a35d48/Build/OpenCV/Source/modules/imgcodecs/include/opencv2/imgcodecs.hpp:46,
>                  from /home/sacha/.hunter/_Base/ebea5d2/9f20262/8a35d48/Build/OpenCV/Source/modules/imgcodecs/src/precomp.hpp:45,
>                  from /home/sacha/.hunter/_Base/ebea5d2/9f20262/8a35d48/Build/OpenCV/Source/modules/imgcodecs/src/loadsave.cpp:46:
> /usr/include/gnu/stubs.h:7:27: fatal error: gnu/stubs-32.h: No such file or directory
>  # include <gnu/stubs-32.h>

ruslo:
Android Studio doesn't work well currently, the reason and detailes can be found here:
* https://github.com/ruslo/hunter/issues/618

Closing.

Issue #164, opened 2015-06-08
BZip2 is broken on iOS


ruslo:
As mentioned [here](https://github.com/ruslo/hunter/issues/110) if you don't need `bzip2` in `boost.iostreams` you can use `IOSTREAMS_NO_BZIP2=1`

@headupinclouds Let me know if this is critical for you I will take a look. As far as I remember the problem is that `bzip2` package build some executable. So I guess it will be not so hard to apply patch with fix
##

headupinclouds:
I don't need bzip2.  This is the solution I've been using.  Thanks.

ruslo:
Fixed https://github.com/ruslo/hunter/pull/700

PR #700, opened 2017-03-15
BZip2 1.0.6-p2

ruslo:


ruslo:
Fix https://github.com/ruslo/hunter/issues/164

PR #714, opened 2017-03-21
add FakeIt to docs

dvirtz:


ruslo:
Thanks, updated pages:
* https://docs.hunter.sh/en/latest/packages/testing.html

PR #717, opened 2017-03-23
Update sparsehash download and SHA1

akalsi87:


cmake/projects/sparsehash/hunter.cmake
```diff
@@ -19,17 +18,13 @@ hunter_add_version(
   VERSION
   "2.0.2"
   URL
-  https://github.com/aadityakalsi/sparsehash/archive/sparsehash-2.0.2-cmake.tar.gz
+  https://github.com/aadityakalsi/sparsehash/archive/v2.0.2-hunter.tar.gz
```
ruslo: Do you want me to create `hunter-packages` repository for `sparsehash`?

cmake/projects/sparsehash/hunter.cmake
```diff
@@ -19,17 +18,13 @@ hunter_add_version(
   VERSION
   "2.0.2"
```
ruslo: I think it make sense to show here that version is not clean and was patches, e.g. use `v2.0.2-hunter` instead of `2.0.2`.

cmake/projects/sparsehash/hunter.cmake
akalsi87: yes, that makes sense..

ruslo: Done: https://github.com/hunter-packages/sparsehash

akalsi87:
@ruslo Can you merge the current branch? We can open another issue for the changes. It's better to fix the broken build, and add non-functional fixes later?

ruslo:
> Can you merge the current branch?

No, there is no example introduced. This package has devastating effect on all root directory, I can't merge it before running tests.

ruslo:

PR #720, opened 2017-03-26
Add imshow library for glfw windowing

headupinclouds:


ruslo:
We need an example for testing. Closing for https://github.com/ruslo/hunter/pull/721.

PR #723, opened 2017-03-27
OpenSSL 1.1.0e

ruslo:


ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1470
* https://travis-ci.org/ingenue/hunter/builds/215460180

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1472
* https://travis-ci.org/ingenue/hunter/builds/215533558

ruslo:
Released:
* https://github.com/ruslo/hunter/releases/tag/v0.18.27

zhuhaow:
I run a few experiments and it seems like there is no other option to compile OpenSSL 1.1.0 for Windows without NASM. Though it is possible to use `no-asm` to run the perl configuration script, it still tries to use `ml64` and spits some errors when compiling (even if I add the path of ml64).

```
        set ASM=ml64
        "C:\Perl\bin\perl.exe" "ms\uplink-x86_64.pl" "auto" crypto\uplink-x86_64
.asm
no assembler found on %PATH% at ms\../crypto/perlasm/x86_64-xlate.pl line 98.
        ml64 /c /Cp /Cx /Zi /Focrypto\uplink-x86_64.obj "crypto\uplink-x86_64.as
m"
Microsoft (R) Macro Assembler (x64) Version 14.00.24210.0
Copyright (C) Microsoft Corporation.  All rights reserved.

 Assembling: crypto\uplink-x86_64.asm
crypto\uplink-x86_64.asm(1) : error A2088:END directive required at end of file
NMAKE : fatal error U1077: '"c:\Program Files (x86)\Microsoft Visual Studio 14.0
\VC\bin\amd64\ml64.EXE"' : return code '0x1'
Stop.
NMAKE : fatal error U1077: '"c:\Program Files (x86)\Microsoft Visual Studio 14.0
\VC\bin\amd64\nmake.exe"' : return code '0x2'
Stop.
```

As per the issue [here](https://github.com/openssl/openssl/issues/967), nasm will be a requirement for build OpenSSL and `no-asm` should not be used since it's only for debugging.
I'm wondering if it is possible to download nasm (it's less then 500KB compressed) by the CMake script. I'm not quite sure if it violates some principle of Hunter because it does adds a few things that are not part of OpenSSL. If that's a problem, we can either:
1. Put it in a temp folder and delete it afterwards
2. Just leave it in the source code folder.

Anyway, I would like to help to fix the OpenSSL building on Windows. Do let me know your thoughts.

ruslo:
@zhuhaow what is the point of adding unrelated comment to the **pull request** that is **closed** already? Please open new separate issue.

PR #730, opened 2017-03-29
Add pkg.doctest

piribes:


ruslo:
Thanks, updated page:

PR #603, opened 2016-12-01
Added czmq package, with example

ErniBrown:
Added czmq package, as requested by @roalz in #599 

Using [hunter-packages/czmq](https://github.com/hunter-packages/czmq), which add cmake config files and use standard cmake scheme

PR #746, opened 2017-04-04
Czmq

ErniBrown:
Hi,
As discussed in #603 I updated pull request for czmq packages


examples/czmq/main.cpp
```diff
@@ -0,0 +1,17 @@
+#include <czmq.h>
+
+
+int main()
+{
+	zsock_t *push = zsock_new_push ("inproc://example");
```
chfast: Tabs mixed with spaces.

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1534
* https://travis-ci.org/ingenue/hunter/builds/218739913

examples/czmq/main.cpp
ruslo: @ErniBrown it's an example so it's not so critical, however try to avoid tabs, see:
* http://0.readthedocs.io/en/latest/git.html#never-use-tabs

ErniBrown:
Testing finished, I need some help to understand what to do now:

- Appveyor was cancelled, not sure why, should I do something?
- [Travis analyze](https://travis-ci.org/ingenue/hunter/jobs/218739918) failed, but I do not understand what's the error, it seems like the warnings broke the build. I'm not even sure what this test do...
- [Travis Android Crosscompile](https://travis-ci.org/ingenue/hunter/jobs/218739917) failed. It was not a fail in czmq, but in zeromq: during configure zeromq performs a tryrun on some expected returtn values. Since we are not on the target, but on the build host those test cannot be executed. Instead cache variables must be set. Those variables are target dependant, so I don't think it wise to set them into the  zeromq project. They must be set at a higher level, i put them in the cmake toolchain file for my use. Should i add a cache set for the test? Where?

ruslo:
> Appveyor was cancelled, not sure why, should I do something?

I did it, need to reprioritize stuff, all builds failed:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1541

> failed, but I do not understand what's the error, it seems like the warnings broke the build. I'm not even sure what this test do...

That's how Clang Analyzer works, it emits warnings. Polly's analyze toolchain use script which exits with 1 if warning found.

> It was not a fail in czmq, but in zeromq

Same logic, exclude toolchain of fix zeromq.

> I do not have a mac machine to test this, maybe I shold disable the test?

Yes, just exclude it.

ErniBrown:
I only removed the tab from the example, it should not affect testing if you want to stop them

ruslo:
> I only removed the tab from the example, it should not affect testing if you want to stop them

Okay, thanks. Build cancelled.

ErniBrown:
I updated the pull request in order to support czmq version 4.0.2

ruslo:
@ErniBrown Thanks, please squash everything into one commit.

ErniBrown:
Do you need some more improvements? Do you think the pull request will be usefull for hunter?

ruslo:
> Do you need some more improvements? Do you think the pull request will be usefull for hunter?

I haven't received notification about the last update. Please send some message explicitly next time. If you don't want to use GitHub for it, you can do it with Gitter (public/private room).

Testing:
* https://travis-ci.org/ingenue/hunter/builds/240085169?utm_source=github_status&utm_medium=notification
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1804

ruslo:
> Testing

All Windows tests failed.

ErniBrown:
I installed visual studio a couple days ago, but it's a long time since I used it, and a lot has changed: at first i was not even able to open a project. I'm quite busy at the moment, so It will take me some time to check windows tests.

ruslo:
In this case it will be easier to just exclude them for now. Use `dummy` toolchain as the only one working (testing table of appveyor.yml can't be empty). Branch [pkg.czmq](https://github.com/ingenue/hunter/tree/pkg.czmq).

ErniBrown:
@ruslo I disabled windows testing, nothing else was changed

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1858
* https://travis-ci.org/ingenue/hunter/builds/244463364?utm_source=github_status&utm_medium=notification

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.19.15

ruslo:
All tests failed when test use uploaded binaries:
* https://travis-ci.org/ingenue/hunter/builds/244470956?utm_source=github_status&utm_medium=notification

meaning that package is not relocatable in fact.

You can check it by:
```
> TOOLCHAIN=gcc PROJECT_DIR=examples/czmq ./jenkins.py --verbose --clear-except
> grep -IR '/_Base/' /.../_testing/Hunter/_Base/.../Install/lib/cmake/czmq/
/.../lib/cmake/czmq/czmqTargets.cmake:
  INTERFACE_LINK_LIBRARIES
  "/.../_Base/.../Install/lib/libzmq.so;/usr/lib/x86_64-linux-gnu/libuuid.so;-lpthread"
```

As you can see full path to libzmq.so saved here.

Also by default we should build static library.

Read this wiki page:
* https://github.com/ruslo/hunter/wiki/example.find_package.config

ErniBrown:
Ok, I switched to config-mode, but I've got a question: the [original czmq cmake](https://github.com/zeromq/czmq/blob/master/CMakeLists.txt) links libzmq.so.
What should I do when we choose to build static library? Link static zmq?
Should it be possible to choose a mix where zeromq is shared, and czmq static?

ruslo:
> Ok, I switched to config-mode

It's not only about config-mode, we already using `find_package(czmq REQUIRED CONFIG)`. The error occurs because the `*Config.cmake` file badly generated. Error should be fixed in czmq package itself.

> I've got a question: the original czmq cmake links libzmq.so

This is wrong too

> What should I do when we choose to build static library? Link static zeromq?

Yes, by default libraries are static.

> Should it be possible to choose a mix where zeromq is shared, and czmq static?

It may be tricky but in general you can create any kind of configuration with local `config.cmake`.

ErniBrown:
No, i mean i switched the way czmq finds zeromq. The original cmake use a find in the project root directory. This leads to having the full path in the czmqTargets.cmake. By switching to config mode and namespaced targets i solved this.
I almost solved the shared-static issue. Now I'm able to build everything shared or everything static. Mixed targets would not be so easy, maybe I will work on this later.

I will open a new pull request tomorrow. The fix does not affect directly hunter, but since I have to change czmq i need to update cmake/projects/czmq/hunter.cmake and modify [hunter-packages/czmq](https://github.com/hunter-packages/czmq)

Mystfit:
Is this currently building on Windows? I get a lot of unresolved externals when using the [czmq basic example](https://github.com/ruslo/hunter/tree/master/examples/czmq).

Examining the generated czmq.lib file, I noticed that one of the unresolved externals is marked as UNDEF and not SECT, so I have a suspicion that czmq.lib is either being built as a shared library, or some internal defines aren't being set and the functions are incorrectly exported.

I also noticed that I'm getting a lot of warnings about 'inconsistent dll linkage' so I think that's playing into this issue.

ruslo:
> Is this currently building on Windows?

All toolchains removed from Windows testing, see `pkg.czmq` branch:
* https://github.com/ingenue/hunter/blob/pkg.czmq/appveyor.yml#L12-L46

So I guess no, it's not working on Windows.

> I'm now attempting to find the best place to add this in the hunter configuration.

See https://github.com/hunter-packages/czmq
Branch `hunter-v4.0.2` with latest release: https://github.com/hunter-packages/czmq/tree/hunter-v4.0.2

PR #755, opened 2017-04-06
Update MySQL to 6.1.9, also fixes debug/Windows compile issues

caseymcc:


examples/MySQL-client/CMakeLists.txt
```diff
@@ -9,6 +9,10 @@ project(MySQL-client-example)
 hunter_add_package(MySQL-client)
 find_package(MySQL-client REQUIRED)
 
+if(MSVC) #LIBCMT conflicts with MSVCR
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /NODEFAULTLIB:LIBCMT.lib")
```
ruslo: Looks like mysql-client doesn't respect defaults and use non-default standard library. Is it `/MT` vs `/MD` issue?

ruslo:
Testing:
* https://travis-ci.org/ingenue/hunter/builds/219323206

examples/MySQL-client/CMakeLists.txt
caseymcc: Not sure, I checked on my local build when I saw the issue at first and the lib was compiling correctly with the applications settings and was still giving the issue. I found a stackoverflow question that said this could also be caused when compiling a debug build with a release library, which I thought might be the problem so I tracked down why the debug builds were disable. I had thought that fixed the issue but in the end it did not. So I had to put that in to make sure the builds succeeded.

It was also odd as VC14 only considered this a warning and did not classify this issue as an error like all the previous versions.

cmake/projects/MySQL-client/hunter.cmake
```diff
@@ -19,9 +19,17 @@ hunter_add_version(
...
+    VERSION
+    "6.1.9"
+    URL
+    "https://github.com/hunter-packages/mysql-client/archive/v1.6.9.tar.gz"
+    SHA1
+    c1ba60a09238a2bc29fe79bc13609e0cfb6f489a
+)
 
 # https://github.com/ruslo/hunter/issues/705
 # hunter_cacheable(MySQL-client)
```
caseymcc: I don't think so, internally it is using GET_TARGET_PROPERTY(LIB_LOCATION ${LIB} LOCATION) which I believe is going to be a full directory path.

caseymcc: It should be noted that the code using this is to setup merging the libraries into one. It might not affect the cmake config and therefore might be cache-able. I don't have a project fully setup to use this all just yet. I plan to get around to it in a few days and can test then.

examples/MySQL-client/CMakeLists.txt
caseymcc: Looks like you are right, I didnt see it originally but MySQL has the following:

     IF(NOT WINDOWS_RUNTIME_MD)
         FOREACH(flag 
          CMAKE_C_FLAGS_RELEASE    CMAKE_C_FLAGS_RELWITHDEBINFO 
          CMAKE_C_FLAGS_DEBUG      CMAKE_C_FLAGS_DEBUG_INIT 
          CMAKE_CXX_FLAGS_RELEASE  CMAKE_CXX_FLAGS_RELWITHDEBINFO
          CMAKE_CXX_FLAGS_DEBUG    CMAKE_CXX_FLAGS_DEBUG_INIT)
          STRING(REPLACE "/MD"  "/MT" "${flag}" "${${flag}}")
          STRING(REPLACE "/Zi"  "/Z7" "${flag}" "${${flag}}")
     etc..

I can't see where WINDOWS_RUNTIME_MD would be defined, so I am guessing this is always going to run. If I alter the example to use

    set(CMAKE_CXX_FLAGS_RELEASE "/MT")
    set(CMAKE_CXX_FLAGS_DEBUG "/MTd")

it compiles fine.

cmake/projects/MySQL-client/hunter.cmake
ruslo: You can verify this by simply installing library and `grep`-ing for the full path string.

examples/MySQL-client/CMakeLists.txt
ruslo: > If I alter the example to use ... it compiles fine

I think that package itself should be fixed here, not an example. Everything should be plain simple on the user's side. I personally think that this code should be removed. Another alternative is to add an option `MYSQL_CLIENT_USE_DEFAULTS`, when this option is `TRUE` don't add/change any compiler flags.

cmake/projects/MySQL-client/hunter.cmake
caseymcc: I don't have an OSX machine setup to test against.

examples/MySQL-client/CMakeLists.txt
caseymcc: I am not sure how to fix this. From what I can tell MySQL is going to compile in /MT unless it is overridden by the end user with WINDOWS_RUNTIME_MD, which does not seem to be an accepted practice from the MySQL team. Therefore anything compiled to use MySQL will have to be /MT as well. Unless there is a way to force that from the library then the end user will have to know that they need todo this.

The only other thing I can think todo is modify MySQL to somehow detect the host applications run-time library setup and set WINDOWS_RUNTIME_MD accordingly.

ruslo: > Therefore anything compiled to use MySQL will have to be /MT as well

Yes, that's the problem. Now users who want to use MySQL should recompile all packages with `/MT` instead of default `/MD`. Which is unmaintainable on practice.

> The only other thing I can think todo is modify MySQL to somehow detect the host applications run-time library setup and set WINDOWS_RUNTIME_MD accordingly

We can't know this a priori. First we install library MySQL-client then we build our project and add MySQL-client paths...

Solution is **to not modify flags** at all, either by removing code `STRING(REPLACE "/MD"  "/MT" "${flag}" "${${flag}}")` or by protecting it with option:

```cmake
# MySQL-client/CMakeLists.txt
option(MYSQL_CLIENT_USE_DEFAULTS "CMake default behaviour" OFF)
if(NOT MYSQL_CLIENT_USE_DEFAULTS)
  # old code to modify compiler flags
  ...
  STRING(REPLACE "/MD"  "/MT" "${flag}" "${${flag}}")
endif()
```

```cmake
# In Hunter: MySQL-client/hunter.cmake
...
hunter_cmake_args(MySQL-client CMAKE_ARGS MYSQL_CLIENT_USE_DEFAULTS=ON)
```

caseymcc: Ill just remove the flag modifications, I was worried that the MySQL team may have chosen to use /MT for a reason (performance, reliability, etc...). But it seems that is not the case as the default for the DB and the C++ Connector is either (version 5.55 and above). I am guessing the C Connector has just been left behind.

caseymcc: Got the modification testing, btw is there an easier way to handle testing on my branch.

Generally, I clone the ingenue/pkg.{name}, then pull the branch I am testing from https://github.com/ruslo/hunter or my own hunter repo. Then push that to a pkg.{name} branch on my repo that starts a travis/appveyor test. Once I get the test to succeed then I have to backout the merge from the test brand and then setup a PR to ingenue. Am I doing something more than needs to be or is that just the way it is?

caseymcc:
Ok, I think it looks good now, update the MySQL-client package (replaced release tag) to remove compiler flags. Tests look good:

https://travis-ci.org/caseymcc/hunter/builds/220619867
https://ci.appveyor.com/project/caseymcc/hunter/build/1.0.109

Updated the sha1 on here and removed the travis changes on pkg.mysql-client.

examples/MySQL-client/CMakeLists.txt
ruslo: > Am I doing something more than needs to be or is that just the way it is?

If you want to test it yourself then it sounds like a solution, clone `pkg.{name}` and merge it with your updates.

ruslo:
Testing:
* https://travis-ci.org/ingenue/hunter/builds/220790450

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.18.40

PR #765, opened 2017-04-17
pkg-config search path overridden to <root-id>/{lib,share}/pkconfig

pretyman:
Hi Ruslan, 

You think that maybe it should be guarded with `if(NOT MSVC)`?

```CMake
if(NOT MSVC)
  set(ENV{PKG_CONFIG_LIBDIR} ...)
endif()
```

Should the string containing the paths be split up in 2 variables as the one in the `hunter_autotools_project.cmake` to make it shorter? I just didn't do that way to not pollute the global namespace.

Small change, but with big impact, would be nice to run tests on all packages :->

Thanks

### DIFF
--- a/cmake/modules/hunter_autotools_project.cmake
+++ b/cmake/modules/hunter_autotools_project.cmake
@@ -8,7 +8,7 @@
 #
 # Adds to the environment variables:
 #   PATH=<root-id>/bin
-#   PKG_CONFIG_PATH=<root-id>/{lib,share}/pkgconfig
+#   PKG_CONFIG_LIBDIR=<root-id>/{lib,share}/pkgconfig
 #
 # Adds to autotools flags:
 #   CPPFLAGS=-I<root-id>/include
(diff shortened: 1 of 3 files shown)


cmake/modules/hunter_autotools_project.cmake
```diff
@@ -315,11 +314,6 @@ function(hunter_autotools_project target_name)
       "PATH=${PARAM_GLOBAL_INSTALL_DIR}/bin:${default_path}"
   )
 
-  # PKG_CONFIG_PATH environment variable
```
ruslo: `hunter_finalize` will not be called in scheme unless there is `hunter_add_package` (i.e. dependency). I think this should be kept.

ruslo:
> You think that maybe it should be guarded with if(NOT MSVC)?

Yes, okay.

> 2 variables as the one in the hunter_autotools_project.cmake to make it shorter?

👍 

> Small change, but with big impact, would be nice to run tests on all packages

Not possible on practice. Just choose few packages to cover maximum amount of code/schemes. You need to change [PACKAGE_INTERNAL_DEPS_ID](https://docs.hunter.sh/en/latest/reference/user-modules/hunter_download.html) for those packages.

pretyman:
Updated. Couldn't DRY with the info on PKG_CONFIG_LIB so I cross referenced it with comments.

Will look at packages to bump  PACKAGE_INTERNAL_DEPS_ID

pretyman:
@ruslo Do you recall which package brought in the pkg-config changes into `hunter_autotools_project.cmake`?

ruslo:
> Do you recall which package brought in the pkg-config changes into hunter_autotools_project.cmake?

No, I don't. Please "grep" hunter_autotools_project string. Also not all of them added to CI testing.

I'm testing `hunter_dump_cmake_flags` with `odb-boost`:
* https://github.com/ruslo/hunter/blob/1e4523f79c7c7d5dd03735854f7488715cd6c181/cmake/modules/hunter_dump_cmake_flags.cmake#L12

not sure if it fits.

pretyman:
Updated: bumped OpenCV's PACKAGE_INTERNAL_DEPS_ID

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1599
* https://travis-ci.org/ingenue/hunter/builds/224828047

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1600
* https://travis-ci.org/ingenue/hunter/builds/224845985

ruslo:
Released:
* https://github.com/ruslo/hunter/releases/tag/v0.18.44

PR #776, opened 2017-04-27
examples: don't use cache servers for testing

NeroBurner:
The examples are used for testing. Is it intended, that these tests use the cache server instead of really compiling the packages?
If it was not intended this patch disables the use of hunter cache servers

### DIFF
--- a/examples/common.cmake
+++ b/examples/common.cmake
@@ -3,6 +3,9 @@
 
 cmake_minimum_required(VERSION 3.0)
 
+# don't use cache servers
+set(HUNTER_USE_CACHE_SERVERS NO)
+
 ### Include HunterGate module from git submodule
 set(gate_dir "${CMAKE_CURRENT_LIST_DIR}/../gate")
 set(gate_module "${gate_dir}/cmake/HunterGate.cmake")


ruslo:
> Is it intended, that these tests use the cache server instead of really compiling the packages?

Yes, because packages can have dependencies. For example with `HUNTER_USE_CACHE_SERVERS=NO` we will not be able to test Qt examples.

The point is: when you're updating sources the SHA1 of package changed and we got "Cache miss" from server. For non-CMake packages when sources don't change but instruction changed developer should update [PACKAGE_INTERNAL_DEPS_ID](https://docs.hunter.sh/en/latest/reference/user-modules/hunter_download.html) to force "Cache miss".

NeroBurner:
Thanks for the fast response :+1: 

I don't fully understand your answer, some follow up questions:
- packages can have dependencies: the dependencies are downloaded from cache to save the user the compilation time. Or do you mean binary dependencies specified by `PACKAGE_DEPENDS_ON`?
- not be able to test Qt examples: I don't understand why this would be the case without the cache server. Shouldn't hunter build all the needed dependencies? For a user which has a different toolchain the cache server will just have misses and the qt-dependencies need to be compiled anyway. Does it have to do something with `PACKAGE_DEPENDS_ON` flag?

ruslo:
> packages can have dependencies: the dependencies are downloaded from cache to save the user the compilation time

Yes. For example we can test Qt examples and related modules without building Qt from scratch on Travis (we can't, because of timeouts).

> Or do you mean binary dependencies specified by PACKAGE_DEPENDS_ON?

As document mentioned PACKAGE_DEPENDS_ON is a workaround, forget about it :)

> For a user which has a different toolchain the cache server will just have misses and the qt-dependencies need to be compiled anyway

Yes, so what? Not sure I understand this part of question.

NeroBurner:
Now it is much clearer to me!
Because of the cache only changed packages are compiled from source, other packages which did not change (and dependencies of the changed package) are not recompiled

nevermind the last question :)

ruslo:
Okay, I'm glad it helps 👍

PR #742, opened 2017-04-03
Add benchmark package

piribes:
Use a hunterized fork.
PR: [benchmark#365](https://github.com/google/benchmark/pull/363)

PR #788, opened 2017-05-10
Add: cpp_redis

szatan:
Adds cpp_redis, a Redis client from https://github.com/Cylix/cpp_redis

### DIFF
--- a/cmake/configs/default.cmake
+++ b/cmake/configs/default.cmake
@@ -265,3 +265,4 @@ endif()
 
 hunter_config(zookeeper VERSION 3.4.9-p2)
 hunter_config(tacopie VERSION 2.4.0-h1)
+hunter_config(cpp_redis VERSION 3.5.0-h1)
--- a/cmake/projects/cpp_redis/hunter.cmake
+++ b/cmake/projects/cpp_redis/hunter.cmake
@@ -1,0 +1,31 @@
+# cmake/projects/cpp_redis/hunter.cmake
+
+# !!! DO NOT PLACE HEADER GUARDS HERE !!!
+
+# Load used modules
+include(hunter_add_version)
+include(hunter_download)
+include(hunter_pick_scheme)
+include(hunter_cacheable)
+
+# List of versions here...
+hunter_add_version(
+    PACKAGE_NAME
+    cpp_redis
+    VERSION
+    "3.5.0-h1"
+    URL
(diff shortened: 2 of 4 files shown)


ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1684
* https://travis-ci.org/ingenue/hunter/builds/230748653?utm_source=github_status&utm_medium=notification

szatan:
Most of the tests in https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1684 fail, but I have no idea why. Most of the time the output just says that the build failed without providing any further explanation.

I'd rather have passing tests than exclude/disable them, but I don't know how to approach it.
Could you point me in the right direction?

ruslo:
> the output just says that the build failed without providing any further explanation

I guess they are executed in parallel and you have to move a little bit upper. First toolchain that failed on Windows:
```
tcp_client_iface.hpp(41) : error C2039: 'uint32_t' : is not a member of 'std'
```

It's missing `#include <cstdint>` I think:
* http://en.cppreference.com/w/cpp/types/integer

ruslo:
I guess if you want to debug toolchains you can comment out all tests except one in your fork and run tests one by one. Some hints can be found in this comment:
* https://github.com/ruslo/hunter/pull/742#issuecomment-293812754

szatan:
This looks like something that should be fixed in the cpp_redis upstream.
I think the best course of action would be to exclude the tests for now, and bug the upstream guys to fix it.
Then, once it's handled, I'll just update the hunterized version of cpp_redis.

Would that be ok with you?

ruslo:
> Would that be ok with you?

Sure, no problems.

szatan:
ok. I pushed the exclusion PR: https://github.com/ingenue/hunter/pull/71

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1688
* https://travis-ci.org/ingenue/hunter/builds/230785934?utm_source=github_status&utm_medium=notification

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1689
* https://travis-ci.org/ingenue/hunter/builds/231064316?utm_source=github_status&utm_medium=notification

ruslo:
Released:
* https://github.com/ruslo/hunter/releases/tag/v0.18.51

PR #831, opened 2017-06-12
Added OpenJPEG package and example

ikliashchou:


ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1836
* https://travis-ci.org/ingenue/hunter/builds/242380905?utm_source=github_status&utm_medium=notification

ruslo:
All tests failed:
```
Target "foo" links to target "OpenJPEG::OpenJPEG" but the target was not found.
```

xsacha:
https://github.com/uclouvain/openjpeg/blob/master/CMakeLists.txt#L20

Looks like namespace is OPENJPEG and library is openjp2

Try linking to OPENJPEG::openjp2

ruslo:
Same:

```
Target "foo" links to target "OpenJPEG::openjp2" but the target was not found.
```

You can test package locally using `jenkins.py` script:
* https://github.com/ruslo/hunter/wiki/dev.hunter.root#jenkinspy

xsacha:
Case sensitive, so the Open should be capitalised as I had it.

ikliashchou:
The issue seems to be not only with library name (it links only if specified like this
`target_link_libraries(foo openjp2)`).
Also header search path is not set so header `#include <openjpeg-2.1/openjpeg.h>
` could not be found.

If I add include folder manually like this `/Users/user/hunter-branch/_Base/xxxxxxx/a15ae7a/f6e1edd/Install/include` into Xcode project header search paths, then example could be successfully built.

Any hints how can I make this path to `include` being added automatically with `find_package`?

xsacha:
The include path didn't work because the target wasn't valid. That's why you have to use OPENJPEG::openjp

The problem is it doesn't install the CONFIG to the correct location so it is not found. It installs to:
lib/openjpeg-2.1/OpenJPEGConfig.cmake
It needs to be cmake/lib/${PROJECT_NAME}

cmake/projects/fmt/hunter.cmake
```diff
@@ -8,6 +8,18 @@ include(hunter_download)
 include(hunter_pick_scheme)
 
 # List of versions here...
+
+hunter_add_version(
```
ruslo: Please create a separate pull request for this change

ikliashchou: Sorry, that what I wanted to do, but missed.

ruslo:
@ikliashchou I think you've destroyed original pull request...

ikliashchou:
I still have to finish this work, so I will try to make it once again when ready.

xsacha:
Although it has no dependencies, I think libopenjp2 will have to be hunterised to get it to work. Just for the config only. Or perhaps this could be done upstream?
https://github.com/uclouvain/openjpeg/blob/master/CMakeLists.txt#L328

ruslo:
> Although it has no dependencies, I think libopenjp2 will have to be modified to get it to work

Yes.

> Or perhaps this could be done upstream but I doubt they will change the versioning system?

If you add `HUNTER_ENABLED=OFF` as a default option value it will not affect original code:
* https://docs.hunter.sh/en/latest/creating-new/cmake-dependencies.html#maintenance

ikliashchou:
We decided not to go with OJ in our project so no more interest for adding this package at the moment. Closing for now.

ikliashchou:
I came back to this question and made some changes to OpenJPEG Cmake that made it compatible with Hunter. @ruslo What should I do next to get it into Hunter?  Will you create a separate fork in hunter-packages, that then can be tagged and referred?

ruslo:
> Will you create a separate fork in hunter-packages, that then can be tagged and referred?

Please open package request issue to discuss details. See the `<!--- Use this part of template if you want to leave a package request. Remove the rest. -->` part from:
- https://raw.githubusercontent.com/ruslo/hunter/master/.github/ISSUE_TEMPLATE.md

Issue #855, opened 2017-06-28
ceres-solver: add suitesparse support


NeroBurner:
## dependencies of ceres-solver:

- [x] Eigen3
- [x] glog
- [x] suitesparse
  - [x] lapack + blas

### lapack
I intend to use lapack from http://netlib.org/lapack/#_lapack_version_3_7_1_2

It has a cmake project, but it requires a fortran compiler. On Ubuntu this is provided by the `gfortran` package. On Windows the mingw-w64 suite ships a gfortran compiler

### suitesparse
there is a project on github providing a cmake project for suitesparse
https://github.com/jlblancoc/suitesparse-metis-for-windows

It works also on Ubuntu, but needs some updates:
- find the current lapack version with the cmake config
- the provided cmake-config exports targets spqr, amd etc instead of namespaced variables
- for using this suitesparse package we need to find the package and then include a usesuitesparse file https://github.com/jlblancoc/suitesparse-metis-for-windows/blob/master/example-projects/spqr/CMakeLists.txt#L26
- transient dependencies is flaky, need to investigate some more

## Questions

- is is possible to have suitesparse as optional dependency of ceres-solver? How would we define to use or not use the optional dependency in the project
- is it possible to define a fortran compiler like mingw-gfortran (build dlls) and build the rest with msvc? 

Otherwise a project using suitesparse (and therefore lapack) on windows needs to be compiled with mingw completely

ruslo:
> is is possible to have suitesparse as optional dependency of ceres-solver? How would we define to use or not use the optional dependency in the project

It depends on ceres-solver, if it can be built without suitesparse then it's possible in practice. You just need to add an option:
```cmake
option(CERES_SOLVER_WITH_SUITESPARSE "Build with Suitesparse support" ON)

if(CERES_SOLVER_WITH_SUITESPARSE)
  find_package(Suitesparse ...)
endif()
```

For enabling/disabling `CERES_SOLVER_WITH_SUITESPARSE` in Hunter see:
* https://github.com/ruslo/hunter/wiki/example.custom.config.id#custom-cmake-options

Of course if Suitesparse will not be in Hunter then `CERES_SOLVER_WITH_SUITESPARSE` should be set to `OFF` by default in package or in `hunter.cmake` by `hunter_cmake_args`.

> is it possible to define a fortran compiler like mingw-gfortran (build dlls) and build the rest with msvc?

It's not possible right now, see feature request:
* https://github.com/ruslo/hunter/issues/495

[...]

PR #860, opened 2017-06-29
add LAPACK 3.7.1

NeroBurner:
- build dynamic libraries
- needs a fortran compiler to build
  - on Ubuntu: `sudo apt install gfortran`
  - on Windows: mingw-w64 has a fortran compiler

dependency for suitesparse as stated in Issue #855


examples/LAPACK/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+find_package(LAPACK CONFIG REQUIRED)
+
+# Test double library creation
+find_package(LAPACK CONFIG REQUIRED)
+
+
+find_package(cblas CONFIG REQUIRED)
+
+add_executable(foo foo.cpp)
+target_include_directories(foo PRIVATE ${CBLAS_INCLUDE_DIRS})
```
ruslo: This should be a property of imported target

examples/LAPACK/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+  endif()
+endforeach()
+
+find_package(LAPACK CONFIG REQUIRED)
+
+# Test double library creation
+find_package(LAPACK CONFIG REQUIRED)
+
+
+find_package(cblas CONFIG REQUIRED)
```
ruslo: This is a little bit confusing. Are we using LAPACK or cblas? According to the C++ code we depends only on cblas, meaning that `find_package(LAPACK ...)` not needed.

cmake/projects/LAPACK/hunter.cmake
```diff
@@ -0,0 +1,46 @@
...
+    "${_hunter_lapack_base_url}/lapack/lapack-3.7.1.tgz"
+    SHA1
+    e71584cc334a7f711b215261c2a35adee49df2f0
+)
+
+# pass cmake arguments
+hunter_cmake_args(
+    LAPACK
+    CMAKE_ARGS
+    BUILD_SHARED_LIBS=ON # build as shared libs
```
ruslo: By default libraries should be static.

examples/LAPACK/foo.cpp
```diff
@@ -0,0 +1,14 @@
+#include <iostream> // std::cout
+#include <cblas.h>
```
ruslo: I think we are conflicting with OpenBLAS package here:
* https://github.com/hunter-packages/OpenBLAS/blob/hunter/cblas.h

Do you know how they are related? Is it two implementation of the same standard?

NeroBurner: I don't know. I think they are solving the same problem in slightly different ways. The creation of cblas can be turned off. I only added it to use it in the example. For suitesparse cblas is not used.

All the different blas, OpenBlas, ATLAS, LAPACK, LAPACKE are confusing me. I know suitesparse can use BLAS and LAPACK. About the others I don't know

examples/LAPACK/CMakeLists.txt
NeroBurner: then I need to patch the sources. Could you make a fork of https://github.com/Reference-LAPACK/lapack so that we can add a hunter branch

examples/LAPACK/foo.cpp
xsacha: When I compile OpenBLAS on Linux, it provides a Lapack as well if gfortran is detected.

examples/LAPACK/CMakeLists.txt
NeroBurner: I think clbas is using lapack/blas and is just a C wrapper for the libraries. I included cblas only to get a C interface for the example. I wasn't able to directly use the lapack/blas libraries

examples/LAPACK/foo.cpp
NeroBurner: I can't cross compile the hunter OpenBlas with mingw-w64. Does it work natively on Windows?

```
[ 75%] Performing build step for 'OpenBLAS'
/bin/sh: 1: ./getarch: not found
make[3]: *** [config.h] Error 127
OpenBLAS: Detecting fortran compiler failed. Cannot compile LAPACK. Only compile BLAS.
gemv.c: In function 'sgemv_':
gemv.c:223:29: error: 'GEMM_MULTITHREAD_THRESHOLD' undeclared (first use in this function)
   if ( 1L * m * n < 2304L * GEMM_MULTITHREAD_THRESHOLD )
                             ^
gemv.c:223:29: note: each undeclared identifier is reported only once for each function it appears in
In file included from ../common.h:747:0,
                 from gemv.c:40:
../common_stackalloc.h:32:46: warning: unused variable 'stack_check' [-Wunused-variable]
 #define STACK_ALLOC_PROTECT_SET volatile int stack_check = 0x7fc01234;
                                              ^
../common_stackalloc.h:56:3: note: in expansion of macro 'STACK_ALLOC_PROTECT_SET'
   STACK_ALLOC_PROTECT_SET                                                  \
   ^
gemv.c:219:3: note: in expansion of macro 'STACK_ALLOC'
   STACK_ALLOC(buffer_size, FLOAT, buffer);
   ^
Makefile:841: recipe for target 'sgemv.obj' failed
make[4]: *** [sgemv.obj] Error 1
Makefile:126: recipe for target 'libs' failed
make[3]: *** [libs] Error 1
```

examples/LAPACK/CMakeLists.txt
NeroBurner: is it possible to build cblas only for the example? Then there isn't a conflict and the user can add a buildflag for his/her project when cblas is really needed

NeroBurner:
I have updated the configuration and added a local config file for the LAPACK example. But when building the example no `cblas-config.cmake` file can be found. I've looked in the install folder and CBLAS just wasn't built/installed. When I add the CMAKE_ARGS back to the `projects/LAPACK/hunter.cmake` file then CBLAS is built again

```
CMake Error at CMakeLists.txt:35 (find_package):
  Could not find a package configuration file provided by "cblas" with any of
  the following names:

    cblasConfig.cmake
    cblas-config.cmake

  Add the installation prefix of "cblas" to CMAKE_PREFIX_PATH or set
  "cblas_DIR" to a directory containing one of the above files.  If "cblas"
  provides a separate development package or SDK, be sure it has been
  installed.

-- Configuring incomplete, errors occurred!
```

examples/LAPACK/CMakeLists.txt
ruslo: Since this pull request is about adding `lapack` I expect at least one simple example with `hunter_add_package(lapack)` + `find_package(lapack ...)` + `#include <lapack.h>` without any other instructions. If there are other important variants of using this package they should be added separately, e.g. `examples/lapack-cblas`. In those examples we can use custom local `config.cmake` with custom `CMAKE_ARGS` for enabling/disabling components.

examples/LAPACK/CMakeLists.txt
ruslo: Done: https://github.com/hunter-packages/lapack

examples/LAPACK/foo.cpp
ruslo: > Does it work natively on Windows?

There are no working toolchains on Windows in `pkg.openblas`:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1047

examples/LAPACK/CMakeLists.txt
NeroBurner: the install of lapack (without CBLAS) provides only libraries, no headers. So there is no `lapack.h` to include.
To have an example with c-interface I added cblas. I'll move the current example to `examples/lapack/cblas` but I don't know how to create an example just using `liblapack.a` and `libblas.a`

NeroBurner:
sent a patch upstream to add the include directories to the target `cblas` https://github.com/Reference-LAPACK/lapack/pull/170

also added the patch on top of v3.7.1 and tagged v3.7.1-p0

the example `LAPACK-CBLAS` does still not work, because the cmake arguments are not passed on to the LAPACK build (the local `config.cmake` is not working). If I set `CBLAS=ON` in `projects/LAPACK/hunter.cmake` then the cblas target is built and installed

NeroBurner:
We could add a components schema to activate CBLAS or LAPACKE (c-interfaces for BLAS and LAPACK)

If a user just wants the libraries he can use
```
hunter_add_package(LAPACK)
```

if he wants the CBLAS interface
```
hunter_add_package(LAPACK COMPONENTS CBLAS)
```

and for both optional c interfaces
if he wants the CBLAS interface
```
hunter_add_package(LAPACK COMPONENTS CBLAS LAPACKE)
```

This would be easier for the user than creating a local `config.cmake` file.

how much work would this be?

examples/LAPACK/CMakeLists.txt
ruslo: > provides only libraries, no headers

How they are expected to be used!? 😕 Probably that's because of it's written in Fortran (?)

examples/LAPACK-CBLAS/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
+# Copyright (c) 2015, Ruslan Baratov
+# All rights reserved.
+
+cmake_minimum_required(VERSION 3.0)
+
+
+set(TESTING_CONFIG_OPT FILEPATH "${CMAKE_SOURCE_DIR}/config.cmake")
```
ruslo: Prefer `CMAKE_CURRENT_LIST_DIR`:
* http://cgold.readthedocs.io/en/latest/tutorials/cmake-sources/includes.html#id3

examples/LAPACK/CMakeLists.txt
NeroBurner: I think the libs are used directly and the users create the appropriate headers themselves. See https://github.com/jlblancoc/suitesparse-metis-for-windows/blob/master/SuiteSparse/CHOLMOD/Include/cholmod_blas.h for an example

I think that CBLAS and LAPACKE came later as an easier way to use the libraries

examples/LAPACK-CBLAS/CMakeLists.txt
NeroBurner: done, thx for the Info :)

NeroBurner:
yes I'm sure the `config.cmake` file is included. I added a `message(FATAL_ERROR "file included")` to the config file and in the configure step of the example the message is shown and the configuration errors out (as intended)

examples/LAPACK-CBLAS/config.cmake
```diff
@@ -0,0 +1,3 @@
+include(hunter_cmake_args)
```
ruslo: Not needed

examples/LAPACK-CBLAS/config.cmake
```diff
@@ -0,0 +1,3 @@
...
+
+hunter_cmake_args(LAPACK CMAKE_ARGS CBLAS=ON)
```
ruslo: It should be `hunter_config`, not `hunter_cmake_args`:

NeroBurner:
that fixed it :+1:

cmake/projects/LAPACK/hunter.cmake
```diff
@@ -0,0 +1,43 @@
...
+    https://github.com/hunter-packages/lapack/archive/v3.7.1-p0.tar.gz
+    SHA1
+    41aafc8d37e9483ffefc4baed41e5fae1a15d69a
+)
+
+# pass cmake arguments
+hunter_cmake_args(
+    LAPACK
+    CMAKE_ARGS
+    #BUILD_SHARED_LIBS=ON # build as shared libs
```
ruslo: Please remove unused commented options.

NeroBurner: done

NeroBurner:
the PR https://github.com/Reference-LAPACK/lapack/pull/170 was merged, CBLAS should now work without any additional patchset.
https://github.com/Reference-LAPACK/lapack/pull/171 to fix LAPACKE is still open, but I hope it will get merged soon

I'd like to recreate the tag v3.7.1-p0 on top of the current master and remove the custom branch hunter_v3.7.1. What are your thoughts on it?

ruslo:
> Reference-LAPACK/lapack#171 to fix LAPACKE is still open, but I hope it will get merged soon

Merged now.

> I'd like to recreate the tag v3.7.1-p0 on top of the current master and remove the custom branch hunter_v3.7.1. What are your thoughts on it?

I think it make sense to keep the `hunter_v3.7.1` branch so it will be clear where other patches will go if needed. Of course you can do `git merge --ff-only` so `master` and `hunter_v3.7.1` will point to the same commit.

NeroBurner:
nice, that was fast :)

ff-only was not possible, did a `git reset --hard master` and recreated the tag `v3.7.1-p0`

what is left to do before merge? I'd like to squash the commits before that please

ruslo:
> what is left to do before merge?

Please add `examples/LAPACK`:

> I think we still need examples/LAPACK. Start with simple hunter_add_package(LAPACK) without find_package and any targets. We can add more info later.

NeroBurner:
added and squashed

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1906
* https://travis-ci.org/ingenue/hunter/builds/249585861?utm_source=github_status&utm_medium=notification

NeroBurner:
I've been trying to use this lapack package with ceres-solver, but when building the examples (or any other binary) I get missing fortran functions. The LAPACK config of static libraries does not properly propagate the dependency to gfortran and quadmath. When I build blas and lapack as dynamic libraries then the fortran libraries get resolved.

With the statically built blas and lapack I can make it work if I add the following to the ceres-solver project
```
enable_language(Fortran)
```

how do we add this dependency to the static blas and lapack libs?

ruslo:
I don't have experience with Fortran so I'm not sure I understand all details.

> With the statically built blas and lapack I can make it work if I add the following to the ceres-solver project

Sounds like you need to update `ceres-solver`.

NeroBurner:
I think the runtime dependency should be propagated by the cmake-configure file of LAPACK, otherwise all projects using LAPACK need to be modified. But I don't know how

ruslo:
I'm not sure that I understand enough all the details of the issue to give any advice here.

NeroBurner:
I think for LAPACK and BLAS it is better to create a dynamic lib (so or dll). Then the consuming projects don't need to be updated.

For Windows it should be possible to let a MSVC project consume the `LAPACK.dll` built by mingw-gfortran or any other fortran compiler.

ruslo:
We can start with that as a workaround and use `BUILD_SHARED_LIBS=ON` only in `examples/LAPACK`. But I'm afraid it will add extra mess in future. Probably you will have to remove all Linux testing because by default toolchains doesn't use `-fPIC` needed for mixing static/shared. Also I don't think mixing static and shared is a good idea in general.

NeroBurner:
the hunter-examples for LAPACK and SuiteSparse build fine with static libs. They don't seem to use fortran provided functions (maybe because they are very basic examples)

The projects LAPACK and SuiteSparse define -fPIC. So they could be linked into another dynamic library. I think that `-fPIC` is only important if you are linking a static library into another dynamic library

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1919
* https://travis-ci.org/ingenue/hunter/builds/250702376?utm_source=github_status&utm_medium=notification

NeroBurner:
looks good :+1:

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1921
* https://travis-ci.org/ingenue/hunter/builds/250844864?utm_source=github_status&utm_medium=notification

ruslo:
Released:
* https://github.com/ruslo/hunter/releases/tag/v0.19.24

PR #866, opened 2017-07-04
Check for presence of msys when configuring boost and redirect to boo…

cdglove:
…ststrap.sh in that case.

Under MSYS, trying to use boost from hunter results in an error message that;

/bin/sh bootstrap.bat is inavlid.

Checking for the presence of MSYS and redirecting to bootstrap.sh, as under cygwin and linux, resolves the issue.


cdglove:
Nevermind, appears to be fixed at head on master.

jgillich:
Is it? I'm getting:

```
/bin/sh: bootstrap.bat: command not found
```

Full log: https://ci.appveyor.com/project/JakobGillich/autominer/build/66

Using hunter v0.19.213 and cmake -G "MSYS Makefiles". Do I need to update hunter? The version is only 10 days old so I assume not.

@ruslo

ruslo:
I don't know. I'm not using MSYS daily and MSYS + Boost is not regularly tested:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.3146
* https://github.com/ingenue/hunter/blob/pkg.boost/appveyor.yml#L46-L48

jgillich:
Alright thanks, I think it should just switch to plain mingw. God I hate Windows. :(

Issue #795, opened 2017-05-15
xorg-macros: `--strip` autoconf issue


headupinclouds:
I noticed this in a `glfw` package build in an ubuntu unit test:

https://travis-ci.org/headupinclouds/ogles_gpgpu/jobs/232261502#L1041-L1048

This isn't critical for me, but I'll try to debug it later this week.  I'm not sure if this is specific to the package or the autoconf scheme stuff.

ruslo:
I guess build script read environment variable `INSTALL`. If so we can add `INSTALL` to this list:
* https://github.com/ruslo/hunter/blob/master/scripts/clear-all.cmake

ruslo:
Fixed by https://github.com/ruslo/hunter/pull/870

PR #870, opened 2017-07-07
clear INSTALL variable

headupinclouds:
see: https://github.com/ruslo/hunter/issues/795


ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.19.26

PR #877, opened 2017-07-15
add stormlib package + example

wheybags:
Can be moved to a hunter-packages fork

### DIFF
--- a/cmake/configs/default.cmake
+++ b/cmake/configs/default.cmake
@@ -108,6 +108,7 @@ hunter_config(SDL2 VERSION 2.0.4-p4)
 hunter_config(SDL_mixer VERSION 2.0.1-p1)
 hunter_config(SQLite3 VERSION autoconf-3080803) #R-Tree enabled
 hunter_config(Sober VERSION 0.1.3)
+hunter_config(stormlib VERSION 9.21-p1)
 hunter_config(Sugar VERSION 1.2.2)
 hunter_config(SuiteSparse VERSION 4.5.1-p0)
 hunter_config(TIFF VERSION 4.0.2-p3)
--- a/cmake/projects/stormlib/hunter.cmake
+++ b/cmake/projects/stormlib/hunter.cmake
@@ -1,0 +1,27 @@
+# cmake/projects/Example/hunter.cmake
+
+# !!! DO NOT PLACE HEADER GUARDS HERE !!!
+
+# Load used modules
+include(hunter_add_version)
+include(hunter_download)
+include(hunter_pick_scheme)
+include(hunter_cacheable)
+
+# List of versions here...
+hunter_add_version(
+    PACKAGE_NAME
+    stormlib
+    VERSION
+    "9.21-p1"
+    URL
+    "https://github.com/hunter-packages/StormLib/archive/v9.21-p1.tar.gz"
+    SHA1
+    db5b811ef5171b163e9abc4d41f1323fb43882a2
+)
+
+# Pick a download scheme
+hunter_pick_scheme(DEFAULT url_sha1_cmake) # use scheme for cmake projects
+
+hunter_cacheable(stormlib)
+hunter_download(PACKAGE_NAME stormlib)
--- a/examples/stormlib/CMakeLists.txt
+++ b/examples/stormlib/CMakeLists.txt
@@ -1,0 +1,14 @@
+cmake_minimum_required(VERSION 3.0)
+
+# Emulate HunterGate:
+# * https://github.com/hunter-packages/gate
+include("../common.cmake")
+
+project(example-stormlib)
+
+hunter_add_package(stormlib)
+
+find_package(stormlib CONFIG REQUIRED)
+
+add_executable(example-stormlib main.cpp)
+target_link_libraries(example-stormlib stormlib::stormlib)
--- a/examples/stormlib/main.cpp
+++ b/examples/stormlib/main.cpp
@@ -1,0 +1,65 @@
+#include <StormLib.h>
+#include <stdio.h>
+
+int ExtractFile(char * szArchiveName, char * szArchivedFile, char * szFileName)
+{
+    HANDLE hMpq   = NULL;          // Open archive handle
+    HANDLE hFile  = NULL;          // Archived file handle
+    FILE* handle = NULL;          // Disk file handle
+    int    nError = ERROR_SUCCESS; // Result value
+
+    // Open an archive, e.g. "d2music.mpq"
+    if(nError == ERROR_SUCCESS)
+    {
+        if(!SFileOpenArchive(szArchiveName, 0, STREAM_FLAG_READ_ONLY, &hMpq))
+            nError = GetLastError();
+    }
+    
+    // Open a file in the archive, e.g. "data\global\music\Act1\tristram.wav"
+    if(nError == ERROR_SUCCESS)            
+    {
+        if (!SFileOpenFileEx(hMpq, szArchivedFile, 0, &hFile))
+            nError = GetLastError();
+    }
+
+    // Create the target file
+    if(nError == ERROR_SUCCESS)
+    {
+        handle = fopen(szFileName, "wb");
+        if (!handle)
+            nError = -1;
+    }
+
+    // Read the file from the archive
+    if(nError == ERROR_SUCCESS)
+    {
+        char  szBuffer[0x10000];
+        DWORD dwBytes = 1;
+
+        while(dwBytes > 0)
+        {
+            SFileReadFile(hFile, szBuffer, sizeof(szBuffer), &dwBytes, NULL);
+            if (dwBytes > 0)
+                fwrite(szBuffer, 1, dwBytes, handle);
+        }
+    }        
+
+    // Cleanup and exit
+    if (handle != NULL)
+        fclose(handle);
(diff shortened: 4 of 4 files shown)


ruslo:
Created: https://github.com/hunter-packages/StormLib

wheybags:
moved

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1966
* https://travis-ci.org/ingenue/hunter/builds/254740745?utm_source=github_status&utm_medium=notification

ruslo:
> /home/travis/build/ingenue/hunter/examples/stormlib/main.cpp:1:22: fatal error: Stormlib.h: No such file or directory

wheybags:
Should be fixed now. Their example code had the wrong case in the header filename. windows...

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.1990
* https://travis-ci.org/ingenue/hunter/builds/256076288?utm_source=github_status&utm_medium=notification

ruslo:
```
Undefined symbols for architecture x86_64:
  "_ltc_mp", referenced from:
```

wheybags:
Ok, finally got an osx vm up to debug this, and I don't get what's going on here at all. It works on every other platform, and I used an object inspector to make sure the symbol was there and it is.
nm -gU on the libtomcrypt static lib outputs:
```
/Users/wheybags/stormlib/build/hunter/_Base/xxxxxxx/f77a954/3b0b532/Install/lib/libtomcryptd.a(crypt_ltc_mp_descriptor.c.o):
0000000000000188 C _ltc_mp
```

It's just a struct, declared here: https://github.com/hunter-packages/libtomcrypt/blob/1.17-p1/src/misc/crypt/crypt_ltc_mp_descriptor.c

and declared as extern here: https://github.com/hunter-packages/libtomcrypt/blob/1.17-p1/src/headers/tomcrypt_math.h#L416

Is there some weirdness on osx around extern structs? Do I need to do some equivalent of the windows style declspec dllimport crap?

@ruslo any idea what's going on here, because I'm stumped :(

ruslo:
Library source is C and you're using it in C++ example. I guess we need `extern C`:
* https://stackoverflow.com/q/3789340/2288008

wheybags:
But the whole thing is only ever included through tomcrypt.h: https://github.com/hunter-packages/libtomcrypt/blob/1.17-hunter/src/headers/tomcrypt.h#L73 , which has an extern c wrapper: https://github.com/hunter-packages/libtomcrypt/blob/1.17-hunter/src/headers/tomcrypt.h#L15

ruslo:
> But the whole thing is only ever included through tomcrypt.h

Are we talking about tomcrypt or stormlib? :)

In `examples/stormlib/main.cpp` I see `#include <StormLib.h>`.

wheybags:
I'm not sure which package contains the problem :(
Stormlib links to libtomcrypt, and uses that symbol in only two places https://github.com/hunter-packages/StormLib/blob/v9.21_hunter/src/SBaseCommon.cpp#L187 and https://github.com/hunter-packages/StormLib/blob/v9.21_hunter/src/libtomcrypt/src/pk/rsa/rsa_verify_simple.c (disregard the weird path on the second one, it was using a bundled version of the lib with that file added in).

Those files are getting the definition of the ltc_mp symbol from https://github.com/hunter-packages/StormLib/blob/v9.21_hunter/src/StormCommon.h#L49 and https://github.com/hunter-packages/StormLib/blob/v9.21_hunter/src/libtomcrypt/src/pk/rsa/rsa_verify_simple.c#L11 , so it should be picking up on the extern "C" from there, and indeed it works on linux and windows.

ruslo:
Looks weird indeed. Add usage of `ltc_mp` to `tomcrypt` example and you will see the same error so it's not `stormlib` issue. May be some mess with defines? While building `tomcrypt` we are using one set and while using `tomcrypt` another set? And linker think that those structs `ltc_math_descriptor` don't match?

wheybags:
oooook, so (hesistant to say this but), this looks like it might a linker bug?

I added an extra symbol into that file, so:
```patch
diff --git a/src/headers/tomcrypt_math.h b/src/headers/tomcrypt_math.h
index aee6105..ba60fba 100644
--- a/src/headers/tomcrypt_math.h
+++ b/src/headers/tomcrypt_math.h
@@ -414,6 +414,7 @@ typedef struct {
 } ltc_math_descriptor;
 
 extern ltc_math_descriptor ltc_mp;
+extern int testInt;
 
 int ltc_init_multi(void **a, ...);
 void ltc_deinit_multi(void *a, ...);
diff --git a/src/misc/crypt/crypt_ltc_mp_descriptor.c b/src/misc/crypt/crypt_ltc_mp_descriptor.c
index 0577d1d..f0c781f 100644
--- a/src/misc/crypt/crypt_ltc_mp_descriptor.c
+++ b/src/misc/crypt/crypt_ltc_mp_descriptor.c
@@ -11,3 +11,5 @@
 #include "tomcrypt.h"
 
 ltc_math_descriptor ltc_mp;
+
+int testInt = 500;
```

^Added testInt to the .c and the header.

Then I made an example program:

```c
#include <tomcrypt.h>

int main(int argc, char** argv)
{
    ltc_mp = ltm_desc;  
    printf("%d\n", testInt);

    printf("%s\n", ltc_mp.name);
    return 0;
}
```

And this works just fine, links perfectly and prints: 
```
500
LibTomMath
```
As you might expect.

But, if I comment out the first printf, the one using testInt, it fails to link.
So it seems like the linker is ignoring that file unless there is another symbol in there that's being used besides ltc_mp.

So, yeah, how should I proceed on this? Am I misunderstanding some subtlety of the osx linker?

ruslo:
> oooook, so (hesistant to say this but), this looks like it might a linker bug?

Might be. Try to create [SSCCE](http://www.sscce.org/) and if linker will fail in this case you can open bugreport.

> Am I misunderstanding some subtlety of the osx linker?

I don't think so. Code looks good, error is weird.

> So, yeah, how should I proceed on this?

Up to you. You can exclude OSX from testing or if adding dummy `extern int` will fix the issue you can add this as a workaround.

wheybags:
Minimal example here: https://gist.github.com/wheybags/0cf9abe991d1c667e5218bf59ab24649
I'm using yosemite, if you have access to the newest version of osx/xcode, would you mind verifying this?

ruslo:
> Minimal example here

Makefile, really? :)

> I'm using yosemite, if you have access to the newest version of osx/xcode, would you mind verifying this?

Latest Xcode 9 available on Travis.

GitHub + Travis:
* https://github.com/forexample/osx-extern-struct-linker-error

wheybags:
Hey, it was either that or a shell script :p
Ok, I'll send a bug report to apple I guess, and in the meantime I'll add
some nasty hacky workarounds to the libtomcrypt and stormlib packages
Thanks!

wheybags:
Ok, so updated with a workaround someone pointed out to me on IRC (forgot your name workaround guy, sorry :( )

cmake/configs/default.cmake
```diff
@@ -108,11 +108,12 @@ hunter_config(SDL2 VERSION 2.0.4-p4)
 hunter_config(SDL_mixer VERSION 2.0.1-p1)
 hunter_config(SQLite3 VERSION autoconf-3080803) #R-Tree enabled
 hunter_config(Sober VERSION 0.1.3)
+hunter_config(stormlib VERSION 9.21-p1)
```
ruslo: Please split these changes.

wheybags:
done: https://github.com/ruslo/hunter/pull/971

ruslo:
> done: #971

Merged, please rebase.

wheybags:
done

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2202
* https://travis-ci.org/ingenue/hunter/builds/266315803?utm_source=github_status&utm_medium=notification

ruslo:
> Testing

Some toolchains failed, feel free to exclude them from testing.

wheybags:
disabled analyze: https://github.com/ingenue/hunter/pull/113
Also, updated, should fix MSYS2 build

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2214
* https://travis-ci.org/ingenue/hunter/builds/266526369?utm_source=github_status&utm_medium=notification

ruslo:
> should fix MSYS2 build

Still failed:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2214/job/y6pw3biw4eymeesl

wheybags:
Ok, updated again, should work now.

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2239
* https://travis-ci.org/ingenue/hunter/builds/268638207?utm_source=github_status&utm_medium=notification

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2241
* https://travis-ci.org/ingenue/hunter/builds/268642116?utm_source=github_status&utm_medium=notification

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.19.83

PR #890, opened 2017-07-20
Fix reference to wiki in hunter_fatal_error

nmoehrle:
Calls to `hunter_fatal_error` cause a `hunter_internal_error` due to the incorrect reference of [HUNTER_WIKI](https://github.com/ruslo/hunter/blob/master/cmake/modules/hunter_wiki.cmake#L5) as hunter_WIKI.


cmake/modules/hunter_fatal_error.cmake
```diff
@@ -8,7 +8,7 @@ include(hunter_wiki)
 
 function(hunter_fatal_error)
   cmake_parse_arguments(hunter "" "WIKI" "" "${ARGV}")
-  if(NOT hunter_WIKI)
```
ruslo: No, this one is correct. I guess the problem is in function calling `hunter_fatal_error` without `WIKI` argument. Like `hunter_fatal_error("Some message")`, should be `hunter_fatal_error("Some message" WIKI error.something)`.

ruslo: It make sense to use `x_` here instead of `hunter_`:
```cmake
cmake_parse_arguments(x "" "WIKI" "" "${ARGV}")
if(NOT x_WIKI)
  hunter_internal_error("Expected wiki")
endif()
```

nmoehrle:
Sorry I misunderstood the usage and came to the wrong conclusion. Thanks for the clarification :-)

nmoehrle:
Unfortunately there are quite a few projects which use this function incorrectly, searching for `hunter_fatal_error` in this repository reveals many of those, and if used incorrectly the projects error message get swallowed by the internal error.

Do you think it would be an option to either raise a `hunter_fatal_error` referencing the incorrect usage of `hunter_fatal_error` or linking to a default wiki page?

cmake/modules/hunter_fatal_error.cmake
```diff
@@ -7,15 +7,15 @@ include(hunter_internal_error)
...
 function(hunter_fatal_error)
-  cmake_parse_arguments(hunter "" "WIKI" "" "${ARGV}")
-  if(NOT hunter_WIKI)
-    hunter_internal_error("Expected wiki")
+  cmake_parse_arguments(x "" "WIKI" "" "${ARGV}")
+  if(NOT x_WIKI)
+    hunter_internal_error("Expected argument WIKI")
   endif()
   message("")
   foreach(x ${hunter_UNPARSED_ARGUMENTS})
```
ruslo: This one need to changed too.

ruslo:
> Do you think it would be an option to either raise a hunter_fatal_error referencing the incorrect usage of hunter_fatal_error or linking to a default wiki page?

`hunter_internal_error` can be used in this case instead

NeroBurner:
I think this PR got lost in time. Is it still relevant?

ruslo:
Actually it looks good. I guess it was lost because I haven't received notification about new commit and nobody sent me a message explicitly.

PR #921, opened 2017-07-31
ceres: add OpenMP to example, needed by MinGW

NeroBurner:
MinGW-w64 cross compiling needs OpenMP to successfully link the example `ceres-solver-suitesparse`. `g++` links fine without.

I don't know if it is better to just update the examples to search for OpenMP when compiling with mingw, or update the `ceres-solver` cmake config file to instruct it to always search for OpenMP (if ceres was built with OpenMP)


ruslo:
> or update the ceres-solver cmake config file to instruct it to always search for OpenMP (if ceres was built with OpenMP)

We already using fork, so I think this option is better.

NeroBurner:
should be fixed with https://github.com/ruslo/hunter/pull/993

PR #937, opened 2017-08-07
add vim modeline to all modules

NeroBurner:
adds modeline to all hunter modules with tabstopwith information (make tabs 2 spaces wide and expand tabulators to spaces)

### DIFF
--- a/cmake/modules/hunter_add_package.cmake
+++ b/cmake/modules/hunter_add_package.cmake
@@ -54,3 +54,5 @@ macro(hunter_add_package)
     hunter_status_debug("load: ${x} ... end")
   endforeach()
 endmacro()
+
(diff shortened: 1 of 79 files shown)


ruslo:
I don't think editor specific information should be saved in file. What is the point of this? Why you can't just save this settings globally if you want? (I'm a Vim user by the way)

NeroBurner:
you could save it locally, but then every txt file would have tabwidth 2 with expandtab. With the modeline could work on two projects with CMakefiles, which have different style guidelines

ruslo:
> With the modeline could work on two projects with CMakefiles, which have different style guidelines

Still something that should be done globally on user side, not per CMake file in the world.

NeroBurner:
closing, because not wanted

PR #942, opened 2017-08-07
Add cotire package

chfast:
I'd like to add cotire CMake module as a Hunter package. Their releases consist of unpacked cotire.cmake module only. Using `url_sha1_download` works ok, but I'd like to file to be move to any common path for CMake modules, e.g. `Install/cmake`.

### DIFF
--- a/examples/cotire/CMakeLists.txt
+++ b/examples/cotire/CMakeLists.txt
@@ -1,0 +1,17 @@
+# Copyright (c) 2017, Pawel Bylica
+# All rights reserved.
+
+cmake_minimum_required(VERSION 3.0)
+
+# Emulate HunterGate:
+# * https://github.com/hunter-packages/gate
+include("../common.cmake")
+
+project(download-cotire)
+
+# download cotire
+hunter_add_package(cotire)
+message(${COTIRE_ROOT})
+
+# This works, but is quite ugly. Can we get include(cotire) to work?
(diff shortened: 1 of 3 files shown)


examples/cotire/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+# * https://github.com/hunter-packages/gate
+include("../common.cmake")
+
+project(download-cotire)
+
+# download cotire
+hunter_add_package(cotire)
+message(${COTIRE_ROOT})
+
+# This works, but is quite ugly. Can we get include(cotire) to work?
```
ruslo: You can create `cotireConfig.cmake`. Take a look at Gauze:
* https://github.com/hunter-packages/gauze/blob/master/cmake/Config.cmake.in

chfast: I don't think creating a hunterized fork of this project make sense, unless they are going to pick it up upstream.

Would this work if the cotire had a simple CMakeLists.txt with install() that installs the cotire.cmake module?

ruslo: > I don't think creating a hunterized fork of this project make sense

Why not?

> Would this work if the cotire had a simple CMakeLists.txt with install() that installs the cotire.cmake module?

Yes, that's what I mean. Plus add generation of `cotireConfig.cmake` and we have a nice package :)

ruslo:

Issue #918, opened 2017-07-28
LLVM on VS and NMake


dvirtz:
Trying to build LLVM on Visual Studio gives the error
```
Broken package:  LLVM build broken on Visual Studio, see: 
http://llvm.1065342.n5.nabble.com/Install-Problem-of-Compiler-rt-with-Visual-Studio-Express-2013-for-Windows-Desktop-td66575.html  
(use MINGW or NMake generators)
```

1. This issue is quite old, wasn't it fixed already?
2. Trying to build with NMake I'm getting
```
[ 60%] Linking CXX static library ..\..\..\..\lib\clang\3.6.2\lib\windows\clang_rt.asan-i386.lib                                                                          
LINK : fatal error LNK1181: cannot open input file '../sanitizer_common/CMakeFiles/RTSanitizerCommonLibc.i386.dir/sanitizer_coverage_mapping_libcdep.cc.obj'              
```
I believe it's due to the filenames being longer than 260. Is there any way to make Hunter use shorter paths?

ruslo:
> I believe it's due to the filenames being longer than 260. Is there any way to make Hunter use shorter paths?

Discussed few times already. I have no good workaround/fix for it. The only way is to minimize `HUNTER_ROOT` path, e.g. try `C:\Hunter`. Note that it's neither Hunter nor CMake issue, it's a limitation of build tools on Windows. CMake is not a build system, it's a generator for real build system, not much we can do.

dvirtz:
Well, after setting HUNTER_BINARY_DIR to avoid the long filenames, I get hit by [this](https://bugs.llvm.org/show_bug.cgi?id=25650) Clang's bug.

Can I get an access to the LLVM packages to see if upgrading the versions can help?

ruslo:
> Well, after setting HUNTER_BINARY_DIR to avoid the long filenames

Oh, yeah, there was one workaround. I haven't check that it's working, may behave incorrectly in some cases, be aware :)

> Can I get an access to the LLVM packages to see if upgrading the versions can help?

You can check it locally/fork, why do you need access? :)

dvirtz:
@ruslo can you make a hunter fork of Compiler RT?
I need it in order to fix the installation when using VS.
I also need new branches for 4.0.1 on https://github.com/hunter-packages/llvm and https://github.com/hunter-packages/clang

ruslo:
> @ruslo can you make a hunter fork of Compiler RT?

[...]

PR #943, opened 2017-08-08
add LLVM 4.0.1

dvirtz:


dvirtz:
resolves https://github.com/ruslo/hunter/issues/918.

I didn't update Libcxx as it isn't used on any example and I don't have a way to test it

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2115
* https://travis-ci.org/ingenue/hunter/builds/262126222?utm_source=github_status&utm_medium=notification

ruslo:
```
CMake Error at cmake/modules/CheckCompilerVersion.cmake:45 (message):
  Host Visual Studio must be at least 2015
```

```
CMake Error at cmake/modules/AddLLVM.cmake:1340 (string):
  string begin index: -1 is out of range 0 - 23
```

Please exclude all toolchain that failed not by timeout. For timeout I will upload binaries from local machine.

dvirtz:
I don't see any timeout, do you?

```
CMake Error at cmake/modules/AddLLVM.cmake:1340 (string):
  string begin index: -1 is out of range 0 - 23
```

This happens because Hunter passes `-DCMAKE_CONFIGURATION_TYPES=${configuration} ` even on single configuration generators and LLVM expects it being set only on multi-configuration generators.
What do you think about changing the CMake scheme to do
```cmake
  if(CMAKE_CONFIGURATION_TYPES)
    set(build_type_opts "-DCMAKE_CONFIGURATION_TYPES=${configuration}")
  else()
    set(build_type_opts "-DCMAKE_BUILD_TYPE=${configuration}")
  endif()
```
and then passing only `build_type_opts` to CMake?

ruslo:
> I don't see any timeout, do you?

I saw that `osx-10-11` takes 48 minutes, but now I'm not sure about reason of failure.

> What do you think about changing the CMake scheme to do

We can use only `CMAKE_CONFIGURATION_TYPES` on multi and only `CMAKE_BUILD_TYPE` on single. But check [should be](http://cgold.readthedocs.io/en/latest/tutorials/build-types.html#detect-multi-single):
```cmake
string(COMPARE EQUAL "${CMAKE_CFG_INTDIR}" "." is_single)
if(is_single)
  set(build_type_opts "-DCMAKE_BUILD_TYPE=${configuration}")
else()
  set(build_type_opts "-DCMAKE_CONFIGURATION_TYPES=${configuration}")
endif()
```

> LLVM expects it being set only on multi-configuration generators

LLVM should not expect this. It's a separate issue and should be fixed in LLVM.

ruslo:
> set build type differently on single and multi configuration generators

Please open separate pull request with this change. Want to run few tests.

dvirtz:
Done. https://github.com/ruslo/hunter/pull/948

ruslo:
> Done. #948

Merged, thanks. Please rebase this one.

cmake/projects/PocoCpp/hunter.cmake
```diff
@@ -8,6 +8,17 @@ hunter_add_version(
     PACKAGE_NAME
     PocoCpp
     VERSION
+    1.7.8-p0
+    URL
+    "https://github.com/hunter-packages/poco/archive/v1.7.8-p0.zip"
```
ruslo: rebase + squash?

dvirtz:
sorry, it always takes me some tries to make it right :-)

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2126
* https://travis-ci.org/ingenue/hunter/builds/262791336?utm_source=github_status&utm_medium=notification

ruslo:
Tests failed. I see timeout, the rest should be removed/fixed.

ruslo:
New toolchain added, retest:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2131
* https://travis-ci.org/ingenue/hunter/builds/262829881?utm_source=github_status&utm_medium=notification

dvirtz:
OK, VS 2015 now fails because of long paths. Can I modify AppVeyor receipe to use `HUNTER_BINARY_DIR`?

ruslo:
> OK, VS 2015 now fails because of long paths. Can I modify AppVeyor receipe to use HUNTER_BINARY_DIR?

No need, I want to upload it from local machine anyway.

ruslo:
Toolchain `osx-10-11` failed on my local machine with error:
```
=== BUILD TARGET clang_rt.builtins_x86_64h_osx OF PROJECT LLVM WITH CONFIGURATION Release ===

Check dependencies
No architectures to compile for (ARCHS=x86_64h, VALID_ARCHS=i386 x86_64).

** BUILD FAILED **
```

Don't ask me what it means, I have no idea :) I guess should be excluded too.

Working so far:
* `vs-14-2015`: https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2135
* `libcxx`: https://travis-ci.org/ingenue/hunter/jobs/262829885

ruslo:
> I guess should be excluded

`default` is not very interesting because it's quite the same as `libcxx` so I vote for excluding too. It will save some time on local upload. Same for Linux `default` ~ `gcc`.

ruslo:
`gcc` is working now:
* https://travis-ci.org/ingenue/hunter/jobs/262829883

We have `vs-14-2015` on Windows, `libcxx` on OSX and `gcc` on Linux. Please exclude the rest.

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2144
* https://travis-ci.org/ingenue/hunter/builds/263496307?utm_source=github_status&utm_medium=notification

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.19.68

PR #952, opened 2017-08-10
fix pack_git_submodule with subdir

NeroBurner:
- `git archive HEAD:subdir -o archive.tar.gz` creates a different
  archive (different SHA1) each time
- calling `git archive HEAD -o archive.tar.gz` with the right working
  directory produces the same archive (same SHA1) each time


ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.19.65

PR #954, opened 2017-08-11
Add OpenCV 3.3.0

xsacha:


ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2138
* https://travis-ci.org/ingenue/hunter/builds/263388581?utm_source=github_status&utm_medium=notification

ruslo:
iOS build failed. Log on Travis is not full, I confirmed fail locally.
Error:
```
CMake Error at cmake/OpenCVCompilerOptimizations.cmake:399 (message):
  Required baseline optimization is not supported: NEON
  (CPU_BASELINE_REQUIRE=;NEON)
```

If you don't know how to fix it you can use this code in `configs/default.cmake`:
```cmake
if(IOS)
  hunter_config(OpenCV VERSION /* old version */)
else()
  hunter_config(OpenCV VERSION /* new version */)
endif()
```

xsacha:
I want iOS to work because I target it.
Isn't iPhone 5 the minimum you can target nowadays anyway? iPhone 3GS and iPhone 4 had NEON.
Does the toolchain just need fixing?

ruslo:
I don't know the reason of failure.

xsacha:
Looks like toolchain does not support NEON. Not sure why it is not the default but what if you add -mfpu=neon for armv7?

Surely this is an issue in all packages if they are being compiled without NEON support this whole time even though all armv7 iPhone support it?

ruslo:
> cmake/OpenCVCompilerOptimizations.cmake:399

Can you explain what is happening here and how this support is tested?

xsacha:
Looks like it uses these files to run the test:
"${OpenCV_SOURCE_DIR}/cmake/checks/cpu_neon.cpp"
"${OpenCV_SOURCE_DIR}/cmake/checks/cpu_fp16.cpp"

With these relevant vars:
    ocv_update(CPU_NEON_FLAGS_ON "-mfpu=neon")
    ocv_update(CPU_NEON_FLAGS_CONFLICT "-mfpu=[^ ]*")
    ocv_update(CPU_FP16_FLAGS_ON "-mfpu=neon-fp16")
    ocv_update(CPU_FP16_IMPLIES "NEON")
    ocv_update(CPU_FP16_FLAGS_CONFLICT "-mfpu=[^ ]*")

For some reason NEON does not make it in to the final baseline optimisations so it complains.

ruslo:
> Looks like it uses these files to run the test

No, it's not about these tests.

We are not entering this block:
* https://github.com/hunter-packages/opencv/blob/581d74d5acccc1b64bdb2f962101d1f38cf54a71/cmake/OpenCVCompilerOptimizations.cmake#L264-L267

Because `ARM` is not set:
* https://github.com/hunter-packages/opencv/blob/581d74d5acccc1b64bdb2f962101d1f38cf54a71/cmake/OpenCVCompilerOptimizations.cmake#L252

`ARM` is empty because we are not entering this code:
* https://github.com/hunter-packages/opencv/blob/581d74d5acccc1b64bdb2f962101d1f38cf54a71/cmake/OpenCVDetectCXXCompiler.cmake#L120

Because `CMAKE_SYSTEM_PROCESSOR` was not set by toolchain. `CMAKE_SYSTEM_PROCESSOR` is not set because iOS toolchain can be used for multiarchitecture build (e.g. armv7 + i386).

To fix this problem both OpenCV and Polly should be improved:
* https://github.com/ruslo/polly/commit/b6d9289043a0f2ec60b8b27a1aa229b96a349979
* https://github.com/hunter-packages/opencv/commit/8646be90f15f3e0b146132398e2b9c84f34747e8

Please use this OpenCV release:
* https://github.com/hunter-packages/opencv/releases/tag/v3.3.0-p1

Unfortunatelly it means that since OpenCV 3.3 it will be not possible to build multiarch iOS libraries.

xsacha:
That isn't very useful without multiarch. Since it was working in the previous build it should be possible to get multiarch again. I suppose it is due to the new baseline optimisations. Should it be an upstream bug?

ruslo:
> Should it be an upstream bug?

They're using local custom iOS toolchain and Python script to create multiarch release. I doubt they will fix anything.

> Since it was working in the previous build it should be possible to get multiarch again. I suppose it is due to the new baseline optimisations.

I have no idea what is changed and if it's possible to fix it. Probably it's something about using `ENABLE_NEON=ON` which is default in Hunter (?)

ruslo:
Conflict.

cmake/configs/default.cmake
```diff
@@ -85,7 +85,11 @@ hunter_config(NASM VERSION 2.12.02)
 hunter_config(OpenBLAS VERSION 0.2.20-p0)
 hunter_config(OpenCL VERSION 2.1-p3)
 hunter_config(OpenCL-cpp VERSION 2.0.10-p0)
-hunter_config(OpenCV VERSION 3.2.0-p2)
+if(IOS)
+  hunter_config(OpenCV VERSION 3.2.0-p2)
+else()
+  hunter_config(OpenCV VERSION 3.3.0-p0)
```
ruslo: Is OpenCV-Extra 3.2 compatible with OpenCV 3.3? I see 3.3.0 here:
* https://github.com/ruslo/hunter/blob/master/cmake/projects/OpenCV-Extra/hunter.cmake

Probably they should be updated simultaneously (?)

xsacha: I guess I need the same if check for iOS on extras?

xsacha:
I guess so. I'm having new issues compiling other projects for iOS as well (DLib stopped compiling).

cmake/configs/default.cmake
ruslo: Yes, that's what I meant.

xsacha:
@ruslo 
Any problems with this one? I haven't figured out how to fix iOS yet.

ruslo:
OpenCV:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2208
* https://travis-ci.org/ingenue/hunter/builds/266469907?utm_source=github_status&utm_medium=notification

OpenCV-extra:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2209
* https://travis-ci.org/ingenue/hunter/builds/266470014?utm_source=github_status&utm_medium=notification

xsacha:
One of the OpenCV-extra builds went over 4MB log at 99% built.

ruslo:
Uploading.

OpenCV:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2211
* https://travis-ci.org/ingenue/hunter/builds/266525264?utm_source=github_status&utm_medium=notification

OpenCV-extra:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2212
* https://travis-ci.org/ingenue/hunter/builds/266525295?utm_source=github_status&utm_medium=notification

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.19.76

PR #995, opened 2017-08-30
doc: fix spelling errors

NeroBurner:


docs/packages/pkg/OpenCV.rst
```diff
@@ -1,6 +1,8 @@
 .. spelling::
 
     OpenCV
+    GStreamer
+    backend
```
ruslo: `backend` can be removed from here. In text `back-end` should be used:
* https://en.oxforddictionaries.com/definition/us/back_end

NeroBurner:
fixed back-end

PR #981, opened 2017-08-24
[WIP] Doc add packages

NeroBurner:
move some package descriptions from wiki-pages to sphix documentation

PR #997, opened 2017-08-31
docs: add template for pkg stormlib and SDL_mixer

NeroBurner:


wheybags:
Do these docs generate some page? I didn't know they existed or I would have added them

ruslo:
@wheybags was added recently: https://github.com/ruslo/hunter/pull/981 (moved from wiki to be precise).

NeroBurner:
to create files with templates for packages with missing documentation file use the following script in the folder `maintenance`

```
python3 docs_create_missing_stubs.py
```

this creates the files `maintenance/packages/only_hunter/<pkg_name>.rst`

You can add these files then to `docs/packages/pkg/` folder and update the `__FIXME__` tags

PR #999, opened 2017-08-31
ceres-solver: add example dynamic LAPACK

NeroBurner:
this PR adds an example testing ceres-solver built with SuiteSparse using a dynamic LAPACK lib. I'm using this configuration in all my projects. It would be nice that this configuration is tested and maybe even cached

should we maybe merge the ceres-tests into one testing branch `pkg.ceres-solver` instead of the two we have now?

### DIFF
--- a/examples/ceres-solver-suitesparse-dynLAPACK/CMakeLists.txt
+++ b/examples/ceres-solver-suitesparse-dynLAPACK/CMakeLists.txt
@@ -1,0 +1,29 @@
+# Copyright (c) 2015, Ruslan Baratov
+# All rights reserved.
+
+cmake_minimum_required(VERSION 3.0)
+
+set(TESTING_CONFIG_OPT FILEPATH "${CMAKE_CURRENT_LIST_DIR}/config.cmake")
+
+# Emulate HunterGate:
+# * https://github.com/hunter-packages/gate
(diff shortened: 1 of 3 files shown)


examples/ceres-solver-suitesparse-dynLAPACK/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+cmake_minimum_required(VERSION 3.0)
+
+set(TESTING_CONFIG_OPT FILEPATH "${CMAKE_CURRENT_LIST_DIR}/config.cmake")
+
+# Emulate HunterGate:
+# * https://github.com/hunter-packages/gate
+include("../common.cmake")
+
+project(download-ceres-solver)
+# build LAPACK as shared library
```
ruslo: I don't think these comments are needed. You can see from `TESTING_CONFIG_OPT` that we are loading custom `config.cmake`, and in `config.cmake` it's clear what is happenning.

ruslo:
`ceres-solver-suitesparse` testing merged into `pkg.ceres-solver` branch:
* https://github.com/ingenue/hunter/blob/pkg.ceres-solver/.travis.yml

Feel free to add `ceres-solver-dynLAPACK` example.

PR #1008, opened 2017-09-01
add examples for dynamic LAPACK and SuiteSparse

NeroBurner:
- build LAPACK as dynamic library
- use dynamic LAPACK library in SuiteSparse-dynLAPACK


ruslo:
Branches `pkg.lapack` and `pkg.suitesparse` updated.

PR #1013, opened 2017-09-03
small fix for SDL package

wheybags:
https://github.com/hunter-packages/SDL2/commit/c0ef6bfa5a16618efe5e58a1d89297094fc33c04

### DIFF
--- a/cmake/configs/default.cmake
+++ b/cmake/configs/default.cmake
@@ -120,7 +120,7 @@ hunter_config(QtCMakeExtra VERSION 1.0.27)
 hunter_config(QtQmlManager VERSION 1.0.0)
 hunter_config(RapidJSON VERSION 1.0.2-p2)
 hunter_config(RapidXML VERSION 1.13)
-hunter_config(SDL2 VERSION 2.0.4-p4)
+hunter_config(SDL2 VERSION 2.0.4-p5)
 hunter_config(SDL_image VERSION 2.0.1-p1)
 hunter_config(SDL_mixer VERSION 2.0.1-p1)
 hunter_config(SQLite3 VERSION autoconf-3080803) #R-Tree enabled
--- a/cmake/projects/SDL2/hunter.cmake
+++ b/cmake/projects/SDL2/hunter.cmake
@@ -10,6 +10,17 @@ hunter_add_version(
         PACKAGE_NAME
         SDL2
         VERSION
+        "2.0.4-p5"
+        URL
+        "https://github.com/hunter-packages/SDL2/archive/2.0.4-p5.tar.gz"
+        SHA1
+        14dcb4c439f0dbd71ad0b92184b35e3a7f29e3ca
+)
+
+hunter_add_version(
+        PACKAGE_NAME
+        SDL2
+        VERSION
         "2.0.4-p4"
         URL
         "https://github.com/hunter-packages/SDL2/archive/2.0.4-p4.tar.gz"


ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2324
* https://travis-ci.org/ingenue/hunter/builds/271637459?utm_source=github_status&utm_medium=notification

ruslo:
`vs-11-2012` not working

wheybags:
Ok, sorry for the wait, tbh I forgot about this issue :p
Pushed a fix, should work now. Side note: why do big projects like SDL have such terrible cmake files? :(

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2590
* https://travis-ci.org/ingenue/hunter/builds/290678980?utm_source=github_status&utm_medium=notification

ruslo:
Windows not working.
Here is instructions how you can test it yourself on AppVeyor:
* https://docs.hunter.sh/en/latest/creating-new/create/cmake.html#ci-testing

wheybags:
ok, fixed and build passed:
https://travis-ci.org/wheybags/hunter/builds/291184519
https://ci.appveyor.com/project/wheybags/hunter

Super nice that I can run CI builds myself now, that'll save a lot of time.

The issue was that the cmake scripts for SDL were using the wrong SDL_config.h, using a generated one instead of the specific hardcoded windows version (why? God knows?). When I changed it to use the right config, it was picking up on that header enabling the xaudio2 audio backend, but the lib file for xaudio2 is in the dx part of the windows SDK, which it wasn't finding or linking to or dealing with in any way basically. For now, I just commented out the line enabling xaudio2, but we probably need to find a better solution. Could hunter ship a cmake module for finding the windows sdk? and how do we pick a version? AFAIK you can specify a windows api target version, can we extrapolate it from that? And do people actually do that in practice (I know I don't).

ruslo:
> and how do we pick a version?

Belongs to toolchain:
* https://github.com/ruslo/polly/blob/a9d58114d3c8bfffe33128e6bc189d6a21bc3502/vs-14-2015-win64-sdk-8-1.cmake#L19

>  the lib file for xaudio2 is in the dx part of the windows SDK

If this library is part of SDK and should **always** exist, then you can create helper module to find it, say similar to `avfoundation` framework on OSX:
* https://docs.hunter.sh/en/latest/packages/pkg/avfoundation.html

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2605
* https://travis-ci.org/ingenue/hunter/builds/291202552?utm_source=github_status&utm_medium=notification

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2608
* https://travis-ci.org/ingenue/hunter/builds/291206624?utm_source=github_status&utm_medium=notification

wheybags:
Thanks.

As for
> Belongs to toolchain:

What if you don't specify a toolchain? I normally don't when building my projects, just let it pick up on my compiler automatically. Does cmake have auto-generated "toolchains" for that, and would they set an sdk version?

ruslo:
> What if you don't specify a toolchain?

It will use defaults.

> Does cmake have auto-generated "toolchains" for that, and would they set an sdk version?

You can set custom default toolchain in your project:
* https://github.com/ethereum-mining/ethminer/blob/e27a96284645af2b5cd7d3e8e36bab34f1a6ab3f/CMakeLists.txt#L4

Currently toolchain is an only way to customize compiiler/flags for 3rd parties:
* https://docs.hunter.sh/en/latest/overview/customization/toolchain-id.html

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.19.131

Issue #838, opened 2017-06-16
Add package `folly`


ssbanerje:
https://github.com/facebook/folly

Dependencies and build based on CMake. Would be nice to have in hunter

isaachier:
CMake build currently only supported on Windows. Not sure why.

tdauth:
For Linux, I am using a combination of ExternalProject and Bash scripts in my project to build Folly: https://github.com/tdauth/cpp-futures-promises.

Wangle has CMake support and could be added to Hunter: https://github.com/facebook/wangle

isaachier:
I was working on a CMake build for folly but I saw a lot of the autotools stuff was not tested for in the existing CMake build, so I felt it would be too much work.

OvermindDL1:
Isn't `folly` entirely cmake based no, no more autotools or anything?
https://github.com/facebook/folly

isaachier:
@OvermindDL1 thanks for the update. It seems you are correct. Maybe I'll patch it for Hunter after all.

OvermindDL1:
For note, as documented at https://github.com/facebook/folly/issues/734#issuecomment-365728656 the required dependencies is:  `Folly depends on zlib, openssl, libevent, double-conversion, glog, gflags, lz4, liblzma, snappy and boost`

isaachier:
Ya I packaged the dependencies months ago. The only remaining task was folly itself.

OvermindDL1:
From same thread, just adding this for completion:  `Folly also depends on zstd.`

PR #1017, opened 2017-09-04
Add double-conversion package.

isaachier:
One of dependencies for Folly (see #838).


examples/double-conversion/main.cpp
```diff
@@ -0,0 +1,6 @@
+#include <double-conversion/double-conversion.h>
```
ruslo: Looks like `*Config.cmake` is not correct:
```
fatal error: double-conversion/double-conversion.h: No such file or directory
```

isaachier: Not sure what you mean. It works for me using this command:

`TOOLCHAIN=default PROJECT_DIR=examples/double-conversion ./jenkins.py`

ruslo: > It works for me using this command

I can reproduce error on Travis:
* https://travis-ci.org/ingenue/hunter/builds/271825246?utm_source=github_status&utm_medium=notification

isaachier: Then why does only one build fail [here](https://travis-ci.org/ruslo/hunter/builds/271757246)?

isaachier: That build failure I referenced is actually a documentation error. Fixed in the change I just pushed.

ruslo: > Then why does only one build fail here?

It's different testing - only unit-tests and documentation, `example/double-conversion` is not tested.

isaachier: You are right. I had double-conversion installed locally, masking the error. Thanks. Will fix soon.

isaachier:
By using `include_directories(${double-conversion_INCLUDE_DIRS})`, I was able to fix compilation. However, I'm not sure based on the project's CMake files why this include directory isn't propagated automatically to the dependent project.

examples/double-conversion/main.cpp
```diff
@@ -0,0 +1,6 @@
+#include <double-conversion.h>
```
ruslo: It should be `#include <double-conversion/double-conversion.h>`. If you will use `-I<root>/include/double-conversion` then you will pollute includes with internal headers. I.e. when user will do `#include <utils.h>` it will include `<root>/include/double-conversion/utils.h`.

examples/double-conversion/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+
+include("../common.cmake")
+
+project(download-double-conversion)
+
+hunter_add_package(double-conversion)
+
+# now double-conversion can be used
+find_package(double-conversion CONFIG REQUIRED)
+include_directories(${double-conversion_INCLUDE_DIRS})
```
ruslo: It should be part of imported target.

docs/packages/pkg/double-conversion.rst
```diff
@@ -0,0 +1,21 @@
...
+
+-  `Official <https://github.com/google/double-conversion>`__
+-  `Example <https://github.com/ruslo/hunter/blob/master/examples/double-conversion/CMakeLists.txt>`__
+-  Added by `Isaac Hier <https://github.com/isaachier>`__ (`pr-1017 <https://github.com/ruslo/hunter/pull/1017>`__)
+
+.. code-block:: cmake
+
+    hunter_add_package(double-conversion)
+    find_package(double-conversion CONFIG REQUIRED)
+    target_link_libraries(double-conversion)
```
ruslo: Not a valid CMake code. Use `target_link_libraries(foo double-conversion)` or `target_link_libraries(... double-conversion)`.

cmake/projects/double-conversion/hunter.cmake
```diff
@@ -0,0 +1,22 @@
...
+    double-conversion
+    VERSION
+    "3.0.0"
+    URL
+    "https://github.com/google/double-conversion/archive/v3.0.0.tar.gz"
+    SHA1
+    d900d4d946beb493a03cd6c9b180ed2bcbd0f20e
+)
+
+hunter_configuration_types(double-conversion CONFIGURATION_TYPES Release)
```
ruslo: What is the reason of this one? Default configuration is Debug + Release, user can set his own configuration in local config (see https://github.com/ruslo/hunter/wiki/example.hunter_configuration_types).

examples/double-conversion/main.cpp
isaachier: That makes sense. I was actually basing it on the generated Cmake config. But this is something I can change in a fork.

examples/double-conversion/CMakeLists.txt
isaachier: Ya I was just double checking it wasn't ideal. Will keep researching.

docs/packages/pkg/double-conversion.rst
isaachier: Good catch

cmake/projects/double-conversion/hunter.cmake
isaachier: I saw some other libraries do this. Thought it was a standard approach for low level libraries (like this one).

isaachier: I saw some other libraries do this. Thought it was a standard approach for low level libraries (like this one).

ruslo: > I saw some other libraries do this

They do it for other reasons. Usually because of the bugs and limitations.

examples/double-conversion/main.cpp
ruslo: > But this is something I can change in a fork.

I've created https://github.com/hunter-packages/double-conversion and added you to the team (check your email). You can push directly but it's better if you create pull request for review reasons. I've created branch `hunter-v3.0.0` from upstream tag `v3.0.0`.

isaachier: Thanks. I actually looked at the source and the headers themselves seem to include other local headers without using the directory path. For example, see https://github.com/google/double-conversion/blob/master/double-conversion/bignum.h#L31. I may have to fix these inclusions to fix the include directive.

isaachier:
Got it. Thanks!

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2357
* https://travis-ci.org/ingenue/hunter/builds/274530379?utm_source=github_status&utm_medium=notification

cmake/configs/default.cmake
```diff
@@ -82,6 +82,7 @@ hunter_config(Leathers VERSION 0.1.6)
 hunter_config(Leptonica VERSION 1.74.2-p4)
 hunter_config(Libcxx VERSION 3.6.2) # Clang
 hunter_config(Libcxxabi VERSION 3.6.2) # Clang
+hunter_config(Libevent VERSION 2.1.8)
```
ruslo: Doesn't belong here.

isaachier: Sorry my mistake.

isaachier:
I have an easy fix for clang-analyzer, but it won't work with `-DNDEBUG`. clang-analyzer checks assertions to understand valid scenarios, but `-DNDEBUG` causes `#define assert(x)  // do nothing`.

isaachier:
Even with the recent fix to double-conversion, the analyzer bug is still present due to the `-DNDEBUG` setting. See this quote from clang-analyzer docs for details:

> ALWAYS analyze a project in its "debug" configuration
> Most projects can be built in a "debug" mode that enables assertions. Assertions are picked up by the
> static analyzer to prune infeasible paths, which in some cases can greatly reduce the number of false
> positives (bogus error reports) emitted by the tool.
>
> Another option is to use --force-analyze-debug-code flag of scan-build tool which would enable
> assertions automatically.

https://clang-analyzer.llvm.org/scan-build.html

ruslo:
> the analyzer bug is still present due to the -DNDEBUG setting

If you don't know how to fix it or don't want to fix it then just exclude it. There is no need to make all platforms or toolchains working, it's not a requirement.

isaachier:
I mean I have a fix but it means changing the overall analyzer workflow to use debug builds instead of release builds.

ruslo:
> I mean I have a fix but it means changing the overall analyzer workflow to use debug builds instead of release builds.

Release + Debug build is default configuration for Hunter.

> https://clang-analyzer.llvm.org/scan-build.html

Polly toolchain `analyze` is not using `scan-build` tool.

I personally think that analyzer should work with both Debug and Release. The assertion problem can be solved by Clang analyzer aware assert-like macro.

Such issue is out of scope of "adding new package" discussion. Please just exclude this toolchain and open separate issue for discussion if you want to improve behaviour.

jenkins.py
```diff
@@ -111,6 +111,10 @@ def run():
     print('Skip (https://github.com/ruslo/hunter/issues/30)')
     sys.exit(0)
 
+  if (project_dir == 'examples/double-conversion') and (toolchain == 'analyze'):
```
ruslo: Please update `.travis.yml` instead. See repository https://github.com/ingenue/hunter and example of pull request https://github.com/ingenue/hunter/pull/56/files. Pick target branch `pkg.double-conversion`.

ruslo: Please remove this one.

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2383
* https://travis-ci.org/ingenue/hunter/builds/275408478?utm_source=github_status&utm_medium=notification

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2393
* https://travis-ci.org/ingenue/hunter/builds/275776683?utm_source=github_status&utm_medium=notification

ruslo:
Released https://github.com/ruslo/hunter/releases/tag/v0.19.100

PR #1020, opened 2017-09-07
docs: libuv: fix spelling errors

NeroBurner:
- fix spelling errors by making header uv_ssize_t to inline code


ruslo:
Thanks for patch, but I've already handled it :) See https://github.com/ruslo/hunter/commit/6d4733309b8151219fd6930a3ef42e8584286792.

NeroBurner:
ah, yes, that's the other way to fix it :+1:

PR #1023, opened 2017-09-08
Add a missing bracket.

f-sim:
The additions for the intel compiler included a bug.


ruslo:
Here is test instructions:
* https://github.com/ruslo/hunter/wiki/dev.hunter.root#jenkinspy

Please use it before submitting patches.

Released: https://github.com/ruslo/hunter/releases/tag/v0.19.96

PR #1040, opened 2017-09-14
Improve package docs. Add autotools docs.

isaachier:


docs/creating-new/autotools.rst
```diff
@@ -0,0 +1,64 @@
...
+
+    hunter_configuration_types(Foo CONFIGURATION_TYPES Release)
+    hunter_pick_scheme(DEFAULT url_sha1_autotools)
+    hunter_cacheable(Foo)
+    hunter_download(PACKAGE_NAME Foo)
+
+Note that the build may not be cacheable if autotools generation expands
+absolute paths. Try using ``hunter_cacheable`` and see if it works.
+
+**Important**: this build will not provide any CMake config. Therefore, projects
```
ruslo: Usually autotools projects generate `*.pc` files which can be used to create imported targets.
For example `x11` target:
* creates `x11-xcb.pc` and `x11.pc` meaning we can read those files to create `x11Config.cmake`/`x11-xcbConfig.cmake` files with imported `PkgConfig::x11`/`PkgConfig::x11-xcb` targets
* Usage example: https://github.com/ruslo/hunter/blob/0dc25c663ea30f597487912e4ed4c68eb78c7385/examples/x11/CMakeLists.txt#L12-L16
* PKGCONFIG_EXPORTED_TARGETS: https://github.com/ruslo/hunter/blob/0dc25c663ea30f597487912e4ed4c68eb78c7385/cmake/projects/x11/hunter.cmake#L35-L40
* Patch `*.pc` files with hardcoded paths: https://github.com/ruslo/hunter/blob/0dc25c663ea30f597487912e4ed4c68eb78c7385/cmake/projects/x11/hunter.cmake#L42-L50

docs/creating-new/autotools.rst
```diff
@@ -0,0 +1,99 @@
...
+.. code-block:: cmake
+
+    hunter_cmake_args(
+        foo
+        CMAKE_ARGS
+            PKGCONFIG_EXPORT_TARGETS=foo
+    )
+
+In the above example, package ``foo`` generates a file ``foo.pc`` in the
+autotools build. CMake then uses ``foo.pc`` to generate a CMake config file
```
ruslo: > CMake then uses ``foo.pc`` to generate a CMake config file

It's not CMake, it's Hunter:
* Template https://github.com/ruslo/hunter/blob/0dc25c663ea30f597487912e4ed4c68eb78c7385/scripts/pkgconfig-export-targets.cmake.in#L8-L10
* Autotools build scheme: https://github.com/ruslo/hunter/blob/0dc25c663ea30f597487912e4ed4c68eb78c7385/cmake/schemes/url_sha1_autotools.cmake.in#L62-L70
* Helper module: https://github.com/ruslo/hunter/blob/0dc25c663ea30f597487912e4ed4c68eb78c7385/cmake/modules/hunter_pkgconfig_export_target.cmake#L15-L16

docs/creating-new/autotools.rst
```diff
@@ -0,0 +1,99 @@
...
+
+.. code-block:: cmake
+
+    hunter_add_package(foo)
+    find_package(foo REQUIRED)
+    add_executable(bar ${BAR_SOURCES})
+    target_include_directories(bar PUBLIC ${FOO_INCLUDE_DIRS})
+    target_link_libraries(bar PUBLIC ${FOO_LIBRARIES})
+
+However, many autotools projects generate pkg-config files. These can be used
```
ruslo: I prefer this one appear before. With "Usually autotools-based project generate pkg-config files" first, then "If no pkg-config files generated then Findfoo.cmake should be added to cmake/find".

docs/creating-new/autotools.rst
```diff
@@ -0,0 +1,99 @@
...
+Therefore, projects using autotools **must provide a file ``Findfoo.cmake`` in
+the ``cmake/find`` directory**. Dependent projects will use code similar to the
+following:
+
+.. code-block:: cmake
+
+    hunter_add_package(foo)
+    find_package(foo REQUIRED)
+    add_executable(bar ${BAR_SOURCES})
+    target_include_directories(bar PUBLIC ${FOO_INCLUDE_DIRS})
```
ruslo: I'm okay with something like this in example but not in documentation. `Findfoo.cmake` should provide imported target instead of `FOO_{INCLUDE_DIRS,LIBRARIES}`. I.e. it should be something like `target_link_libraries(bar PUBLIC foo::foo)`.

docs/creating-new/custom.rst
```diff
@@ -4,6 +4,179 @@
...
+    -- [hunter] HUNTER_ROOT: /.../Hunter
+    -- [hunter] [ Hunter-ID: 7912489 | Config-ID: 9ec2ff8 | Toolchain-ID: c018e63 ]
+    -- [hunter] IOS_SIM_ROOT: /.../Hunter/_Base/7912489/9ec2ff8/c018e63/Install (ver.: 1.8.2)
+    ...
+    -- downloading...
+         src='https://github.com/phonegap/ios-sim/archive/1.8.2.tar.gz'
+    -- [download 100% complete]
+    ios_sim: /.../Hunter/_Base/7912489/9ec2ff8/c018e63/Install/ios-sim
+
+### Default behaviour
```
ruslo: Spell checker wants `behavior` here.

ruslo: `### Something` is not RST format, should be:
```
Default behavior
================
```

docs/creating-new/custom.rst
```diff
@@ -4,6 +4,179 @@
...
+
+    hunter_pick_scheme(DEFAULT url_sha1_ios_sim) # Use new custom scheme
+    hunter_download(PACKAGE_NAME ios_sim)
+
+.. code-block:: bash
+
+    > grep ios_sim cmake/config/default.cmake
+    hunter_config(ios_sim VERSION 1.8.2)
+
+#. Using
```
ruslo: Fix format

docs/creating-new/custom.rst
```diff
@@ -4,6 +4,179 @@
...
+        "${CMAKE_COMMAND}" -E make_directory "${CMAKE_INSTALL_PREFIX}"
+        COMMAND
+        "${CMAKE_COMMAND}" -E copy build/Release/ios-sim "${CMAKE_INSTALL_PREFIX}"
+    )
+    > cmake -H. -B_builds -DCMAKE_INSTALL_PREFIX=`pwd`/_install
+    > cmake --build _builds/
+    > ls _install/
+    ios-sim
+
+#. Add new package
```
ruslo: Fix format

docs/creating-new/custom.rst
```diff
@@ -4,6 +4,179 @@
...
+    > wget https://github.com/phonegap/ios-sim/archive/1.8.2.tar.gz
+    > openssl sha1 1.8.2.tar.gz
+    SHA1(1.8.2.tar.gz)= 4328b3c8e6b455631d52b7ce5968170c9769eb1e
+    > tar xf 1.8.2.tar.gz
+    > cd ios-sim-1.8.2/
+    > xcodebuild -target ios-sim -configuration Release
+    > ls build/Release/ios-sim
+    build/Release/ios-sim
+
+#. Test it using ExternalProject_Add
```
ruslo: Fix format

docs/creating-new/custom.rst
```diff
@@ -4,6 +4,179 @@
...
 ------------------------
 
-* How to add a new `custom-build project`_
+Non-CMake projects can be added too. But sometimes it's not a trivial task
+(for example there are a 3 custom schemes for `OpenSSL <https://github.com/ruslo/hunter/blob/master/cmake/projects/OpenSSL/hunter.cmake>`_. In general it's better to
+apply a patch to an existing CMake build and use :ref:`cmake-simple` add
+instruction. Anyway here is a guide how to add a project with custom build:
 
-.. _custom-build project: https://github.com/ruslo/hunter/wiki/usr.adding.new.package.custom.scheme
+#. Test it manually
```
ruslo: Fix format

isaachier: This is actually valid. See here: http://www.sphinx-doc.org/en/stable/rest.html#lists-and-quote-like-blocks.

docs/creating-new/custom.rst
isaachier: Same

docs/creating-new/custom.rst
isaachier: Same

docs/creating-new/custom.rst
isaachier: Same

docs/creating-new/custom.rst
isaachier: OK

isaachier: OK

docs/creating-new/custom.rst
isaachier: Never mind it is not rendering properly.

ruslo:
Thank you!

isaachier:
My pleasure. Just hope it helps the next person who wants to add an autotools package. In order to "bootstrap" this package manager, we probably need to get a bunch of autotools packages done first. My original goal was just to add [folly](https://github.com/facebook/folly) and [Apache Thrift](https://thrift.apache.org), but neither could be packaged immediately without several dependencies being packaged first.

PR #1043, opened 2017-09-14
upload-cache-to-github.py: Limit retry time

chfast:


ruslo:
Can you add this as an option? I don't think it make sense to have such low values. The point of big values here is that if we have a lot of jobs and API is overloaded then some of them will wait for the long time allowing others to work. If you will have quite similar values the mean will be the same.

Issue #1066, opened 2017-09-30
Protobuf Hash mismatch


ibayer:
```
-- verifying file...
       file='/home/ibayer/.hunter/_Base/Download/Protobuf/3.0.0-p1/e3d5ab6/v3.0.0-hunter-3.tar.gz'
-- SHA1 hash of
    /home/ibayer/.hunter/_Base/Download/Protobuf/3.0.0-p1/e3d5ab6/v3.0.0-hunter-3.tar.gz
  does not match expected value
    expected: 'e3d5ab642e0295d16b86a4a55e06c3ee026ff249'
      actual: '242443553ca20ab279427330e175a3c33ecbf265'
-- Hash mismatch, removing...
```
Seems to be related to #1034 but I don't understand from the discussion what the recommended fix is.

headupinclouds:
If it hasn't already been fixed in the latest hunter release, you could submit a PR with the updated SHA1.

ibayer:
@headupinclouds Will do, can you point me to a PR that fixes other SHA1s? I have used hunter but are clueless about it's internals.

ibayer:
Never mind, I found it (I was greping for the wrong hash).

ibayer:
I updated hunter and it looks like the SHA1 has already been updated :+1: .
Sadly updating hunter caused new trouble #1068.

PR #1067, opened 2017-10-01
update Protobuf Hash

ibayer:
#1066


ibayer:
Has already been fixed.

PR #1070, opened 2017-10-04
Add opentracing-cpp.

isaachier:


ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2505
* https://travis-ci.org/ingenue/hunter/builds/283586306?utm_source=github_status&utm_medium=notification

docs/packages/pkg/opentracing-cpp.rst
```diff
@@ -0,0 +1,25 @@
...
+
+.. _pkg.opentracing-cpp:
+
+opentracing-cpp
+===============
+
+-  `Official <https://github.com/opentracing/opentracing-cpp>`__
+-  `Hunterized <https://github.com/hunter-packages/opentracing-cpp>`__
+-  `Example <https://github.com/ruslo/hunter/blob/master/examples/opentracing-cpp/CMakeLists.txt>`__
+-  Added by `Isaac Hier <https://github.com/isaachier>`__ (`pr-1070 <https://github.com/ruslo/hunter/pull/1070>`__)
```
ruslo: @isaachier Do you think in general we need pull request number and link here? I've just checked that we can do few clicks from documentation to find out origin: `Edit on GitHub` -> `History` -> First commit -> Link to pull request. It was useful when this page was part of the GitHub wiki which doesn't belong to main repository. I guess it's not the case right now.

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2506
* https://travis-ci.org/ingenue/hunter/builds/283606496?utm_source=github_status&utm_medium=notification

docs/packages/pkg/opentracing-cpp.rst
ruslo: From https://github.com/ruslo/hunter/pull/1072#discussion_r142896621:

> I think we should add the PR, then one can see since when the project is part of hunter (with relase and date)

@NeroBurner It's still possible to reach everything with few clicks (see above). So the question is about "few clicks" vs. making commit and pull **after** PR created (which is a little bit clumsy):
* https://docs.hunter.sh/en/latest/creating-new/create/cmake.html#pull-requests

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.19.117

docs/packages/pkg/opentracing-cpp.rst
isaachier: I find it can be sort of a pain because of the inherent race condition in determining the pull request number, but it isn't too bad either way.

PR #1072, opened 2017-10-05
docs: update create-new-package

NeroBurner:


docs/creating-new/create/cmake.rst
```diff
@@ -343,25 +343,25 @@ Add ``hunter_config`` directive with default version to
 Create example
 ==============
 
-Simple example will be used to test integration of package. Copy template
-example and substitute all strings ``foo`` with ``hunter_box_1``:
+ To test the integration of the package into Hunter a simple example will be used.
```
ruslo: > the integration of the package **into Hunter**

I don't think it's correct. We are not integrating package into Hunter. Question about integrating package into **another project**. That's what we are planning to test.

docs/creating-new/create/cmake.rst
```diff
@@ -381,10 +381,10 @@ Substitute ``unsorted`` with some tag in directive
 
 .. note::
 
-  Since you don't know the pull request number a priori leave it as ``N`` for now.
+  Since you don't know a priory pull request number leave it as ``N`` for now.
```
ruslo: I guess you need to rebase, see https://github.com/ruslo/hunter/pull/1071.

docs/creating-new/create/cmake.rst
```diff
@@ -600,20 +599,22 @@ Excluding tests
...
 has platform-specific problems. Note that you don't have to have all
 toolchains working and there is **no need to fix all issues you see**.
 If at least *something* is working then you can exclude broken
-toolchain and you or somebody else can apply fixes later.
+toolchains and you or somebody else can apply fixes later.
 
-**Do not remove** toolchains from ``.travis.yml``/``appveyor.yml`` configs.
-Comment it out instead to simplify enabling it back. Do not change the order
-of toolchains since it will affect ``git merge``. Leave the link to broken job:
+Please follow the following guidelines when disabling toolchains:
```
ruslo: > follow the following

What about `follow the next guidelines` or `follow these guidelines`?

docs/creating-new/create/cmake.rst
ruslo: Also do we need pull request number, what do you think? See my comment: https://github.com/ruslo/hunter/pull/1070#discussion_r142880694

NeroBurner: I will rebase

I think we should add the PR, then one can see since when the project is part of hunter (with relase and date)

NeroBurner:
fixed, squashed, compiled docs locally, pushed

docs/creating-new/create/cmake.rst
NeroBurner: I was editing the file directly though github while you merged my previous change #1071 

It seems, that when editing the new branch is based on the current master, not on the commit the edit was started on

PR #1080, opened 2017-10-15
support MSVC 19.12

dvirtz:
MSVC 15.5 Preview has compiler version 19.12


dvirtz:
Maybe it will be better to just use "191[0-9]" as the regex instead of updating this every compiler version change?

ruslo:
> Maybe it will be better to just use "191[0-9]" as the regex instead of updating this every compiler version change?

I guess so. Also I don't understand why this is changing. Previous versions looks stable.

dvirtz:
Found [this](https://blogs.msdn.microsoft.com/vcblog/2016/10/05/visual-c-compiler-version/) blog post which explains the versioning changes.

ruslo:
Looks good, thanks.

PR #1082, opened 2017-10-16
Thrift 0.9.2

isaachier:


ruslo:
Merged without testing because default version is not affected.

isaachier:
Cool. For the record, it works: https://travis-ci.org/isaachier/hunter/builds/288647556.

PR #1085, opened 2017-10-18
Added another license filename variant.

vdsrd:


ruslo:
> There is no need to modify this list, you can use HUNTER_INSTALL_LICENSE_FILES instead

Closing

Issue #1112, opened 2017-10-24
GTest package fails on license file copy when using it via `GIT_SUBMODULE`


Knitschi:
I think the problem is caused by the lines

````cmake
# hunter.cmake
...
if(HUNTER_GTest_VERSION VERSION_LESS 1.8.0)
  set(_gtest_license "LICENSE")
else()
  set(_gtest_license "googletest/LICENSE")
endif()
...
````

I am not sure how we can check for GIT_SUBMODULE version here and then use the lower path.

Update:
When I remove the conditional and only use the second case, the problem does not occurr. So it seams that the version comparison must be fixed. I will try to do it with a regular expression.

ruslo:
While you're using `GIT_SUBMODULE` the `HUNTER_GTest_VERSION` will be set to SHA1 of commit of submodule. This will not work. You can add this peculiarity to documentation. I guess you can use HUNTER_INSTALL_LICENSE_FILES as a workaround:
* https://docs.hunter.sh/en/latest/user-guides/cmake-user/using-license-files.html

ruslo:
Closing inactive discussion.

PR #1114, opened 2017-10-24
Fix issue1112

Knitschi:
I did two things.
1. Add the latest release of the GTest package. It seems that this was forgotten.
2. Added a fix for #1112 by modifiying the conditional. When using the package in GIT_SUBMODULE mode, its version is the 40 character long hash, which is matched by the regexp. In this case we assume that the version is newer than 1.8.0 and take the lower branch. This can still cause errors when the submodule is set to the old package repository, but I think in most cases the GIT_SUBMODULE option is used when working on the new package.

### DIFF
--- a/cmake/projects/GTest/hunter.cmake
+++ b/cmake/projects/GTest/hunter.cmake
@@ -196,7 +196,7 @@ hunter_add_version(
 
 # The regexp case is used when the package is added as GIT_SUBMODULE.
 # In this case we assume that the version is later than 1.8.0.
-if(HUNTER_GTest_VERSION VERSION_LESS 1.8.0 AND NOT A MATCHES "^[a-z0-9]{40}$" )
+if(HUNTER_GTest_VERSION VERSION_LESS 1.8.0 AND NOT A MATCHES "[a-z]+" )
   set(_gtest_license "LICENSE")
 else()
   set(_gtest_license "googletest/LICENSE")
--- a/cmake/projects/GTest/hunter.cmake
+++ b/cmake/projects/GTest/hunter.cmake
@@ -194,7 +194,9 @@ hunter_add_version(
     ab0f563b3bd80a1721cb93df034a86ec
 )
 
-if(HUNTER_GTest_VERSION VERSION_LESS 1.8.0)
+# The regexp case is used when the package is added as GIT_SUBMODULE.
+# In this case we assume that the version is later than 1.8.0.
+if(HUNTER_GTest_VERSION VERSION_LESS 1.8.0 AND NOT MATCHES "^[a-z0-9]{40}$")
   set(_gtest_license "LICENSE")
 else()
   set(_gtest_license "googletest/LICENSE")
--- a/cmake/projects/GTest/hunter.cmake
+++ b/cmake/projects/GTest/hunter.cmake
@@ -196,7 +196,7 @@ hunter_add_version(
 
 # The regexp case is used when the package is added as GIT_SUBMODULE.
 # In this case we assume that the version is later than 1.8.0.
-if(HUNTER_GTest_VERSION VERSION_LESS 1.8.0 AND NOT ( MATCHES "^[a-z0-9]{40}$" ))
+if(HUNTER_GTest_VERSION VERSION_LESS 1.8.0 AND NOT A MATCHES "^[a-z0-9]{40}$" )
   set(_gtest_license "LICENSE")
 else()
   set(_gtest_license "googletest/LICENSE")
--- a/cmake/projects/GTest/hunter.cmake
+++ b/cmake/projects/GTest/hunter.cmake
@@ -196,7 +196,7 @@ hunter_add_version(
 
 # The regexp case is used when the package is added as GIT_SUBMODULE.
 # In this case we assume that the version is later than 1.8.0.
-if(HUNTER_GTest_VERSION VERSION_LESS 1.8.0 AND NOT MATCHES "^[a-z0-9]{40}$")
+if(HUNTER_GTest_VERSION VERSION_LESS 1.8.0 AND NOT ( MATCHES "^[a-z0-9]{40}$" ))
   set(_gtest_license "LICENSE")
 else()
   set(_gtest_license "googletest/LICENSE")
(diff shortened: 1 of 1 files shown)


cmake/projects/GTest/hunter.cmake
```diff
@@ -183,7 +183,20 @@ hunter_add_version(
...
+    1.8.0-hunter-p8
+    URL
+    "https://github.com/hunter-packages/googletest/archive/1.8.0-hunter-p8.tar.gz"
+    SHA1
+    ab0f563b3bd80a1721cb93df034a86ec
+)
+
+# The regexp case is used when the package is added as GIT_SUBMODULE.
+# In this case we assume that the version is later than 1.8.0.
+if(HUNTER_GTest_VERSION VERSION_LESS 1.8.0 AND NOT MATCHES "^[a-z0-9]{40}$")
```
ruslo: I don't think it's correct condition. Try:
```cmake
set(A "0.0")
if(A VERSION_LESS 1.0 AND NOT MATCHES "^[a-z0-9]{40}$")
  message("A < 1.0")
  message("A is not SHA1")
endif()
```

I guess CMake interpret it as `"NOT" MATCHES "^[a-z0-9]{40}$"`which is `FALSE`.

Knitschi:
You are right. The check does not work like this. I added an extra pair of parentheses to fix it. Now i tested it in a script.

Knitschi:
I have to rework this, I just saw that cmake does not support `{}` in regular expressions.

Knitschi:
Now I use 
````cmake
if(HUNTER_GTest_VERSION VERSION_LESS 1.8.0 AND NOT A MATCHES "[a-z]+" )
````
which will fail if the SHA does not contain any letters. However this is very unlikely.
Also it assumes that legal version numbers do not contain letters. I think for the GTest project this is also true, because versions are currently hardcoded in the CMakeLists.txt file and they use numbers and dots.

In order to get the more robust forty-letters-or-numbers regexp I would have to do something like
````cmake
function( getSHARegexp regexpOut)
    set(regexp "^")
    foreach( index RANGE 39)
        string(APPEND regexp "[a-z0-9]")
    endforeach()
    set( ${regexpOut} "${regexp}$" PARENT_SCOPE)
endfunction()
````
which I found too ugly to use.

ruslo:
Have you tried to use `HUNTER_INSTALL_LICENSE_FILES` as suggested [here](https://github.com/ruslo/hunter/issues/1112#issuecomment-339063275)? This regex stuff doesn't seem like a good solution wherever it's ugly or not.

Knitschi:
How would you use `HUNTER_INSTALL_LICENSE_FILES` to solve the problem? Can I set multiple pathes to it and hunter will not complain if one of the files is missing? 

The current code in the `GTest/hunter.cmake` file is.

````cmake
if(HUNTER_GTest_VERSION VERSION_LESS 1.8.0)
  set(_gtest_license "LICENSE")
else()
  set(_gtest_license "googletest/LICENSE")
endif()

hunter_cmake_args(
    GTest
    CMAKE_ARGS
    HUNTER_INSTALL_LICENSE_FILES=${_gtest_license}
)
````

Can I change it to

````cmake

set( _gtest_license LICENSE googletest/LICENSE)
hunter_cmake_args(
    GTest
    CMAKE_ARGS
    HUNTER_INSTALL_LICENSE_FILES=${_gtest_license}
)
````
and it would work?

ruslo:
> How would you use HUNTER_INSTALL_LICENSE_FILES to solve the problem?

User's value from `config.cmake` overwrite default from `hunter.cmake`:
* https://github.com/ruslo/hunter/wiki/example.custom.config.id#order

You will force the value you need.

Knitschi:
Now I get it :-)
It works. But should we leave the `hunter.cmake` as it is? There is a good chance that somebody else will run into the same problem. The last regular expression `"[a-z]+"` definitely works and would avoid the problem in most cases. 

Alternatively I can mention the problem in the docs of the GTest package.

ruslo:
> Now I get it :-)

What you have tried exactly? `HUNTER_INSTALL_LICENSE_FILES` in local `config.cmake`?

> But should we leave the hunter.cmake as it is?

My suggestion doesn't involve the change of `hunter.cmake`, we are leaving it as is.

> There is a good chance that somebody else will run into the same problem
> Alternatively I can mention the problem in the docs of the GTest package

I guess it belongs to `GIT_SUBMODULE` feature documentation:
* https://docs.hunter.sh/en/latest/user-guides/hunter-user/git-submodule.html

Need to note that:
* `HUNTER_<package>_VERSION` set to `SHA1` of submodule instead of real version
* `hunter.cmake` can use `HUNTER_<package>_VERSION` to add extra `CMAKE_ARGS` or get a build scheme or anything else. So user is responsible for keeping things consistent.

Knitschi:
Yes I set HUNTER_INSTALL_LICENSE_FILES in the `config.cmake`file:

````cmake
# config.cmake
...
hunter_config( 
    GTest 
    GIT_SUBMODULE googletest 
    CMAKE_ARGS HUNTER_INSTALL_LICENSE_FILES=googletest/LICENSE
)
...
````

I think it would be better to add the regexp to the package file because it may prevent users from running into problems, which may frustrate them. But its your call ...

Should I make pull requests for adding the latest version of the GTest package and for changeing the docs?

ruslo:
> I think it would be better to add the regexp to the package file because it may prevent users from running into problems, which may frustrate them

I don't see the difference, your approach will not work for users which will try older version (not newer version like you). I'm ready to listen for some robust solution.

> Should I make pull requests for adding the latest version of the GTest package and for changeing the docs?

Sure, I guess it's two pull requests actually.

Knitschi:
>I don't see the difference, your approach will not work for users which will try older version (not newer version like you). I'm ready to listen for some robust solution.

That is true, but my guess is, that more people will use the later versions, so the amount of users who run into problems would be smaller. I do not have the knowledge about what data is available in the `config.cmake` file to come up with a real solution for the problem. I will keep this problem in the back of my head. Maybe at some point I will have the knowledge to solve it properly.

So I close the request for now.

ruslo:
> So I close the request for now.

The `1.8.0-hunter-p8` update part was okay.

Knitschi:
I created a separate pull request for that #1122

PR #1129, opened 2017-10-29
Add 'SFML' package

NukeBird:


NukeBird:
Do I forget something?

headupinclouds:
> Do I forget something?

It looks like there are some bundled static libraries for audio support that should be managed as packages:

https://github.com/SFML/SFML/blob/013d053277c980946bc7761a2a088f1cbb788f8c/src/SFML/CMakeLists.txt#L5-L40

https://github.com/SFML/SFML/blob/013d053277c980946bc7761a2a088f1cbb788f8c/src/SFML/Audio/CMakeLists.txt#L58-L104

NukeBird:
@headupinclouds It seems that they are part of configuration file https://github.com/SFML/SFML/blob/013d053277c980946bc7761a2a088f1cbb788f8c/CMakeLists.txt#L464

NukeBird:
>, but SFML seems to be using precompiled binaries for these dependencies. This is a case where we will need to fork the projects.

Oh... Sounds painful D:

cmake/projects/SFML/hunter.cmake
```diff
@@ -0,0 +1,24 @@
...
+    PACKAGE_NAME
+    SFML
+    VERSION
+    2.4.2
+    URL
+    "https://github.com/SFML/SFML/archive/2.4.2.tar.gz"
+    SHA1
+	275C68B51B8509BA90EDA6D5DEF791B497F1F777
+)
+
```
headupinclouds: hunter_cmake_args(
    SFML
    CMAKE_ARGS
        SFML_BUILD_AUDIO
)

headupinclouds:
I see.  So `SFML_BUILD_AUDIO=NO` could be a default option:

https://github.com/SFML/SFML/blob/013d053277c980946bc7761a2a088f1cbb788f8c/CMakeLists.txt#L96-L98

See https://github.com/ruslo/hunter/pull/1129#pullrequestreview-72704522

headupinclouds:
> Oh... Sounds painful D:

It is definitely more work, but not too bad.  Once it is added, everyone in the community will benefit from it.  Maybe the changes will be accepted upstream as an optional solution for the static libs.  One package is already CMake based, and Hunter has autoconf bindings for the other two (although it can be cleaner to just add the required CMake).  I've been interested in experimenting with SFML for a little while, so I'd be interested in helping with the effort.

cmake/projects/SFML/hunter.cmake
NukeBird: Yeah, it make sense. I'll edit it a little bit

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2681
* https://travis-ci.org/ingenue/hunter/builds/294533993?utm_source=github_status&utm_medium=notification

cmake/projects/SFML/hunter.cmake
headupinclouds: 👍

cmake/projects/SFML/hunter.cmake
```diff
@@ -0,0 +1,30 @@
...
+hunter_cmake_args(
+    SFML
+    CMAKE_ARGS
+        SFML_BUILD_AUDIO=NO
+)
+
```
NukeBird: So... Here I must add
`hunter_add_package(OpenAL)`

And fork SFML repo for patching it's CMakeLists? (to add the following line to it: `find_package(OpenAL CONFIG REQUIRED)`)?

headupinclouds: You add `hunter_add_package(OpenAL)` in the CMakeLists.txt file.  Every `find_package(FOO)` should be paired with `hunter_add_package(FOO CONFIG REQUIRED)` (options vary).  When hunter is enabled, then it will build and install `FOO` and the `find_package(FOO)` command will work.  If you disable hunter, `hunter_add_package()` calls becomes hoops, and cmake will look for the package somewhere on your system.

headupinclouds: > And fork SFML repo for patching it's CMakeLists? (to add the following line to it: find_package(OpenAL CONFIG REQUIRED))?

You can create an issue asking for a fork of the project in `hunter-packages` where a minimal set of CMake changes can be applied.

headupinclouds:
> So this code: [HunterGate] must be a part of it too?

The `HunterGate` call (and `cmake/Hunter/HunterGate.cmake` file) is needed in any project that has dependencies managed by hunter (i.e., any project with `find_package()` calls).  If the library has no dependencies, then it doesn't need `HunterGate`.

NukeBird:
https://github.com/SFML/SFML/blob/fae3b65f0567f87fa9925cd42d28df15eb69e79c/src/SFML/Audio/CMakeLists.txt#L68

It seems that we need add here
```
hunter_add_package(OpenAL) #(support is added via #1128 )
hunter_add_package(FLAC)
hunter_add_package(vorbis)
```
And audio module will work fine. But it also means that we need to add support for such packages as FLAC and vorbis


 >You can create an issue asking for a fork of the project in hunter-packages where the CMake changes can be applied and you will be added as a collaborator. Having the forks in a central place facilitates collaboration and testing until (ideally) changes are accepted upstream.

Okay. Will you help me? Work together will be easier (I hope that I will not be an extra ballast for you, haha ^^)

cmake/projects/SFML/hunter.cmake
ruslo: It's already forked: https://github.com/hunter-packages/SFML

I've created branch `hunter-2.4.2` from `2.4.2` tag, please send patches there.

headupinclouds:
> Okay. Will you help me? Work together will be easier

Sure, I can help.

ruslo:

PR #1133, opened 2017-10-29
add ogg + test

headupinclouds:


headupinclouds:
@NukeBird

NukeBird:
@headupinclouds wow! Good job!

I'll try integrate this dependency inside SFML fork #1129 when it falls into the hunter release

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2688
* https://travis-ci.org/ingenue/hunter/builds/294821794?utm_source=github_status&utm_medium=notification

ruslo:
> Testing

Some tests failed. Include fixed here: https://github.com/hunter-packages/ogg/commit/b55363a8c53cca4b71549d18abcb97b90fdf90f7

By the way we already have `ogg`, it's just called `libogg`:
* https://github.com/ruslo/hunter/blob/79b3645ae22222844373e53edc7e900ccd1c6128/cmake/projects/libogg/hunter.cmake

Probably just need to make an update.

headupinclouds:
I've added a release with your include fix, but...

> By the way we already have ogg, it's just called libogg

Okay.  Shall we merge these updates into the existing libogg?

ruslo:
> Okay. Shall we merge these updates into the existing libogg?

Yes, I guess so.

headupinclouds:
The original `libogg` package was added by @Meralis40.  I noticed the `PACKAGE_UNRELOCATABLE_TEXT_FILES` tag.

Here is the text:
```
hunter_add_version(
        PACKAGE_NAME
        libogg
        VERSION
        "1.3.2-cmake3"
        URL
        "https://github.com/Meralis40/ogg/archive/v1.3.2-cmake3.tar.gz"
        SHA1
  646672cabc9eec253c9199a872541e3a182f6062
)

hunter_pick_scheme(DEFAULT url_sha1_cmake)
hunter_cacheable(libogg)
hunter_download(
    PACKAGE_NAME
    libogg
    PACKAGE_UNRELOCATABLE_TEXT_FILES
    "lib/pkgconfig/ogg.pc"
)
```

So there is no `hunter-packages/libogg` for the merge.  I think the `hunter-packages/ogg` repository associated with this release should be used to host patches moving forward.  I'm unsure about what the actual package name should be.  I think `ogg` is probably a better package name, but there are already folks using `libogg`, so that would be a breaking change, or possibly a hunter major version increase?

Option 1) Merge this `ogg` package and either remove `libogg` or keep it with a "warning deprecated package" note? 

Option 2) Keep the `hunter-package/ogg` releases and just change the hunter package name in this PR to `libogg` for backward compatibility.  I believe the default `CMakeLists.txt` uses `target_link_libraries(foo ogg)` whereas the CMake mods in `hunter-packages/ogg` are now using `target_link_libraries(foo ogg::ogg)`.  I think it might be better to change this back to `target_link_libraries(foo ogg)` for compatibility with the upstream lib. 

@ruslo , @Meralis40 :  I have no strong opinions about this.  Since you both merged the previous package, please share your thoughts.

ruslo:
> So there is no hunter-packages/libogg for the merge

They have common base repository so in theory merge is possible:
* https://github.com/xiph/ogg

>  I think the hunter-packages/ogg repository associated with this release should be used to host patches moving forward

Agree.

>  I'm unsure about what the actual package name should be

Repository name is `ogg`, it installs `lib/pkgconfig/ogg.pc` but target name is `libogg::ogg`. Let's keep `libogg` for now, I see no big difference. We can change it later if all CMake stuff will be merged upstream and it became official.

ruslo:

PR #1135, opened 2017-10-30
Add a document section that mentions the HUNTER_<package>_VERSION var…

Knitschi:
As discussed in #1114 I added a few lines to the documentation that mention the `HUNTER_<package>_VERSION` variable. However I did not add this to the Git Submodule documentation because I think the section that handles setting the `CMAKE_ARGS` in the `hunter.cmake` is a better fit.

I have no experience with Sphinx so I am not sure if the link to the Git Submodule package will work as expected. Please check that.

### DIFF
--- a/docs/creating-new/create/cmake.rst
+++ b/docs/creating-new/create/cmake.rst
@@ -310,6 +310,14 @@ Options set by ``hunter_cmake_args`` have lower precedence than options set
 by ``hunter_config(... CMAKE_ARGS ...)`` (see
 `order <https://github.com/ruslo/hunter/wiki/example.custom.config.id#order>`__).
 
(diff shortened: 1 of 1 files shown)


ruslo:
> I have no experience with Sphinx so I am not sure if the link to the Git Submodule package will work as expected. Please check that.

It should be as simple as `source ./jenkins.sh` and `./make.sh`, see:
* https://docs.hunter.sh/en/latest/creating-new/create/cmake.html#add-documentation

> However I did not add this to the Git Submodule documentation because I think the section that handles setting the CMAKE_ARGS in the hunter.cmake is a better fit

I see no point of going such deep in documentation for adding new package. This is something user should know when he will try to use `GIT_SUBMODULE`, so it definitely should belong there.

Knitschi:
Ok, I will make another pull request for that.

PR #1200, opened 2017-11-14
Edited Windows testing commands

jowr:
Fixed the `Path` handling on Windows

### DIFF
--- a/docs/creating-new/create/cmake.rst
+++ b/docs/creating-new/create/cmake.rst
@@ -449,9 +449,6 @@ On Windows:
 
 .. code-block:: none
 
-  > where python
-  C:\Python3\python.exe
-
   > git clone https://github.com/ruslo/polly
   > cd polly
   [polly]> set PATH=%CD%\bin;%PATH%
@@ -467,7 +464,7 @@ On Windows:
 
   [hunter]> set TOOLCHAIN=vs-12-2013
   [hunter]> set PROJECT_DIR=examples\hunter_box_1
-  [hunter]> python jenkins.py
+  [hunter]> ./jenkins.py
 
 .. admonition:: Stackoverflow
 
--- a/docs/creating-new/create/cmake.rst
+++ b/docs/creating-new/create/cmake.rst
@@ -454,10 +454,10 @@ On Windows:
   [polly]> set PATH=%CD%\bin;%PATH%
 
   > cd hunter
-  [hunter]> where polly.bat
-  C:\...\bin\polly.bat
+  [hunter]> where polly.py
+  C:\...\bin\polly.py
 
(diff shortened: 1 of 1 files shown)


docs/creating-new/create/cmake.rst
```diff
@@ -452,19 +449,25 @@ On Windows:
 
 .. code-block:: none
 
-  > cd hunter
+  > where python
```
ruslo: I don't have Python in my `PATH` and it works fine.

docs/creating-new/create/cmake.rst
```diff
@@ -452,19 +449,25 @@ On Windows:
...
+
+  [hunter]> polly.bat --help
   Python version: 3.5
   usage: polly.py [-h]
   ...
 
   [hunter]> set TOOLCHAIN=vs-12-2013
   [hunter]> set PROJECT_DIR=examples\hunter_box_1
-  [hunter]> .\jenkins.py
+  [hunter]> python jenkins.py
```
ruslo: Using `.\jenkins.py` works fine. You should tune your file associations, see the Stackoverflow question below.

jowr:
You are right, those changes just clutter the PR. I only meant to add the command that changes the PATH on Windows. I hope it is better now.

docs/creating-new/create/cmake.rst
```diff
@@ -452,12 +449,15 @@ On Windows:
...
+  [polly]> set PATH=%CD%\bin;%PATH%
 
-  [hunter]> where polly.py
-  C:\...\bin\polly.py
+  > cd hunter
+  [hunter]> where polly.bat
+  C:\...\bin\polly.bat
 
-  [hunter]> polly.py --help
+  [hunter]> polly.bat --help
```
ruslo: Actually it's not working for me because I have no Python in PATH :)

If file association is set correctly we can check only `polly.py --help`. By the way `jenkins.py` use old name `build.py`:
* https://github.com/ruslo/hunter/blob/0e0e1f87c7b8ef7ace8673cb9380590ce5ee5f5c/jenkins.py#L144-L156

jowr:
OK, I reverted that one as well. You might want to add a comment somewhere regarding the outdated commands. It is quite difficult to get started with hunterization - but I am sure it will be worth it in the end :-)

ruslo:
> You might want to add a comment somewhere regarding the outdated commands.

What commands are outdated? `build.py` vs `polly.py` difference is purely internal and don't make any difference for end user.

Note that currently your patch only adds `set PATH=%CD%\bin;%PATH%` for Windows, and usually Windows users do `System -> Advanced system settings`. All old commands are correct.

Issue #1106, opened 2017-10-22
More global configs


ruslo:
At this point only one global config is available, `default`:
* https://github.com/ruslo/hunter/blob/master/cmake/configs/default.cmake

`HunterGate` designed so another global configs can be used:
* https://github.com/hunter-packages/gate#usage-custom-config

This will work kind of similar to Ubuntu releases:
* https://en.wikipedia.org/wiki/Ubuntu_version_history#Version_timeline

Say we have:
* Config `dapper` with `foo` 1.2.0, `boo` 3.4.4
* Config `hardy` with `foo` 1.3.4, `boo` 3.5.7
* Config `lucid` with `foo` 2.0.1, `boo` 3.6.1

On new releases:
* `foo` 1.2.0 -> `foo` 1.2.1: only `dapper` updated, `hardy`/`lucid` remains the same
* `boo` 3.6.1 -> `boo` 3.6.2: only `lucid` updated, `dapper`/`hardy` remains the same
* etc.

If package `bar` require at least `foo` 1.3 then it will be available in `hardy` and `lucid` but not in `dapper`.

Note that [config-id](https://docs.hunter.sh/en/latest/overview/customization/config-id.html) is not only about versions but also about options. So it's possible to create something like `minimal` where all possible features of packages are off for minimal functionality/size: `FOO_WITH_A=OFF`, `FOO_WITH_B=OFF`, `BAR_WITH_C=OFF`, etc. and `full` with all features are on for maximum funtionality but big size: `FOO_WITH_A=ON`, `FOO_WITH_B=ON`, `BAR_WITH_C=ON` (note that it's orthogonal to `MinSizeRel` build type).

ruslo:
https://semver.org/ like workflow:
* global config `stable` for PATCH updates
* global config `next` for MINOR updates
* global config `edge` for MAJOR updates

E.g.:
* `foo` 1.2.0 -> 1.2.1 update goes to `stable`, `next` and `edge` remains the same
* `boo` 3.4.5 -> 3.7.0 update goes to `next`, `edge` and `stable` remains the same

After some period of time:
* `stable` dropped
* current `next` became `stable` and start accepting PATCH updates
* current `edge` became `next` and start accepting MINOR updates
* current `edge` remains as `edge` and continue accepting MAJOR updates

PR #1206, opened 2017-11-16
Add libthrift

isaachier:
Thrift C++ library without compiler and dependencies.


cmake/configs/default.cmake
```diff
@@ -219,6 +219,7 @@ hunter_config(libjson-rpc-cpp VERSION 0.7.0-p3)
 hunter_config(libmill VERSION 1.18)
 hunter_config(libogg VERSION 1.3.2-cmake3)
 hunter_config(libsodium VERSION 1.0.10)
+hunter_config(libthrift VERSION 0.10.0)
```
ruslo: Two equal projects with the different names?

isaachier: They pass in different cmake args. Also, because libthrift has no dependency on bison, it is cachable.

ruslo: Both use URL `https://github.com/hunter-packages/thrift/archive/v0.10.0-p2.tar.gz` - hence they are the same packages.

> They pass in different cmake args

Can be controlled by custom `CMAKE_ARGS`:

`CMAKE_ARGS` in `hunter.cmake` just set some reasonable default values. The rest is up to user. Also see this feature request: https://github.com/ruslo/hunter/issues/1106

isaachier: What about `hunter_cacheable` and the fact that libthrift supports many more platforms?

ruslo: Set best `CMAKE_ARGS` configuration as a default and add warnings to documentation about other options. If cacheable/uncacheable depends on CMAKE_ARGS then user can break it just by specifying custom options. Having two packages that do install same files is a badly designed solution.

isaachier:
OK I think I understand. One last question: why does Protobuf have 3 packages?

ruslo:
> why does Protobuf have 3 packages?

What do you mean? I see only one:
* https://docs.hunter.sh/en/latest/packages/all.html?highlight=protobuf

isaachier:
I guess you are right. Just three examples.

PR #1216, opened 2017-11-18
add KEEP_PACKAGE_SOURCES option to hunter_config()

headupinclouds:
PR for initial review/discussion:
* append `KEEP_PACKAGE_SOURCES` option in hunter_calculate_config_sha1.cmake if specified
* introduce _hunter_keep_package_sources in hunter_download.cmake as logical OR of global `HUNTER_KEEP_PACKAGE_SOURCES` and local `HUNTER_${h_name}_KEEP_PACKAGE_SOURCES`

### DIFF
--- a/cmake/modules/hunter_calculate_config_sha1.cmake
+++ b/cmake/modules/hunter_calculate_config_sha1.cmake
@@ -128,11 +128,6 @@ function(hunter_calculate_config_sha1 hunter_self hunter_base user_config)
         )
       endif()
 
-      string(COMPARE NOTEQUAL "{HUNTER_${x}_KEEP_PACKAGE_SOURCES}" "" have_keep_package_sources)
-      if(have_keep_package_sources)
-        file(APPEND "${input_file}" " KEEP_PACKAGE_SOURCES")
-      endif()
-      
       file(APPEND "${input_file}" ")\n")
     endif()
   endforeach()
--- a/cmake/modules/hunter_calculate_config_sha1.cmake
+++ b/cmake/modules/hunter_calculate_config_sha1.cmake
@@ -128,6 +128,11 @@ function(hunter_calculate_config_sha1 hunter_self hunter_base user_config)
         )
       endif()
 
+      string(COMPARE NOTEQUAL "{HUNTER_${x}_KEEP_PACKAGE_SOURCES}" "" have_keep_package_sources)
+      if(have_keep_package_sources)
+        file(APPEND "${input_file}" " KEEP_PACKAGE_SOURCES")
+      endif()
+      
       file(APPEND "${input_file}" ")\n")
     endif()
   endforeach()
--- a/cmake/modules/hunter_config.cmake
+++ b/cmake/modules/hunter_config.cmake
@@ -93,4 +93,6 @@ macro(hunter_config)
     set(HUNTER_${_hunter_current_project}_KEEP_PACKAGE_SOURCES ON)
   endif()
 
+  message("MOD: ${_hunter_current_project}  ${HUNTER_${_hunter_current_project}_KEEP_PACKAGE_SOURCES}")
+
 endmacro()
--- a/cmake/modules/hunter_config.cmake
+++ b/cmake/modules/hunter_config.cmake
@@ -90,9 +90,12 @@ macro(hunter_config)
   endif()
 
   if(_hunter_KEEP_PACKAGE_SOURCES)
-    set(HUNTER_${_hunter_current_project}_KEEP_PACKAGE_SOURCES ON)
+    set_property(
+      GLOBAL
+      PROPERTY
+      "HUNTER_${_hunter_current_project}_KEEP_PACKAGE_SOURCES"
+      ON
+      )
   endif()
-
-  message("MOD: ${_hunter_current_project}  ${HUNTER_${_hunter_current_project}_KEEP_PACKAGE_SOURCES}")
 
 endmacro()
--- a/cmake/modules/hunter_config.cmake
+++ b/cmake/modules/hunter_config.cmake
@@ -18,11 +18,12 @@ macro(hunter_config)
         "error.unexpected.hunter_config"
     )
   endif()
+  set(_hunter_optional KEEP_PACKAGE_SOURCES)
   set(_hunter_one_value VERSION GIT_SUBMODULE GIT_SUBMODULE_DIR)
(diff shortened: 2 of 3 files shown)


cmake/modules/hunter_download.cmake
```diff
@@ -429,7 +429,7 @@ function(hunter_download)
   file(
       APPEND
       "${HUNTER_DOWNLOAD_TOOLCHAIN}"
-      "set(HUNTER_KEEP_PACKAGE_SOURCES \"${HUNTER_KEEP_PACKAGE_SOURCES}\" CACHE INTERNAL \"\")\n"
+      "set(HUNTER_KEEP_PACKAGE_SOURCES \"${_hunter_keep_package_sources}\" CACHE INTERNAL \"\")\n"
```
ruslo: Variable `_hunter_keep_package_source` is not defined at this point.

cmake/modules/hunter_calculate_config_sha1.cmake
```diff
@@ -128,6 +128,11 @@ function(hunter_calculate_config_sha1 hunter_self hunter_base user_config)
         )
       endif()
 
+      string(COMPARE NOTEQUAL "{HUNTER_${x}_KEEP_PACKAGE_SOURCES}" "" have_keep_package_sources)
```
ruslo: I don't see the point of changing `config-id`. Logically it's not affected.

cmake/modules/hunter_config.cmake
```diff
@@ -87,4 +88,9 @@ macro(hunter_config)
   else()
     hunter_user_error("Expected VERSION option for 'hunter_config' command")
   endif()
+
+  if(_hunter_KEEP_PACKAGE_SOURCES)
+    set(HUNTER_${_hunter_current_project}_KEEP_PACKAGE_SOURCES ON)
```
headupinclouds: LINK

cmake/modules/hunter_calculate_config_sha1.cmake
headupinclouds: I see.  The `set` call in `hunter_config. cmake` was being cl

When I had the following log at the bottom of `hunter_config.cmake`
```
  message("MOD: ${_hunter_current_project}  ${HUNTER_${_hunter_current_project}_KEEP_PACKAGE_SOURCES}")
```

It was clearing getting cleared on the third call.  

`grep MOD _logs/polly/default/log.txt  | grep GTest`

```
MOD: GTest
MOD: GTest  ON
MOD: GTest
```

So the package source wasn't being kept.  I noticed the third call to `hunter_config()` was coming from this generated file, and this mod fixed it:

```
MOD: GTest
MOD: GTest  ON
MOD: GTest  ON  # fixed
``` 

This all seems to be due to the misplaced variable in hunter_config.cmake [here](https://github.com/ruslo/hunter/pull/1216/files/13fdc60b17e81560b0c5e7cc0e24ebf53927ae36#r151868052).

headupinclouds:
I've pushed a modified version up for discussion.  I'll prepare a minimal test to reproduce the issue I'm seeing.

cmake/modules/hunter_download.cmake
ruslo: All packages that depends on this package will be build with `HUNTER_KEEP_PACKAGE_SOURCES=ON` even if you want to keep sources only for one package. TL;DR I think this should be reverted.

cmake/modules/hunter_download.cmake
```diff
@@ -382,6 +382,10 @@ function(hunter_download)
   file(REMOVE "${HUNTER_PACKAGE_HOME_DIR}/CMakeLists.txt")
   file(REMOVE "${HUNTER_DOWNLOAD_TOOLCHAIN}")
 
+  if(HUNTER_KEEP_PACKAGE_SOURCES OR HUNTER_${h_name}_KEEP_PACKAGE_SOURCES)
+    set(_hunter_keep_package_sources ON)
```
ruslo: Let's add
```cmake
else()
  set(_hunter_keep_package_sources OFF)
endif()
```
just to be sure.

cmake/modules/hunter_download.cmake
headupinclouds: Reverted (will submit clean pr once it is working)

cmake/modules/hunter_download.cmake
headupinclouds: Changes applied [here](https://github.com/ruslo/hunter/pull/1216/commits/964c7fac0cada93077e51a22970d3f2fd4841afa)

The `HUNTER_${h_name}_KEEP_PACKAGE_SOURCES` is still getting cleared by the 3rd call to `hunter_config()` 

`grep "MOD: GTest" _logs/polly/libcxx/log.txt`

```
MOD: GTest
MOD: GTest  ON
MOD: GTest
```

which I believe comes from an evaluation of this call:

```
hunter_calculate_config_sha1.cmake:          "hunter_config(${x} "
```

ruslo: > The HUNTER_${h_name}_KEEP_PACKAGE_SOURCES is still getting cleared by the 3rd call

It's not cleared - it wasn't set in scope we need (there is no `set(HUNTER_..._SOURCES "")` command or `unset(HUNTER_..._SOURCES)`).

TL;DR Use global property: https://github.com/ruslo/hunter/commit/c82aed035c3db6813d29306cc9e5b1f5b1a1fa06

```
MOD: GTest
MOD: GTest  ON
MOD: GTest
```

We are processing config 3 times:
* global default config: https://github.com/ruslo/hunter/blob/353628aee9e834e72c82ae9377c7fff019a11327/cmake/modules/hunter_calculate_config_sha1.cmake#L28-L30
* local config: https://github.com/ruslo/hunter/blob/353628aee9e834e72c82ae9377c7fff019a11327/cmake/modules/hunter_calculate_config_sha1.cmake#L36-L38
* unified config (merge of two configs which is used to calculate config-id and forwarding for all 3rd parties): https://github.com/ruslo/hunter/blob/353628aee9e834e72c82ae9377c7fff019a11327/cmake/modules/hunter_finalize.cmake#L136-L138

First two processed inside function so user's scope is not affected. We need something that will outlive local scope. It can be chain of `PARENT_SCOPE`, cache variables or properties. Global property fits nicely here.

headupinclouds: Very helpful.  Thanks.

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.19.165

PR #1232, opened 2017-11-29
Crypt32 required for all Windows

xsacha:
I found an unresolved symbol issue tracked back to missing crypt32 that appeared to only be resolved for MinGW in Hunter.

libcrypto.lib(e_capi.obj) : error LNK2001: unresolved external symbol __imp_CertOpenStore [Y:\J\workspace\...\...\....vcxproj]

So I have added it for all WIN32 instead.

### DIFF
--- a/cmake/find/FindOpenSSL.cmake
+++ b/cmake/find/FindOpenSSL.cmake
@@ -374,14 +374,14 @@ if(OPENSSL_FOUND)
     set_target_properties(OpenSSL::Crypto PROPERTIES
       INTERFACE_INCLUDE_DIRECTORIES "${OPENSSL_INCLUDE_DIR}")
     find_package(Threads REQUIRED)
-    if(MINGW)
+    if(WIN32)
       # https://github.com/openssl/openssl/issues/1061
-      set(_openssl_mingw_crypt "crypt32")
+      set(_openssl_win32_crypt "crypt32")
     else()
-      set(_openssl_mingw_crypt "")
+      set(_openssl_win32_crypt "")
     endif()
     set_target_properties(OpenSSL::Crypto PROPERTIES
-        INTERFACE_LINK_LIBRARIES "${CMAKE_DL_LIBS};Threads::Threads;${_openssl_mingw_crypt}")
+        INTERFACE_LINK_LIBRARIES "${CMAKE_DL_LIBS};Threads::Threads;${_openssl_win32_crypt}")
     if(EXISTS "${OPENSSL_CRYPTO_LIBRARY}")
       set_target_properties(OpenSSL::Crypto PROPERTIES
         IMPORTED_LINK_INTERFACE_LANGUAGES "C"


xsacha:
2017 but it appears it affects all Windows versions.

This only applies to the FindOpenSSL so I don't think depsid apllies? It doesn't need to be recompiled?

ruslo:
> 2015 but it appears it affects all of Windows

Visual Studio 2015 looks good:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2876

Can you add an example to reproduce the problem?

ruslo:
> This only applies to the FindOpenSSL so I don't think depsid apllies? It doesn't need to be recompiled?

Since `FindOpenSSL` used by dependencies and result of it's work can be cached we have to instruct dependencies that update needed.

xsacha:
> Can you add an example to reproduce the problem?

Any static Windows build that uses libcrypto will have this issue.

> [winnode]      Creating library Y:/J/.../test.lib and object Y:/J/.../test.exp
> [winnode] libcrypto.lib(e_capi.obj) : error LNK2001: unresolved external symbol __imp_CertOpenStore [Y:\J\...\test.vcxproj]
> [winnode] libcrypto.lib(e_capi.obj) : error LNK2001: unresolved external symbol __imp_CertCloseStore [Y:\J\...\test.vcxproj]
> [winnode] libcrypto.lib(e_capi.obj) : error LNK2001: unresolved external symbol __imp_CertEnumCertificatesInStore [Y:\J\...\test.vcxproj]
> [winnode] libcrypto.lib(e_capi.obj) : error LNK2001: unresolved external symbol __imp_CertFindCertificateInStore [Y:\J\...\test.vcxproj]
> ...

ruslo:
> Any static Windows build that uses libcrypto will have this issue.

OpenSSL-imported example use static libraries:
* https://github.com/ruslo/hunter/blob/master/examples/OpenSSL-imported/CMakeLists.txt

Builds fine:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.2876/job/2m47vh0kbop6x4kq

```
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\link.exe
  ...
  C:\projects\hunter\_testing\Hunter\_Base\c73e82e\326fee5\12fa5c4\Install\lib\libssl.lib
  C:\projects\hunter\_testing\Hunter\_Base\c73e82e\326fee5\12fa5c4\Install\lib\libcrypto.lib
  ...
```

xsacha:
I don't understand why this is closed.
Do you have a test to reproduce the cause of this commit? https://github.com/xsacha/hunter/commit/5ec7b520d088387080ebe5ffa4808cbc29ee5c84#diff-400923a8030d1d270a8781e09ba98eba

The issue it links to even says it is required for Windows, not MinGW specifically. I'm not sure how it came to be that the original commit was only pushed for MinGW
It also clearly affects builds using hunter packages. Putting a test app in to hunter in order to trigger this just to confirm a known upstream documentation artifact is both redundant and a waste of everyone's time.

Issue #1032, opened 2017-09-12
The hash value does not match.


DubrovinLeo:
cmake version 3.9.1
msys2
```
curl 7.55.1 (x86_64-w64-mingw32) libcurl/7.55.1 OpenSSL/1.0.2l zlib/1.2.11 libidn2/2.0.4 libssh2/1.8.0 nghttp2/1.23.1 librtmp/2.3
Release-Date: 2017-08-14
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtmp rtsp scp sftp smtp smtps telnet tftp
Features: AsynchDNS IDN IPv6 Largefile SSPI Kerberos SPNEGO NTLM SSL libz TLS-SRP HTTP2 HTTPS-proxy Metalink
```

My CMakeLists.txt :
```
cmake_minimum_required(VERSION 3.0)

include("HunterGate.cmake")
HunterGate(
        URL "https://github.com/ruslo/hunter/archive/v0.19.98.tar.gz"
        SHA1 "e9906a08a9778759a7c590fccff0703bca72dceb"
)

project(mychip8)
set(CMAKE_CXX_STANDARD 11)

hunter_add_package(SDL2)
find_package(SDL2 CONFIG REQUIRED)

set(${PROJECT_NAME}_SOURCES
        main.cpp
        )

add_executable(${PROJECT_NAME}
        ${${PROJECT_NAME}_SOURCES}
        )

target_link_libraries(${PROJECT_NAME}
        SDL2::SDL2main
        SDL2::SDL2)
```

Why in the next error hash values is different?

```
-- verifying file...
       file='C:/.hunter/_Base/Download/SDL2/2.0.4-p4/523d8fd/2.0.4-p4.tar.gz'
-- SHA1 hash of
    C:/.hunter/_Base/Download/SDL2/2.0.4-p4/523d8fd/2.0.4-p4.tar.gz
  does not match expected value
    expected: '523d8fd427f707541c7e600ad4c37e0e7bdc3662'
      actual: '589fcb46963b28e9d968646af54add6f4b19890c'
-- Hash mismatch, removing...
-- Retry after 60 seconds (attempt #5) ...
-- Using src='https://github.com/hunter-packages/SDL2/archive/2.0.4-p4.tar.gz'
-- [download 0% complete]
-- [download 1% complete]
-- [download 2% complete]
-- [download 3% complete]
-- [download 4% complete]
-- [download 5% complete]
[...]

PR #1236, opened 2017-11-30
fix Catch 1.5.9 hash

dvirtz:
hash changed similarly to https://github.com/ruslo/hunter/issues/1032

### DIFF
--- a/cmake/projects/Catch/hunter.cmake
+++ b/cmake/projects/Catch/hunter.cmake
@@ -39,7 +39,7 @@ hunter_add_version(
     URL
     "https://github.com/hunter-packages/Catch/archive/v1.5.9-hunter.tar.gz"
     SHA1
-    737cb1c98fedccceb95e7bfd385e5dea0ad5d047
+    5ac39da617fe3ae91951529fba1ab493b296ab7a
 )
 
 hunter_cmake_args(


cmake/projects/Catch/hunter.cmake
```diff
@@ -39,7 +39,7 @@ hunter_add_version(
     URL
     "https://github.com/hunter-packages/Catch/archive/v1.5.9-hunter.tar.gz"
     SHA1
-    737cb1c98fedccceb95e7bfd385e5dea0ad5d047
```
ruslo: Do you have old archive with this hash? I can upload old files so it will not affect caches and dependencies. Take a look at `~/.hunter/Download` directory: https://docs.hunter.sh/en/latest/reference/layouts/deployed.html#download

dvirtz: yes, uploaded to https://github.com/dvirtz/uploads/blob/master/v1.5.9-hunter.tar.gz

ruslo: It's not an archive, it's some HTML page. I think you've used wrong link.

dvirtz: sorry, try https://github.com/dvirtz/uploads/raw/master/v1.5.9-hunter.tar.gz

ruslo: Released: https://github.com/ruslo/hunter/releases/tag/v0.19.176

PR #1287, opened 2018-01-02
use upstream log4cplus

headupinclouds:
Note: The upstream release is tagged "rc", but there has not been a new release since July.


headupinclouds:
UPDATE: I missed the bundled dependencies on my first read (catch and thread pool): 
https://github.com/log4cplus/log4cplus/blob/0e81395831a2802a92db2efba0eb448f12657263/CMakeLists.txt#L145-L146

PR #1326, opened 2018-01-25
update gpuimage, add example (objective-c++)

headupinclouds:
Note that testing will only work for APPLE toolchains (OSX/iOS).


examples/GPUImage/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+# Emulate HunterGate:
+# * https://github.com/hunter-packages/gate
+include("../common.cmake")
+
+project(download-GPUImage)
+
+hunter_add_package(GPUImage)
+find_package(GPUImage CONFIG REQUIRED)
+
+add_executable(foo foo.mm)
```
ruslo: > Note that testing will only work for APPLE toolchains (OSX/iOS).

You can do something like this:
```cmake
find_package(GPUImage CONFIG REQUIRED)
if(NOT TARGET GPUImage::gpuimage)
  message(FATAL_ERROR "Target not found")
endif()
if(APPLE)
  add_executable(foo foo.mm)
  target_link_libraries(foo GPUImage::gpuimage)
endif()
```

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.3255
* https://travis-ci.org/ingenue/hunter/builds/333402515?utm_source=github_status&utm_medium=notification

ruslo:
Integrated as https://github.com/ruslo/hunter/commit/c8d99cce599a93d81a550abdf466c288c0262773

Released: https://github.com/ruslo/hunter/releases/tag/v0.19.238

PR #1342, opened 2018-02-03
Fix location of version_not_found check

headupinclouds:
* My change will work with CMake 3.0 (minimum requirement for Hunter). **[Yes]**

---

If a `hunter_config()` call specifies a `VERSION` that isn’t defined by a `hunter_add_version()` call (hunter/cmake/<package>/hunter.cmake), perhaps due to a typo, then a call to `hunter_download_server_url()` will be made with an empty `VERSION`, and the reported error looks like this:

```
-- [hunter *** DEBUG *** 2018-02-03T17:12:33] Foo versions available: [1.0.9999999]

[hunter ** INTERNAL **] 'hunter_download_server_url' incorrect usage
[hunter ** INTERNAL **]  option 'SHA1' with one argument is mandatory.
```

In this case may have asked for something like `hunter_config(Foo VERSION 1.0.0)`, or a typo `hunter_config(FOO VERSION 1.0.0y)`, but only a valid `VERSION` `1.0.9999999` exists.

This PR moves the existing HUNTER_PACKAGE_SHA1 version_not_found test above the call to `hunter_download_server_url()`, such that a more intuitive error is reported with the strong: "Version not found: ${ver}. See 'hunter_config' command."

With this change, the log listed above then becomes more human readable:

```
-- [hunter *** DEBUG *** 2018-02-03T17:15:09] FunctionalPlus versions available: [1.0.9999999]

[hunter ** FATAL ERROR **] Version not found: 1.0.0. See 'hunter_config' command.
```

and the developer can more easily diagnose and fix the source of the VERSION mismatch error.

### DIFF
--- a/cmake/modules/hunter_download.cmake
+++ b/cmake/modules/hunter_download.cmake
@@ -1,5 +1,6 @@
-# Copyright (c) 2013-2017, Ruslan Baratov
-# Copyright (c) 2015, Aaditya Kalsi
+# Copyright (c) 2013-2018, Ruslan Baratov
+# Copyright (c) 2015-2018, Aaditya Kalsi
+# Copyright (c) 2018, David Hirvonen
 # All rights reserved.
 
 include(CMakeParseArguments) # cmake_parse_arguments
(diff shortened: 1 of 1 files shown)


ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.19.244

PR #1388, opened 2018-03-23
Add dlib 19.10

xsacha:

  * https://ci.appveyor.com/project/xsacha/hunter-h4uji
  * https://travis-ci.org/xsacha/hunter/builds/357186104


ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.3499
* https://travis-ci.org/ingenue/hunter/builds/357273181?utm_source=github_status&utm_medium=notification

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.3501
* https://travis-ci.org/ingenue/hunter/builds/357287078?utm_source=github_status&utm_medium=notification

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.20.25

xsacha:
Strange, I now try to use this in my project and get an error:
Unknown CMake command "check_required_components".

ruslo:
```cmake
configure_file(cmake_utils/dlibConfig.cmake.in "${CMAKE_CURRENT_BINARY_DIR}/config/dlibConfig.cmake" @ONLY)
```

@xsacha Should be `configure_package_config_file` , not just `configure_file`. I think that current testing use all dependencies and `check_required_components` loaded from one of the `find_package` calls from Config.cmake.in. I guess you have minimal configuration without dependencies.

xsacha:
That's true, we configure without any dependencies but I didn't have any issues on 19.8.
It's very confusing because dlib previously didn't have a configure/install and has just added it in 19.10.
I tried to keep as much the same from upstream as possible but enough from Hunter so it works fine with Hunter.

ruslo:
> I tried to keep as much the same from upstream as possible but enough from Hunter so it works fine with Hunter.

This change is not Hunter related, it's worth adding it upstream.

Issue #1397, opened 2018-04-08
Add "ROS bare-bones" packages


lsolanka:
I would like to propose adding ROS bare-bones stack to Hunter: https://github.com/ros

The current idea is to hunterise some of the core packages separately, in order to be able to reproduce the installation from source: http://wiki.ros.org/kinetic/Installation/Source.

Then the users would use, e.g. `hunter_add_package(ros-ros_comm)` inside their CMakeLists.txt to pull the whole hunterised ros bare-bones. That, in turn, would pull all the transitive dependencies.

I have all packages required for the bare-bones installation from source hunterised, so I would like to start a discussion about potentially making these inside the mainstream Hunter, if it is of interest to the wider community.

The main dependencies are:
- Boost
- OpenCV
- Python (during the build both the python interpreter, as well as PythonLibs for C extensions)
- tinyxml2
- log4cxx
- empy
- Poco

A few caveats:
- ROS uses a 'catkin' build system, which is basically CMake based, however the build proceeds in a non-standard way (i.e. they have the concept of a devel space instead of install-and-find-package). Therefore I had to patch the catkin repository quite extensively to support the 'hunter_add_package' directive.
- I did not have time/resources to hunterise some dependencies of log4cxx, such as libapr1, libapr-utils etc. and these have to be linked in as shared libraries, otherwise the logging will only print to stdout.
- Perfect hunterisation is not possible given the complexity of the ROS build system and thus pushing upstream is unlikely
- ROS has many more dependencies required for build than those stated above, which normally have to be resolved using the 'rosdep' tool, so I'm not sure how well the packages will integrate with the Hunter CI here.
- The consequence is that the build might work nicely on Ubuntu, but not on Windows that much. I haven't had any time to test on Windows as I'm not a windows user.

The work was done partly during my working hours in [SeeByte Ltd.](http://www.seebyte.com), so it would be fair to at least acknowledge them in the docs if the packages become part of the community.

TODO list for all the partial packages:
- [x] https://github.com/lsolanka/console_bridge => #1403
- [x] https://github.com/lsolanka/catkin => #1407
- [x] https://github.com/lsolanka/roscpp_core => #1412
- [x] https://github.com/lsolanka/genmsg => #1410
- [x] https://github.com/lsolanka/geneus => #1416 
- [x] https://github.com/lsolanka/gencpp => #1416
- [x] https://github.com/lsolanka/genpy => #1416
- [x] https://github.com/lsolanka/gennodejs => #1416
[...]

PR #1403, opened 2018-04-14
Add `ros_console_bridge` package

lsolanka:
I've followed [this guide](https://docs.hunter.sh/en/latest/creating-new/create/cmake.html) step by step carefully, until section [CI Testing](https://docs.hunter.sh/en/latest/creating-new/create/cmake.html#ci-testing), which instructs me to create several branches based on the package name.

I stopped there because I would first like to get feedback on package naming: The original repository is called `console_bridge` (https://github.com/ros/console_bridge) but I thought because this is part of ROS, each package could have `ros-` prepended to its name, i.e. in this case `ros-console_bridge`, to distinguish the set of ROS-related packages.

I'm not sure if that is desired so I thought I would first open the PR and then when the package names are clarified I would finish the CI build.

This PR is part of this "New package" request: #1397.


ruslo:
> The original repository is called console_bridge (https://github.com/ros/console_bridge) but I thought because this is part of ROS, each package could have ros- prepended to its name

Yes, I guess so. Actually it probably make sense to rename ros related forks. What do you think?

ruslo:
Renamed.

I kept names for:
* https://github.com/hunter-packages/catkin
* https://github.com/hunter-packages/angles (C++ namespace angles inside)
* https://github.com/hunter-packages/geometry2 (C++ namespace tf2 inside, should it be named tf2?)

lsolanka:
> https://github.com/hunter-packages/geometry2 (C++ namespace tf2 inside, should it be named tf2?)

Sounds good to me to rename. It contains only tf2_* and a geometry2 metapackage which other catkin packages cannot depend on anyway (catkin policy).

So the users would then do:
```cmake
hunter_add_package(tf2)
find_package(catkin COMPONENTS tf2 tf2_ros) # or just find_package(tf2) if they prefer not to use catkin
```


> https://github.com/hunter-packages/catkin

Sounds fine as well, users would then do:
```cmake
hunter_add_package(catkin)
find_package(catkin COMPONENTS catkin_pkg1 catkin_pkg2)
```
which is more natural than `hunter_add_package(ros_catkin)` and then `find_package(catkin)`

Also if catkin and tf2 do not have a `ros_` prefix, I would suggest to keep `actionlib` with the original name instead of `ros_actionlib`. This way, the usual set of packages required in 90% of situations would be:
```cmake
hunter_add_package(catkin)
hunter_add_package(ros_comm)
hunter_add_package(actionlib)
hunter_add_package(tf2)
```

ruslo:
Repositories renamed, please check.

lsolanka:
Looks good.

docs/packages/pkg/ros_console_bridge.rst
```diff
@@ -0,0 +1,23 @@
...
+.. _pkg.ros_console_bridge:
+
+ros_console_bridge
+==================
+
+-  `Official <https://github.com/ros/console_bridge>`__
+-  `Hunterized <https://github.com/hunter-packages/ros_console_bridge>`__
+-  `Example <https://github.com/ruslo/hunter/blob/master/examples/ros_console_bridge/CMakeLists.txt>`__
+-  Added by
+    - `Lukas Solanka <https://github.com/lsolanka>`__ (`pr-N <https://github.com/ruslo/hunter/pull/N>`__)
```
ruslo: Use `1403`  instead of `N`

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.3621
* https://travis-ci.org/ingenue/hunter/builds/366850780?utm_source=github_status&utm_medium=notification

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.3624
* https://travis-ci.org/ingenue/hunter/builds/366867648?utm_source=github_status&utm_medium=notification

PR #1436, opened 2018-05-15
Don't clobber LD_LIBRARY_PATH for autotools projects

ambroff:
Prepend the global install dir to LD_LIBRARY_PATH, rather than
just overwriting an existing value. Without this patch, a toolchain
relying on LD_LIBRARY_PATH to find the C library will fail.

### DIFF
--- a/cmake/modules/hunter_autotools_project.cmake
+++ b/cmake/modules/hunter_autotools_project.cmake
@@ -111,7 +111,7 @@ function(hunter_autotools_project target_name)
   set(default_path "/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin")
   set(shell_env_path "PATH=${PARAM_GLOBAL_INSTALL_DIR}/bin:${default_path}")
 
-  set(shell_ld_path "LD_LIBRARY_PATH=${PARAM_GLOBAL_INSTALL_DIR}/lib")
+  set(shell_ld_path "LD_LIBRARY_PATH=${PARAM_GLOBAL_INSTALL_DIR}/lib:$ENV{LD_LIBRARY_PATH}")
 
   set(d1 "${PARAM_GLOBAL_INSTALL_DIR}/lib/pkgconfig")
   set(d2 "${PARAM_GLOBAL_INSTALL_DIR}/share/pkgconfig")


ruslo:
> Without this patch, a toolchain relying on LD_LIBRARY_PATH to find the C library will fail.

What toolchain are you using?

ambroff:
We aren't using one of your polly toolchains. One of my coworkers had built GCC 7 and Clang 6 from source, and was using a CMAKE_TOOLCHAIN_FILE which used those compilers. He was relying on LD_LIBRARY_PATH instead of updating the loader search path.

I personally don't think using LD_LIBRARY_PATH is a great way to do this, but we both agreed that just overwriting LD_LIBRARY_PATH instead of appending or prepending was surprising behavior.

ruslo:
> but we both agreed that Hunter just overwriting LD_LIBRARY_PATH instead of appending or prepending was surprising behavior

Hunter creates isolated development environment with libraries and on purpose don't rely on system/global libraries. Also a lot of variables that can leak into build will be unset:
* https://github.com/ruslo/hunter/blob/master/scripts/clear-all.cmake

> One of my coworkers had built GCC 7 and Clang 6 from source

I have custom Clang installed on my local machine and I can build packages without `LD_LIBRARY_PATH` modification. What libraries cause you troubles exactly?

> Without this patch, a toolchain relying on LD_LIBRARY_PATH to find the C library will fail

You mean `libc.so`?

ambroff:
> You mean libc.so?

Yup

> I have custom Clang installed on my local machine and I can build packages without LD_LIBRARY_PATH modification. What libraries cause you troubles exactly?

Yeah it's not a common problem. The issue is that he's targeting RHEL 7 which has an ancient release of glibc (2.17, back from like 2012). So in order for him to get a modern version of libstdc++ he's essentially forced to build all of GCC, including a newer glibc and libstdc++ than is available on the system. Then he built llvm and clang using that newer glibc, libstdc++.

He didn't want to add the lib directory of his toolchain to /etc/ld.so.conf.d/, so he assumed he could just set LD_LIBRARY_PATH and forget about it.

Without either of those, just running the compiler or any program that is linked will result in errors like

`version 'GLIBCXX_3.4.20' not found (required by /path/to/clang)`

> Hunter creates isolated development environment with libraries and on purpose don't rely on system/global libraries.

Yeah I totally get that. We found resetting LD_LIBRARY_PATH surprising, but I understand the goal of only depending on libraries provided by Hunter when possible. The exception of course is the C library and C++ standard library.

I dunno. Maybe this isn't a problem Hunter has to be concerned with. It's only a problem if you're using a modern compiler to target an older system and you don't install it properly. The only reason that I hesitate to close this is that we found the behavior unexpected, even though as you pointed out it does make sense.

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.20.79

PR #1444, opened 2018-05-19
Upgrade opentracing-cpp to version 1.4.1

mdouaihy:
  Here is the links to the Travis with status "All passed":

  * https://travis-ci.org/mdouaihy/hunter/builds/381175656


mdouaihy:
There is a problem with the last release of OpenTracing. I ll work on a fix and issue a new PR.

PR #1465, opened 2018-06-06
OpenBLAS: Fix host Android compiles

xsacha:
This is required as OpenBLAS upstream doesn't support automatically configured arm compiles, so we workaround this by manually setting target and telling it to avoid tests (CROSS=1), which is the same as when we do an actual cross-compile.
This will be fixed in OpenBLAS 0.3.0 but I haven't updated the CMake config for it yet.

Very small update that doesn't affect any targeted builds.

### DIFF
--- a/cmake/projects/OpenBLAS/hunter.cmake
+++ b/cmake/projects/OpenBLAS/hunter.cmake
@@ -40,7 +40,7 @@ endif()
 hunter_cacheable(OpenBLAS)
 hunter_download(
     PACKAGE_NAME OpenBLAS
-    PACKAGE_INTERNAL_DEPS_ID "1"
+    PACKAGE_INTERNAL_DEPS_ID "2"
     PACKAGE_UNRELOCATABLE_TEXT_FILES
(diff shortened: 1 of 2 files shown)


xsacha:
It doesn't change build on anything that could already build. Does deps id need to go up?

ruslo:
> It doesn't change build on anything that could already build. Does deps id need to go up?

It does change build for you so it may change build for somebody too. It may find some tricky way to be saved in cache and hence make a whole mess on this update. It's easier to change one variable to be sure than potentially spend hours trying to understand where the magic really happened.

xsacha:
Added to deps ID

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.4074
* https://travis-ci.org/ingenue/hunter/builds/389196750?utm_source=github_status&utm_medium=notification

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.22.5

PR #1466, opened 2018-06-06
Update link-all.cmake

xsacha:
Allow link-all to fallthrough if there is an error in the python script


ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.22.4

PR #1468, opened 2018-06-09
Add OpenBLAS 0.3.0

xsacha:

  * https://ci.appveyor.com/project/xsacha/hunter-h4uji/build/1.0.13
  * https://travis-ci.org/xsacha/hunter/builds/390176136


cmake/projects/OpenBLAS/hunter.cmake
```diff
@@ -32,15 +44,20 @@ hunter_add_version(
...
+    CMAKE_ARGS
+    NOFORTRAN=1
+  )
+  hunter_pick_scheme(DEFAULT url_sha1_cmake)
 endif()
 hunter_cacheable(OpenBLAS)
 hunter_download(
     PACKAGE_NAME OpenBLAS
     PACKAGE_INTERNAL_DEPS_ID "2"
-    PACKAGE_UNRELOCATABLE_TEXT_FILES
```
ruslo: This change will break old versions like `0.2.20-p0`.

As far as I can see since version `0.3.0-p2` we will not need `OpenBLAS` custom build scheme, right? In this case I'm okay with removing `0.2.*` versions since changing custom build scheme to CMake based build scheme is a super cool event :) We will not need `PACKAGE_INTERNAL_DEPS_ID` and `PACKAGE_UNRELOCATABLE_TEXT_FILES` in this case too.

ruslo: > Or is there a way to make the package_unrelocatable_text_files depend on some variable?

```cmake
if(HUNTER_OpenBLAS_VERSION VERSION_LESS 0.3.0)
  set(
      _openblas_unrelocatable_text_files
      PACKAGE_UNRELOCATABLE_TEXT_FILES
      "lib/cmake/openblas/OpenBLASConfig.cmake
  )
else()
  set(_openblas_unrelocatable_text_files "")
endif()

hunter_download(
    PACKAGE_NAME OpenBLAS
    PACKAGE_INTERNAL_DEPS_ID "2"
    ${_openblas_unrelocatable_text_files}
)
```

xsacha: OK, I've done that. Also made MSVC consistent (it didn't work previously in 0.2.20 anyway).

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.4082
* https://travis-ci.org/ingenue/hunter/builds/390197020?utm_source=github_status&utm_medium=notification

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.4083
* https://travis-ci.org/ingenue/hunter/builds/390200120?utm_source=github_status&utm_medium=notification

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.22.9

PR #1470, opened 2018-06-10
Fix setting of CXXFLAGS in hunter_get_build_flags (Affect all autotools pkg builds)

Bjoe:
There is a typo in `hunter_get_build_flags` modul. `CXXFLAGS` was set with `CFLAGS` instead of `CXXFLAGS`. This affect all autotools builds.

### DIFF
--- a/cmake/modules/hunter_get_build_flags.cmake
+++ b/cmake/modules/hunter_get_build_flags.cmake
@@ -129,7 +129,7 @@ function(hunter_get_build_flags)
     )
     string(STRIP "${cxxflags}" cxxflags)
     hunter_status_debug("  CXXFLAGS=${cxxflags}")
-    set(${PARAM_OUT_CXXFLAGS} ${cflags} PARENT_SCOPE)
+    set(${PARAM_OUT_CXXFLAGS} ${cxxflags} PARENT_SCOPE)
   endif()
 
   if(has_out_ldflags)
--- a/cmake/projects/gstreamer/hunter.cmake
+++ b/cmake/projects/gstreamer/hunter.cmake
@@ -35,7 +35,7 @@ hunter_cmake_args(
 hunter_cacheable(gstreamer)
 hunter_download(
     PACKAGE_NAME gstreamer
-    PACKAGE_INTERNAL_DEPS_ID "1"
+    PACKAGE_INTERNAL_DEPS_ID "2"
     PACKAGE_UNRELOCATABLE_TEXT_FILES
     "lib/gstreamer-1.0/libgstcoreelements.la"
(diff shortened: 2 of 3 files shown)


ruslo:
> This affect all autotools builds

Probably this simply means that all current autotools based projects use only C compiler and don't use C++ compiler. Have you hit any particular issue?

Bjoe:
Yes, but unfortunately is not yet public. I implemented pjproject into hunter and try to build it for android. PjProject has some C++ code (not much) and there I get the problem that the C++ includes not set ... only C headers. In android you can choose std C++ implementations. This means the header files are not in the same directory as the C headers. On Linux or any other host machine, all header files, C and C++ are all on the same place, for example /usr/include. That's why it works ...

ruslo:
> There is a typo in hunter_get_build_flags modul

`hunter_get_build_flags` -> `hunter_autotools_configure_command` -> `hunter_autotools_project`.

I've added "gstreamer" package (which use C++ compiler for one of the component) to testing note:
* https://github.com/ruslo/hunter/blob/5a6facd399e9b9bbdf6e06852cef7fba7733fe3e/cmake/modules/hunter_autotools_project.cmake#L64-L66

Please update `PACKAGE_INTERNAL_DEPS_ID` for those packages so we can run tests. Simply increment value in `hunter.cmake`:
* https://github.com/ruslo/hunter/blob/master/cmake/projects/gstreamer/hunter.cmake
* https://github.com/ruslo/hunter/blob/master/cmake/projects/xau/hunter.cmake

For details:
* https://docs.hunter.sh/en/latest/reference/user-modules/hunter_download.html

Bjoe:
Ok, I update `PACKAGE_INTERNAL_DEPS_ID` on gstreamer and xau. Btw. where I can see the "test builds" ?

ruslo:
Testing.

xau:
* https://travis-ci.org/ingenue/hunter/builds/391449120?utm_source=github_status&utm_medium=notification

gstreamer:
* https://travis-ci.org/ingenue/hunter/builds/391449370?utm_source=github_status&utm_medium=notification

> where I can see the "test builds" ?

https://github.com/ingenue/hunter/branches/all?utf8=%E2%9C%93&query=pkg.

ruslo:
Uploading.

xau: https://travis-ci.org/ingenue/hunter/builds/391453840?utm_source=github_status&utm_medium=notification
gstreamer: https://travis-ci.org/ingenue/hunter/builds/391453738?utm_source=github_status&utm_medium=notification

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.22.13

Issue #558, opened 2016-10-21
Error in hunter_download_cache_meta_file.cmake when using CMake 3.0


Knitschi:
I am tying to update the GTest package and re-read the contribution guide in this context. There it is stated that the package should work with CMake 3.0.

Now I try to compile my example with that version but get the following error from hunter:

````
CMake Error at C:/CppLibraries/hunter-develop/_testing/Hunter/_Base/Download/Hunter/unknown/71fb0c2/Unpacked/cmake/modules/hunter_download_cache_meta_file.cmake:107 (continue):
  Unknown CMake command "continue".
````

The continue command was introduced with CMake 3.2
You also use the lock feature which requires to use the -DHUNTER_SKIP_LOCK=ON option when using cmake versions lower than 3.2

So maybe it is sensible to increase the minimum required CMake version to 3.2.
Which cmake version is used for the automated tests?

ruslo:
> The continue command was introduced with CMake 3.2

Yep, kind of a bug.

> Which cmake version is used for the automated tests?

CMake 3.6+

> So maybe it is sensible to increase the minimum required CMake version to 3.2.

Yes, I like this idea. Also we can get rid of `string(COMPARE EQUAL ...)` stuff because since CMake 3.1 the CMP0054 policy itroduced. The plan is next:

- [ ] Update minimum version to 3.2
- [ ] Update documentation about 3.2
- [ ] Add the `continue` feature in 3.2 section in documentation
- [ ] Save in documentation the latest Hunter version with CMake 3.0 as minimum required. In case somebody want to maintain lower version he can start there
- [ ] Remove `HUNTER_SKIP_LOCK` from code
- [ ] Add a note to documentation of `HUNTER_SKIP_LOCK` that it's no longer needed (keep page for the previous versions)
- [ ] Remove `string(COMPARE EQUAL ...)` commands and use `if(... STREQUAL ...)` instead

Feel free to send the pathes.

PR #1481, opened 2018-06-26
Proper parse of linker flags in pkg config files in pkgconfig_export_target

Bjoe:
If a project provide on MacOSx following pkgconfig `*.pc` file:

 ```
...
 Libs: -L${libdir} -lfoo-apple -lbar -lpthread  -framework CoreAudio -framework CoreFoundation
 ...
```
Then `hunter_pkgconfig_export_target.cmake` is set `INTERFACE_LINK_LIBRARIES` in cmake as following:

` -Lfoo/bar;-lfoo-apple;-lbar;-lpthread;-framework;CoreAudio;-framework;CoreFoundation`

The linker get following parameters:

` ... -Lfoo/bar -lfoo-apple -lbar -lpthread -framework -lCoreAudio -lCoreFoundation`

Linker fails with: "Cannot find CoreAudio library"

This request set `INTERFACE_LINK_LIBRARIES` with following line:

` -Lfoo/bar;-lfoo-apple;-lbar;-lpthread;-framework CoreAudio;-framework CoreFoundation`

The linker gets following parameters:

` ... -Lfoo/bar -lfoo-apple -lbar -lpthread -framework CoreAudio -framework CoreFoundation`

Now it works :-)

### DIFF
--- a/cmake/modules/hunter_pkgconfig_export_target.cmake
+++ b/cmake/modules/hunter_pkgconfig_export_target.cmake
@@ -71,7 +71,9 @@ function(hunter_pkgconfig_export_target PKG_CONFIG_MODULE)
       has_ldflags_other
   )
   if(has_ldflags_other)
-    list(APPEND link_libs ${${PKG_CONFIG_MODULE}_LDFLAGS_OTHER})
+    # turn "-framework;A;-framework;B" into "-framework A;-framework B"
+    string(REPLACE "-framework;" "-framework " ldflags_other "${${PKG_CONFIG_MODULE}_LDFLAGS_OTHER}")
+    list(APPEND link_libs ${ldflags_other})
   endif()
 
   # No need to treat the pkg-config module's _LIBRARY_DIRS and _LIBRARIES
--- a/cmake/modules/hunter_pkgconfig_export_target.cmake
+++ b/cmake/modules/hunter_pkgconfig_export_target.cmake
@@ -33,12 +33,7 @@ function(hunter_pkgconfig_export_target PKG_CONFIG_MODULE)
   hunter_status_debug(
       "PKG_CONFIG_MODULE ${PKG_CONFIG_MODULE} INCLUDE_DIRS: ${${PKG_CONFIG_MODULE}_INCLUDE_DIRS}"
   )
-  string(COMPARE NOTEQUAL
-      "${${PKG_CONFIG_MODULE}_INCLUDE_DIRS}"
-      ""
-      has_include_dirs
-  )
-  if(has_include_dirs)
+  if(NOT "${${PKG_CONFIG_MODULE}_INCLUDE_DIRS}" STREQUAL "")
     set_target_properties("${target_name}"
         PROPERTIES
           INTERFACE_INCLUDE_DIRECTORIES
@@ -53,24 +48,14 @@ function(hunter_pkgconfig_export_target PKG_CONFIG_MODULE)
   hunter_status_debug(
       "PKG_CONFIG_MODULE ${PKG_CONFIG_MODULE} LDFLAGS: ${${PKG_CONFIG_MODULE}_LDFLAGS}"
   )
-  string(COMPARE NOTEQUAL
-      "${${PKG_CONFIG_MODULE}_LDFLAGS}"
-      ""
-      has_ldflags
-  )
-  if(has_ldflags)
+  if(NOT "${${PKG_CONFIG_MODULE}_LDFLAGS}" STREQUAL "")
     list(APPEND link_libs ${${PKG_CONFIG_MODULE}_LDFLAGS})
   endif()
 
   hunter_status_debug(
       "PKG_CONFIG_MODULE ${PKG_CONFIG_MODULE} LDFLAGS_OTHER: ${${PKG_CONFIG_MODULE}_LDFLAGS_OTHER}"
   )
-  string(COMPARE NOTEQUAL
-      "${${PKG_CONFIG_MODULE}_LDFLAGS_OTHER}"
-      ""
-      has_ldflags_other
-  )
-  if(has_ldflags_other)
+  if(NOT "${${PKG_CONFIG_MODULE}_LDFLAGS_OTHER}" STREQUAL "")
     # turn "-framework;A;-framework;B" into "-framework A;-framework B"
     string(REPLACE "-framework;" "-framework " ldflags_other "${${PKG_CONFIG_MODULE}_LDFLAGS_OTHER}")
     list(APPEND link_libs ${ldflags_other})
@@ -78,12 +63,7 @@ function(hunter_pkgconfig_export_target PKG_CONFIG_MODULE)
 
   # No need to treat the pkg-config module's _LIBRARY_DIRS and _LIBRARIES
   # as they are already included in LD_FLAGS
-  string(COMPARE NOTEQUAL
-      "${link_libs}"
-      ""
-      has_link_libs
-  )
-  if(has_link_libs)
+  if(NOT "${link_libs}" STREQUAL "")
     set_target_properties("${target_name}"
         PROPERTIES
           INTERFACE_LINK_LIBRARIES
@@ -98,32 +78,18 @@ function(hunter_pkgconfig_export_target PKG_CONFIG_MODULE)
   hunter_status_debug(
       "PKG_CONFIG_MODULE ${PKG_CONFIG_MODULE} CFLAGS: ${${PKG_CONFIG_MODULE}_CFLAGS}"
   )
-  string(COMPARE NOTEQUAL
-      "${${PKG_CONFIG_MODULE}_CFLAGS}"
-      ""
-      has_cflags)
-  if(has_cflags)
+  if(NOT "${${PKG_CONFIG_MODULE}_CFLAGS}" STREQUAL "")
     list(APPEND compile_opts ${${PKG_CONFIG_MODULE}_CFLAGS})
   endif()
 
   hunter_status_debug(
       "PKG_CONFIG_MODULE ${PKG_CONFIG_MODULE} CFLAGS_OTHER: ${${PKG_CONFIG_MODULE}_CFLAGS_OTHER}"
   )
-  string(COMPARE NOTEQUAL
-      "${${PKG_CONFIG_MODULE}_CFLAGS_OTHER}"
-      ""
-      has_cflags_other
-  )
-  if(has_cflags_other)
+  if(NOT "${${PKG_CONFIG_MODULE}_CFLAGS_OTHER}" STREQUAL "")
     list(APPEND compile_opts ${${PKG_CONFIG_MODULE}_CFLAGS_OTHER})
   endif()
 
-  string(COMPARE NOTEQUAL
-     "${compile_opts}"
-     ""
-     has_compile_opts
-  )
-  if(has_compile_opts)
+  if(NOT "${compile_opts}" STREQUAL "")
       set_target_properties("${target_name}"
           PROPERTIES
             INTERFACE_COMPILE_OPTIONS
--- a/cmake/projects/x11/hunter.cmake
+++ b/cmake/projects/x11/hunter.cmake
@@ -41,7 +41,7 @@ hunter_cmake_args(
 hunter_cacheable(x11)
 hunter_download(
     PACKAGE_NAME x11
-    PACKAGE_INTERNAL_DEPS_ID "4"
+    PACKAGE_INTERNAL_DEPS_ID "5"
     PACKAGE_UNRELOCATABLE_TEXT_FILES
     "lib/libX11-xcb.la"
     "lib/libX11.la"
--- a/cmake/projects/x264/hunter.cmake
+++ b/cmake/projects/x264/hunter.cmake
@@ -25,6 +25,7 @@ hunter_cmake_args(x264 CMAKE_ARGS PKGCONFIG_EXPORT_TARGETS=x264)
 hunter_cacheable(x264)
 hunter_download(
     PACKAGE_NAME x264
+    PACKAGE_INTERNAL_DEPS_ID "1"
     PACKAGE_UNRELOCATABLE_TEXT_FILES "lib/pkgconfig/x264.pc"
 )
 
--- a/cmake/projects/xcb/hunter.cmake
+++ b/cmake/projects/xcb/hunter.cmake
@@ -63,6 +63,6 @@ hunter_pick_scheme(DEFAULT xcb)
 hunter_cacheable(xcb)
 hunter_download(
     PACKAGE_NAME xcb
-    PACKAGE_INTERNAL_DEPS_ID "3"
+    PACKAGE_INTERNAL_DEPS_ID "4"
     PACKAGE_UNRELOCATABLE_TEXT_FILES "${_xcb_text_files}"
 )

### CODE REVIEW

File: cmake/modules/hunter_pkgconfig_export_target.cmake [hunk 1]
```diff
@@ -47,15 +47,42 @@ function(hunter_pkgconfig_export_target PKG_CONFIG_MODULE)
...
+    list(APPEND link_libs ${${PKG_CONFIG_MODULE}_LINK_LIBRARIES})
+  endif()
+
+  hunter_status_debug(
+      "PKG_CONFIG_MODULE ${PKG_CONFIG_MODULE} LIBRARY_DIRS: ${${PKG_CONFIG_MODULE}_LIBRARY_DIRS}"
+  )
+  string(COMPARE NOTEQUAL
+      "${${PKG_CONFIG_MODULE}_LIBRARY_DIRS}"
+      ""
+      has_ldflags
```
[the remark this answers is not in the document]
Bjoe: ok

cmake/modules/hunter_pkgconfig_export_target.cmake
```diff
@@ -47,15 +47,42 @@ function(hunter_pkgconfig_export_target PKG_CONFIG_MODULE)
...
+  )
+  if(has_ldflags)
+	foreach(_dir ${${PKG_CONFIG_MODULE}_LIBRARY_DIRS})
```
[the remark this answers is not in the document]
Bjoe: ok

cmake/modules/hunter_pkgconfig_export_target.cmake
```diff
@@ -47,15 +47,42 @@ function(hunter_pkgconfig_export_target PKG_CONFIG_MODULE)
...
+	endforeach()
+  endif()
+
+  hunter_status_debug(
+      "PKG_CONFIG_MODULE ${PKG_CONFIG_MODULE} LIBRARIES: ${${PKG_CONFIG_MODULE}_LIBRARIES}"
+  )
+  string(COMPARE NOTEQUAL
+      "${${PKG_CONFIG_MODULE}_LIBRARIES}"
+      ""
+      has_ldflags
```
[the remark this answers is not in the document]
Bjoe: ok

cmake/modules/hunter_pkgconfig_export_target.cmake
```diff
@@ -67,11 +94,18 @@ function(hunter_pkgconfig_export_target PKG_CONFIG_MODULE)
       has_ldflags_other
   )
   if(has_ldflags_other)
-    list(APPEND link_libs ${${PKG_CONFIG_MODULE}_LDFLAGS_OTHER})
+	foreach(_ldflag_other ${${PKG_CONFIG_MODULE}_LDFLAGS_OTHER})
```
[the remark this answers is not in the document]
Bjoe: done

Bjoe:
> Can you please add few comments about how it corresponds to *_LDFLAGS, *_LINK_LIBRARIES, *_LIBRARY_DIRS, etc. so it will be easier to understand why we can't just use *_LDFLAGS.

You mean "here" or in the `hunter_pkgconfig_export_target.cmake` file?

ruslo:
> > Can you please add few comments about how it corresponds to *_LDFLAGS, *_LINK_LIBRARIES, *_LIBRARY_DIRS, etc. so it will be easier to understand why we can't just use *_LDFLAGS.

> You mean "here" or in the hunter_pkgconfig_export_target.cmake file?

Few comments to `hunter_pkgconfig_export_target.cmake` module.

Bjoe:
Btw.  I remove string(COMPARE EQUAL ...) commands and use if(... STREQUAL ...) instead [where you mention here](https://github.com/ruslo/hunter/issues/558#issuecomment-255492072)

cmake/modules/hunter_pkgconfig_export_target.cmake
```diff
@@ -51,37 +46,53 @@ function(hunter_pkgconfig_export_target PKG_CONFIG_MODULE)
...
-  string(COMPARE NOTEQUAL
-      "${link_libs}"
-      ""
-      has_link_libs
+  hunter_status_debug(
+      "PKG_CONFIG_MODULE ${PKG_CONFIG_MODULE} LDFLAGS_OTHER: ${${PKG_CONFIG_MODULE}_LDFLAGS_OTHER}"
   )
-  if(has_link_libs)
+  if(NOT "${${PKG_CONFIG_MODULE}_LDFLAGS_OTHER}" STREQUAL "")
+    # Get all other linker flags
```
ruslo: will this code do the same:
```cmake
# turn "-framework;A;-framework;B" into "-framework A;-framework B"
string(REPLACE "-framework;" "-framework " ldflags_other "${${PKG_CONFIG_MODULE}_LDFLAGS_OTHER}")
```

?

Bjoe: Hmm, maybe... let me think, we take the list and works like a string and replaces `;` with ` ` ... hm should also work.

Bjoe:
Let me test your aproach on my mac ... but I have also anothor mandatory fix in `hunter_pkgconfig_export_target.cmake`. Should I create a new pull request or should I use this pull request?

ruslo:
> Should I create a new pull request or should I use this pull request?

If this fix can be send as independent nonconfict patch then it's better to open new pull request for testing/review purposes.

Bjoe:
Sorry take a little bit longer ... but I tested your approach on MacOS. It works, I changed `hunter_pkgconfig_export_target.cmake`

cmake/modules/hunter_pkgconfig_export_target.cmake
```diff
@@ -51,37 +46,45 @@ function(hunter_pkgconfig_export_target PKG_CONFIG_MODULE)
...
-      "PKG_CONFIG_MODULE ${PKG_CONFIG_MODULE} LDFLAGS_OTHER: ${${PKG_CONFIG_MODULE}_LDFLAGS_OTHER}"
+      "PKG_CONFIG_MODULE ${PKG_CONFIG_MODULE} LIBRARY_DIRS: ${${PKG_CONFIG_MODULE}_LIBRARY_DIRS}"
   )
-  string(COMPARE NOTEQUAL
-      "${${PKG_CONFIG_MODULE}_LDFLAGS_OTHER}"
-      ""
-      has_ldflags_other
+  if(NOT "${${PKG_CONFIG_MODULE}_LIBRARY_DIRS}" STREQUAL "")
+    # Get only the paths of the libraries without '-L' and set with the '-L'
+    foreach(_dir ${${PKG_CONFIG_MODULE}_LIBRARY_DIRS})
```
ruslo: There is no need to start variable name with `_`, we are inside function. Can be simplified to:
```cmake
foreach(dir ${${PKG_CONFIG_MODULE}_LIBRARY_DIRS})
  list(APPEND link_libs -L${dir})
endforeach()
```

cmake/modules/hunter_pkgconfig_export_target.cmake
```diff
@@ -51,37 +46,45 @@ function(hunter_pkgconfig_export_target PKG_CONFIG_MODULE)
...
   )
-  if(has_link_libs)
+  if(NOT "${${PKG_CONFIG_MODULE}_LDFLAGS_OTHER}" STREQUAL "")
+    # turn "-framework;A;-framework;B" into "-framework A;-framework B"
+    string(REPLACE "-framework;" "-framework " ldflags_other "${${PKG_CONFIG_MODULE}_LDFLAGS_OTHER}")
+    list(APPEND link_libs ${ldflags_other})
+  endif()
+
+  # No need to treat the pkg-config module's *_LDFLAGS
+  # as they are already included in *_LINK_LIBRARIES, *_LIBRARY_DIRS, *_LIBRARIES and _LDFLAGS_OTHER
```
ruslo: So it was vice versa before, we was using `*_LDFLAGS` and ignoring `*_LINK_LIBRARIES`, `*_LIBRARY_DIRS` and `*_LIBRARIES`. According to the old comment `*_LINK_LIBRARIES`, `*_LIBRARY_DIRS` and `*_LIBRARIES` was part of the `*_LDFLAGS`. Would it be easier to keep old approach since we have to deal with one `*_LDFLAGS` instead of 3 variables?

ruslo: If it will not be possible please extend this comment with detail explanation.

cmake/modules/hunter_pkgconfig_export_target.cmake
Bjoe: Done

Bjoe:
> If this fix can be send as independent nonconfict patch then it's better to open new pull request for testing/review purposes.

I've tried to split this, but it has too many dependencies.

This approach search for libraries in the given path from the "pkg config"/*.pc file.and set the absolute path to the library for the linker to prevent that the linker to guess which library they should use.

For example, if an openssl is installed on the host machine, I see sometimes following linker output:
.... -lssl -lcrypto ..... -L/home/foo/.hunter/_Base/65c8f22/cc75a9d/09c9ea1/Install/lib ....
This looks like for me that the linker is using ssl and crypto libraries from the host machine.

cmake/modules/hunter_pkgconfig_export_target.cmake
Bjoe: Yep can be simplified ... but I added my second "fix request". So this line is changed and you will see why I cannot simplified.

cmake/modules/hunter_pkgconfig_export_target.cmake
Bjoe: I added my second "fix request". Hopefully now it is clear why I choose this approach.

cmake/projects/xcb/hunter.cmake
```diff
@@ -46,6 +46,6 @@ hunter_pick_scheme(DEFAULT xcb)
 hunter_cacheable(xcb)
 hunter_download(
     PACKAGE_NAME xcb
-    PACKAGE_INTERNAL_DEPS_ID "3"
+    PACKAGE_INTERNAL_DEPS_ID "7"
```
ruslo: "3" -> "7" looks strange, please use "4" here

cmake/projects/x264/hunter.cmake
```diff
@@ -25,6 +25,7 @@ hunter_cmake_args(x264 CMAKE_ARGS PKGCONFIG_EXPORT_TARGETS=x264)
 hunter_cacheable(x264)
 hunter_download(
     PACKAGE_NAME x264
+    PACKAGE_INTERNAL_DEPS_ID "4"
```
ruslo: Let's start with the "1"

cmake/projects/x11/hunter.cmake
```diff
@@ -41,7 +41,7 @@ hunter_cmake_args(
 hunter_cacheable(x11)
 hunter_download(
     PACKAGE_NAME x11
-    PACKAGE_INTERNAL_DEPS_ID "4"
+    PACKAGE_INTERNAL_DEPS_ID "8"
```
ruslo: May be "5"?

cmake/modules/hunter_pkgconfig_export_target.cmake
```diff
@@ -51,43 +51,95 @@ function(hunter_pkgconfig_export_target PKG_CONFIG_MODULE)
...
+    hunter_status_debug(
+      "Set target ${target_name} interface link librarise: ${link_libs}"
+    )
     set_target_properties("${target_name}"
         PROPERTIES
           INTERFACE_LINK_LIBRARIES
           "${link_libs}"
     )
-  endif()
+  endif()
```
ruslo: Please clean-up.

cmake/modules/hunter_pkgconfig_export_target.cmake
```diff
@@ -51,43 +51,95 @@ function(hunter_pkgconfig_export_target PKG_CONFIG_MODULE)
...
+  if(NOT "${${PKG_CONFIG_MODULE}_LDFLAGS_OTHER}" STREQUAL "")
+    # turn "-framework;A;-framework;B" into "-framework A;-framework B"
+    string(REPLACE "-framework;" "-framework " ldflags_other "${${PKG_CONFIG_MODULE}_LDFLAGS_OTHER}")
+    list(APPEND link_libs ${ldflags_other})
+  endif()
+
+  # No need to treat the pkg-config module's *_LDFLAGS
+  # as they are already included in *_LINK_LIBRARIES, *_LIBRARY_DIRS, *_LIBRARIES and *_LDFLAGS_OTHER
+
+  list(APPEND link_libs ${link_libs_release}) # TODO remove me!
```
ruslo: Please clean-up.

cmake/modules/hunter_pkgconfig_export_target.cmake
```diff
@@ -21,6 +21,11 @@ function(hunter_pkgconfig_export_target PKG_CONFIG_MODULE)
   if(TARGET "${target_name}")
     return()
   endif()
+  
+  # TODO We need to compare if XYZ_ROOT file path is changed to clear cache entries in CMakeCache.txt
+  
+  set(${PKG_CONFIG_MODULE}_FOUND 0 CACHE INTERNAL "") # TODO Force to read again xy.pc file
```
ruslo: This looks **very** suspicious. Please expain in details what is the problem.

cmake/projects/xcb/hunter.cmake
Bjoe: Well, I expected that I should for every change/commit increment the package internal id to execute test builds .... So I should only increment once in a branch and for every commit in the branch, test builds are executed?

cmake/projects/x264/hunter.cmake
Bjoe: See comment above https://github.com/ruslo/hunter/pull/1481#discussion_r220696618

cmake/projects/x11/hunter.cmake
Bjoe: See comment above https://github.com/ruslo/hunter/pull/1481#discussion_r220696618

Bjoe:
> >   .... -lssl -lcrypto ..... -L/home/foo/.hunter/_Base/65c8f22/cc75a9d/09c9ea1/Install/lib ....
> >   This looks like for me that the linker is using ssl and crypto libraries from the host machine.
>
> Are you sure? Looks good for me.
>
I will test again.

> >    I've tried to split this, but it has too many dependencies.
>
> At this moment this pull request is quite hairy, it's better to split improvements if possible. Will be much easier to test and review.
>
Ok, I fully cleanup this pull request/branch (forced pushed).
Now all should be fine ;-) ...

ruslo:
Testing.

x11:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.4441
* https://travis-ci.org/ingenue/hunter/builds/435666906?utm_source=github_status&utm_medium=notification

x264:
* https://travis-ci.org/ingenue/hunter/builds/435668429?utm_source=github_status&utm_medium=notification
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.4442

xcb:
* https://travis-ci.org/ingenue/hunter/builds/435669061?utm_source=github_status&utm_medium=notification
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.4443

ruslo:
Uploading.

x11:
* https://travis-ci.org/ingenue/hunter/builds/435680020?utm_source=github_status&utm_medium=notification
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.4444

x264:
* https://travis-ci.org/ingenue/hunter/builds/435680460?utm_source=github_status&utm_medium=notification
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.4445

xbc:
* https://travis-ci.org/ingenue/hunter/builds/435680620?utm_source=github_status&utm_medium=notification
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.4446

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.23.29

PR #1483, opened 2018-06-28
Fix hunter error message, WIKI parameter are missing

Bjoe:
There are some `hunter_fatal_error` without any `WIKI` arguments. I change these to `hunter_user_error`


ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.22.19

PR #1491, opened 2018-07-07
OpenBLAS: Don't use CMake on 0.3.0

xsacha:
The CMake build seems to provide an inferior build that causes an infinite loop, so fallback to normal build on 0.3.0.


ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.4173
* https://travis-ci.org/ingenue/hunter/builds/401307135?utm_source=github_status&utm_medium=notification

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.22.24

PR #1495, opened 2018-07-11
Fix OpenCV v3.2.0-p2 archive hash

rsnitsch:
See issue 'v3.2.0-p2 hash mismatch':
https://github.com/hunter-packages/opencv/issues/23


ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.22.26

PR #1506, opened 2018-07-27
Bump CLI11 version to 1.6.1, remove single file generation

henryiii:
I'm bumping the version, and also removing the single file generation since that will break systems that run Python 2.6 needlessly. See CLIUtils/CLI11#148 and ethereum-mining/ethminer#1392.

Not very familar with hunter, @chfast might be able to comment if there's a problem?

  * I don't see an `.appveyor.yml`, so this doesn't even try: https://ci.appveyor.com/project/HenrySchreiner/hunter
  * https://travis-ci.com/henryiii/hunter/jobs/136894146

---


henryiii:
Note: I've just released 1.6.1, which no longer tries to generate the single file version by default. So just updating to 1.6.1 would be enough to fix this problem.

henryiii:
I replaced the available version of CLI11 before, now they sit side-by-side like they are supposed to (Sorry, didn't realize that's how Hunter worked!). Also added a few more versions (latest 1.5.x, and 1.6.1).

ruslo:
> I've tested this package remotely and have excluded all broken builds

This job do documentation build, right way to test a package update is described here (for future):
* https://docs.hunter.sh/en/latest/creating-new/update.html#testing

Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.4268
* https://travis-ci.org/ingenue/hunter/builds/409795344?utm_source=github_status&utm_medium=notification

ruslo:
Uploading:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.4269
* https://travis-ci.org/ingenue/hunter/builds/409799411?utm_source=github_status&utm_medium=notification

PR #1489, opened 2018-07-05
Add OpenCV 3.4.2 and update config

xsacha:
Can't find the Visual Studio issue

PR #1531, opened 2018-08-31
Opencv 3.4.3

xsacha:

  * https://ci.appveyor.com/project/xsacha/hunter-h4uji/build/1.0.19
  * https://travis-ci.org/xsacha/hunter/builds/422878417


xsacha:
Oops. That's why it finished so quick. Will run correct test.

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/build/1.0.4383
* https://travis-ci.org/ingenue/hunter/builds/422878768?utm_source=github_status&utm_medium=notification

xsacha:
For

> CMake Error at cmake/android/OpenCVDetectAndroidSDK.cmake:30 (message):
>       Android SDK: specify path to Android SDK via ANDROID_SDK_ROOT /
>       ANDROID_HOME / ANDROID_SDK variables

Is it possible to set this environment variable first?

ruslo:
> Is it possible to set this environment variable first?

**If** Android SDK required then downloading of it should be managed by Hunter, see package `Android-SDK`:
* https://docs.hunter.sh/en/latest/packages/pkg/Android-SDK.html

But to build librararies we don't need Android SDK, I wonder if there is an option to disable it (such option can be introduced?).

xsacha:
Doesn't seem like that option works because I use it and it still checks for Android SDK.

ruslo:
> Doesn't seem like that option works because I use it and it still checks for Android SDK

OpenCV upstream bug?

xsacha:
Nevermind, it does work. You set BUILD_JAVA=OFF and it no longer checks.

xsacha:
Windows issue looks like problem with OpenCV_ROOT vs OpenCV_DIR? Started in CMake 3.12? Is that right?

ruslo:
> Windows issue looks like problem with OpenCV_ROOT vs OpenCV_DIR? Started in CMake 3.12? Is that right?

I think it's not, there was a problem with Windows before, see https://github.com/ruslo/hunter/pull/1489

cmake/projects/OpenCV/hunter.cmake
```diff
@@ -375,6 +386,7 @@ hunter_cmake_args(
     OpenCV
     CMAKE_ARGS
         BUILD_ANDROID_EXAMPLES=OFF
+        BUILD_JAVA=OFF
```
ruslo: Seems do the trick for Android toolchain:
* https://travis-ci.org/ingenue/hunter/builds/423312320

tonyrewin:
Not sure if it is related here or not, but now there is a sad limitation of OpenBLAS lib:
```sh
CMake Warning at CMakeLists.txt:41 (message):
  CMake support is experimental.  This will not produce the same Makefiles
  that OpenBLAS ships with.  Only x86 support is currently available.
```

Also it needs to set a TARGET env var, but [this](https://github.com/xianyi/OpenBLAS/blob/develop/TargetList.txt) target list is confising...

xsacha:
All architectures are actually supported. This is an old message.

TARGET is required if you are cross-compiling. The hunter script automatically chooses most common targets.

tonyrewin:
Is it compiling on mobiles without openBLAS? Because I did not manage to do that [yet] (https://github.com/ruslo/hunter/issues/1536)

ruslo:
> Otherwise I might have to do if/else on the version

Fine with me

xsacha:
I discovered the issue with Windows.
It assumed that if build_shared_libs is not set, then it must be a shared library.
However, the compile assumed that if build_shared_libs is not set, it must be a static library.

By defining build_shared_libs as either on or off, it compiles and discovers correctly.

`cmake .. -DCMAKE_BUILD_TYPE=Release -G"Visual Studio 15 2017 Win64" -DBUILD_SHARED_LIBS=OFF`

ruslo:
> So how come changing the OpenCV config to treat NOT DEFINED BUILD_SHARED_LIBS as static instead of shared works?

Your change consists of two parts, I guess the build fix is about second part, see https://github.com/hunter-packages/opencv/pull/26#discussion_r231402580

xsacha:
It's only the file  cmake/templates/OpenCVConfig.root-WIN32.cmake.in
The other one is unnecessary.

ruslo:
> It's only the file cmake/templates/OpenCVConfig.root-WIN32.cmake.in
> The other one is unnecessary.

> BUILD_SHARED_LIBS is always set by Hunter to ON or OFF

Okay, now it's clear then. `BUILD_SHARED_LIBS` set by Hunter only for third parties not for the local project. See comments: https://github.com/hunter-packages/opencv/pull/26#pullrequestreview-172373673

cmake/projects/OpenCV/hunter.cmake
```diff
@@ -11,7 +11,30 @@ include(hunter_cmake_args)
 include(hunter_download)
 include(hunter_pick_scheme)
 
+
```
ruslo: Please remove this

cmake/projects/OpenCV/hunter.cmake
```diff
@@ -11,7 +11,30 @@ include(hunter_cmake_args)
...
 # List of versions here...
+hunter_add_version(
+    PACKAGE_NAME
+    OpenCV
+    VERSION                                                     "3.4.3-p1"
```
ruslo: Please fix formatting

cmake/projects/OpenCV/hunter.cmake
```diff
@@ -11,7 +11,30 @@ include(hunter_cmake_args)
...
+3-p1.tar.gz"
+    SHA1
+    3589a7f19fdaef200091231c4255c5e5bb70b217
+)
+
+hunter_add_version(
+    PACKAGE_NAME
+    OpenCV
+    VERSION
+    "3.4.3-p0"
```
ruslo: If this version is not working properly then there is no need to keep it. Since it's not in master yet, it's okay to do a remove.

xsacha: Yes, I'll have to drop that commit as it still fails despite working locally.

-- OpenCV ARCH: x86
-- OpenCV RUNTIME: vc14
-- OpenCV STATIC: OFF

ruslo:
Closing for https://github.com/ruslo/hunter/pull/1613

PR #1639, opened 2018-12-02
Update 'lehrfempp'

craffael:

  * https://ci.appveyor.com/project/craffael/hunter/builds/20723639
  * https://travis-ci.org/craffael/hunter/builds/462560402

---

Update lehrfempp to 0.7.6


cmake/projects/lehrfempp/hunter.cmake
```diff
@@ -13,11 +13,11 @@ hunter_add_version(
     PACKAGE_NAME
     lehrfempp
     VERSION
-    0.7.3
+    0.7.5
```
ruslo: Please add new version, not update existent.

examples/lehrfempp/CMakeLists.txt
```diff
@@ -14,5 +14,5 @@ hunter_add_package(lehrfempp)
 find_package(lehrfempp CONFIG REQUIRED)
 
 add_executable(boo boo.cpp)
-target_link_libraries(boo PUBLIC lf::base)
+target_link_libraries(boo PUBLIC lf.base)
```
ruslo: Note that `foo::foo` target treated specially in CMake, unlike `foo.foo`. E.g. if `foo::foo` not found then CMake will emit an error, if `foo.foo` target is not found then CMake will try to link `-lfoo.foo`. First error will occur on generate step before compilation, second error will occur on link step, after compilation.

craffael: I wasn't aware of that. The reason that I opted for `lf.base` is because `lf::base` is also a namespace in the c++ project. Now at first `lf::base` also looks like the perfect cmake target. However the problem is that there are also third-level namespaces, e.g. there is a c++ namespace called `lf::mesh::hybrid2d`. As far as I know it is very complicated to create a cmake target with the name `lf::mesh::hybrid2d`, and a cmake target with the name `lf::mesh.hybrid2d` doesn't look right. Therefore I wanted to change `lf::` to `lf.`.
I think it's still a better solution to have `lf.mesh.hybrid2d` than `lf::mesh.hybrid2d` which is very inconsistent. But of course if the library is not found, the error only happens at link time.

cmake/projects/lehrfempp/hunter.cmake
craffael: I thought I would replace it because 0.7.3 is probably never going to be used. But I've added it now, just to stay consistent...

examples/lehrfempp/CMakeLists.txt
craffael: or would you decide differently? Thanks a lot for your suggestion!

ruslo: Was never thinking about two level namespacing. I'm trying usually do `foo::bar` C++ namespace -> `foo_bar`  in project target -> `foo::foo_bar` when added as external -> `<root>/lib/libfoo_bar.a` in install directory -> `FOO_BAR_EXPORT` macro for exports, example:
* https://github.com/cgold-examples/fruits/blob/master/lib/fruits/rosaceae/CMakeLists.txt
* https://github.com/cgold-examples/fruits/blob/master/lib/fruits/rosaceae/Pear.hpp

craffael: okay, I've changed it now to `LF::lf.base`. The problem with using underscores is that we follow the [google c++ style guide](https://google.github.io/styleguide/cppguide.html#Namespace_Names) which enforces underscores for composite namespace names.
The library is then called `<root>/lib/liblf.base.a`.

ruslo:
Testing:
* https://ci.appveyor.com/project/ingenue/hunter/builds/20724232
* https://travis-ci.org/ingenue/hunter/builds/462732938?utm_source=github_status&utm_medium=notification

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.23.62

Issue #1632, opened 2018-11-27
Add `ENTT` package


OvermindDL1:
* I'm interested in package `ENTT`
* Here is official site of package https://github.com/skypjack/entt
* This package is using Git: https://github.com/skypjack/entt.git
* Official GitHub repository: https://github.com/skypjack/entt
* Official read-only GitHub mirror: https://github.com/skypjack/entt.git* 
* Official GitHub Release Packages: https://github.com/skypjack/entt/releases
* Used build system: CMake
* This package has no dependencies other than Doxygen when generating docs, which should not be done when it is used as a dependency.

It is about to undergo massive changes before a 3.0 version bump, so the current v2.7.3 version at the very least (and any later) would be fantastic to include in Hunter.  The repo is very receptive to PR's that fix any CMake issues or things to help packaging systems.  The current release is built for C++14 but C++17 is the current 'master' branch with a few efficiency improvements and features that will become the default branch 'here soon' (this is planned to coincide with the 3.0 bump).  Overall it 'should' be fairly simple to include without forking it and if there are such issues than they accept tested PR's readily and quickly.

OvermindDL1:
Oh also, travis supports windows images now, have you thought of dropping the appveyor requirements and using just travis for consistency now?

skypjack:
Hi @OvermindDL1 

>The old 2.* series CMake installation is borked (no config file and such...) so that may need to be forked to be fixed unless @skypjack releases a new 2.* version that generates a correct CMake install command.

Fine for me, you know, I'm in to help but I didn't understand what's wrong with the last v2.  
If you can give me more details or if you know already what should be changed, I'm at your disposal.

>the current 3.x version of 3.0.0-WIP

It's a WIP version. There are still some things to add on `master` before to cut an RC but the last tag marks a stable version that we can use for this.

>Oh also, travis supports windows images now, have you thought of dropping the appveyor requirements and using just travis for consistency now?

I could be interested in this also for `EnTT` and `uvw`. Does it support vs2017 and C++17? The documentation is a bit unclear, at least for me.

OvermindDL1:
> Fine for me, you know, I'm in to help but I didn't understand what's wrong with the last v2.
> If you can give me more details or if you know already what should be changed, I'm at your disposal.

Sure, let me take a look at it again, oh found the issue, it was failing to compile its own tests:
```zsh
[ 30%] Building CXX object test/CMakeFiles/scheduler.dir/entt/process/scheduler.cpp.o
[...]

PR #1654, opened 2018-12-06
Add EnTT package

OvermindDL1:
* I've followed [this guide](https://docs.hunter.sh/en/latest/creating-new/create/cmake.html)
  step by step carefully. **[Yes]**

  * https://ci.appveyor.com/project/OvermindDL1/hunter/builds/20824727
  * https://travis-ci.org/OvermindDL1/hunter/builds/464526816

* I have submitted CI configs to https://github.com/ingenue/hunter targeting `pkg.template` branch,
  see this merged pull request https://github.com/ingenue/hunter/pull/302

* My change will resolve this "New package" request: https://github.com/ruslo/hunter/issues/1632


cmake/projects/EnTT/hunter.cmake
```diff
@@ -0,0 +1,29 @@
...
+include(hunter_cacheable)
+include(hunter_download)
+include(hunter_pick_scheme)
+include(hunter_cmake_args)
+
+hunter_add_version(
+    PACKAGE_NAME
+    EnTT
+    VERSION
+    3.0.0-WIP
```
ruslo: Please make it match git tag: `v3.x.y-WIP`

cmake/projects/EnTT/hunter.cmake
```diff
@@ -0,0 +1,29 @@
...
+    URL
+    "https://github.com/skypjack/entt/archive/v3.x.y-WIP.tar.gz"
+    SHA1
+    00f67f43060c1ea2f832e8bfb6d9c170dc84d5b1
+)
+
+hunter_cmake_args(
+	EnTT
```
ruslo: Indentation

examples/EnTT/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+
+# Emulate HunterGate:
+# * https://github.com/hunter-packages/gate
+include("../common.cmake")
+
+project(download-EnTT)
+
+# DOCUMENTATION_START {
+# EnTT requires at minimum C++17
+set(CMAKE_CXX_STANDARD 17)
```
ruslo: Should be set in toolchain:
* https://docs.hunter.sh/en/latest/overview/customization/toolchain-id.html#simple-toolchains

In Hunter CI testing some toolchains already have C++17 enabled. If lower is not supported then toolchain can be excluded.

examples/EnTT/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+set(CMAKE_CXX_STANDARD_REQUIRED on)
+
+hunter_add_package(EnTT)
+find_package(EnTT CONFIG REQUIRED)
+
+add_executable(foo foo.cpp)
+
+target_include_directories(foo
```
ruslo: Should it be imported target like `EnTT::EnTT`?

cmake/projects/EnTT/hunter.cmake
OvermindDL1: Done, pushing shortly

cmake/projects/EnTT/hunter.cmake
OvermindDL1: Done, pushing shortly

examples/EnTT/CMakeLists.txt
OvermindDL1: Done, pushing shortly

examples/EnTT/CMakeLists.txt
OvermindDL1: Sadly no, I tried various incantations there but the only thing that works was `${ENTT_INCLUDE_DIRS}`, is this a bug in EnTT's build script?  Does @skypjack need to do something and make another release (maybe of a fixed 2.7.y line instead)?

ruslo: > is this a bug in EnTT's build script? Does @skypjack need to do something and make another release (maybe of a fixed 2.7.y line instead)?

Imported target is a best practice, note that if `ENTT_INLUDE_DIRS` is a full path then package is not relocatable.

> And they do end up changing toolchain settings, should EnTT remove that from the CMakeLists.txt and advise to use an appropriate toolchain instead if they want such things set?

Toolchain is a requirement for Hunter, not for CMake in general. The logic is next: if you want to build with libc++ then you have to specify libc++ for all packages in set. It's easier to create toolchain with libc++ than create option for all packages. I don't know what `USE_COMPILE_OPTIONS` do, but I guess you can just add `USE_LIBCPP=OFF` to default `CMAKE_ARGS` and everything will be fine.

docs/packages/pkg/EnTT.rst
```diff
@@ -0,0 +1,38 @@
...
+
+.. _pkg.EnTT:
+
+EnTT
+====
+
+-  `Official <https://github.com/skypjack/entt>`__
+-  `Example <https://github.com/ruslo/hunter/blob/master/examples/EnTT/CMakeLists.txt>`__
+-  Developed by `Michele Caini <https://github.com/skypjack>`__
+-  Added by `Gabriel Robertson <https://github.com/overminddl1>`__ (`pr-1646 <https://github.com/ruslo/hunter/pull/1646>`__)
```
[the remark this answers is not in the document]
OvermindDL1: Fixed.  :-)

OvermindDL1:
_Can't respond to the comment directly for some reason..._

> Imported target is a best practice, note that if ENTT_INLUDE_DIRS is a full path then package is not relocatable.

@skypjack Is it a full path, or is there any way to get it to use the target name instead?

> Toolchain is a requirement for Hunter, not for CMake in general. The logic is next: if you want to build with libc++ then you have to specify libc++ for all packages in set. It's easier to create toolchain with libc++ than create option for all packages. I don't know what USE_COMPILE_OPTIONS do, but I guess you can just add USE_LIBCPP=OFF to default CMAKE_ARGS and everything will be fine.

@skypjack Is force disabling `USE_LIBCPP` fine here, is it okay otherwise if the user's toolchain specifies it instead?

skypjack:
> @skypjack Is it a full path, or is there any way to get it to use the target name instead?

It's generated as:
```
set_and_check(ENTT_INCLUDE_DIRS "@PACKAGE_CMAKE_INSTALL_INCLUDEDIR@")
```
That translates to:
```
set_and_check(ENTT_INCLUDE_DIRS "${PACKAGE_PREFIX_DIR}/include")
```
It was a contribution by @m-waka and I'm pretty sure he knows better than me what's going on under the hood here. :+1: 

> @skypjack Is force disabling `USE_LIBCPP` fine here, is it okay otherwise if the user's toolchain specifies it instead?

It's should be fine, of course.

---

Keep in mind that, in general, if you want to/know how to improve the build system of `EnTT`, I'm always open to suggestions. :wink:

OvermindDL1:
> That translates to:

Ah, so that is likely a fullpath.  @ruslo What does one need to do to support using something like `EnTT::EnTT` for it?  I've never made a header-only library so I'm unsure how to do it in this case...  >.>

> It's should be fine, of course.

Cool, added to the PR then!  :-)

I am curious what it's purpose is in the cmakefile,  it does seem like something *very* toolchain specific and not project specific at all?  Especially since https://github.com/skypjack/entt/blob/master/CMakeLists.txt#L51 is all it seems to do, which is not only toolchain specific but compiler specific as well?

> Keep in mind that, in general, if you want to/know how to improve the build system of EnTT, I'm always open to suggestions.

If anything needs to be done to fix it up that I learn here then I can submit a PR unless you beat me to it.  :-)

OvermindDL1:
> It was something due to the CI stuff if I'm not mistaken, but I must admit that I've forgot the reasons behind it. Maybe @m-waka can help use.

CI's should have their own customized toolchain files to define everything on them, I mostly use [polly](https://github.com/ruslo/polly)'s toolchains in my CI's.  :-)

ruslo:
> @ruslo What does one need to do to support using something like EnTT::EnTT for it?

I took a brief look and I think that the only missing piece is `NAMESPACE`, so we can do `EnTT::EnTT` instead of just `EnTT`.

Example:
* https://github.com/ruslo/leathers/blob/0cc869358faab455170c3aed93affb075a32cf40/CMakeLists.txt#L99
* https://github.com/ruslo/leathers/blob/0cc869358faab455170c3aed93affb075a32cf40/CMakeLists.txt#L133-L137

OvermindDL1:
@ruslo Oh that's significantly easier than what I've done in the past, didn't know it was a built-in option.  ^.^;

@skypjack Would you be for that being added to EnTT and pushing another 3.0.0 alpha release?  :-)

skypjack:
@OvermindDL1 from the links above, it seems that it doesn't affect in any way the current build system and users won't have problems due to this change, so... :+1:

ruslo:
> Is it also a best practice to use a CMake namespaces, even when there is only a single targets inside? It wouldn't be a big deal to do so.

Yes. If user made a mistake in name and will use `EnTt` instead of `EnTT` then CMake will try to use `-lEnTt` - we will have error on link stage (after configure and build). If user made a mistake in name with namespace `EnTt::EnTt` instead of `EnTT::EnTT` then CMake will emit error on configure stage (before build, before link).

OvermindDL1:
> Maybe it would be better to remove these CMake options completely and provide proper Toolchain CMake files for CI.

Instead of making your own you should use the [`polly`](https://github.com/ruslo/polly) set for your CI, it's what I use.  :-)

> Yes. If user made a mistake in name and will use EnTt instead of EnTT then CMake will try to use -lEnTt - we will have error on link stage (after configure and build). If user made a mistake in name with namespace EnTt::EnTt instead of EnTT::EnTT then CMake will emit error on configure stage (before build, before link).

Ah, great point!  I'll definitely keep that in mind!

skypjack:
@m-waka I definitely trust you for what concerns the build system. If the CI stuff accepts your changes, you're free to propose whatever you want by means of a PR.  
Can you just add a comment to the PR that describes how to obtain the same results as if one used `USE_LIBCPP`, `USE_ASAN` and `USE_COMPILE_OPTIONS`? This way, I can point it out to the users that will come asking for them, if any. Thanks!

OvermindDL1:
> Can you just add a comment to the PR that describes how to obtain the same results as if one used `USE_LIBCPP`, `USE_ASAN` and `USE_COMPILE_OPTIONS`? This way, I can point it out to the users that will come asking for them, if any. Thanks!

Those are also options that would effect someone's entire project, not just your code, that's a big big reason not to put it in your cmake files (since your project is all headers and no objects).  That is something that someone who is compiling their own code really need to be aware of and conscious about doing themselves, and if they know what libcpp/clang/asan is then they probably already have specialized toolchains for their system or already using something like polly.

m-waka:
@OvermindDL1 I see what you mean and I will 
1. Remove those options entirely
2. Adapt the CI configurations
As you recommend it in favor of writing an own toolchain file, what would be the best way to use **polly** with travis(using conan)/appveyor?

When I am ready I will open a PR for EnTT. I will let you know when it is merged.

skypjack:
@OvermindDL1 I used `polly` for long time (and contributed back to it - maybe @ruslo still remembers my first, annoying PRs when I didn't know well the project yet). Btw, it could be that one doesn't want to use it/include it in a project for something simple or completely unrelated - ie when I want to compile tests for `EnTT`. I won't use/don't think it's worth it to use a bazooka to kill a mosquito! Therefore a note aimed to help in a smooth transition could help. That's all. :+1: :smile:

OvermindDL1:
> As you recommend it in favor of writing an own toolchain file, what would be the best way to use polly with travis(using conan)/appveyor?

I just have it `git clone`'d into my docker image for my CI with a pull update on run (I use concourse instead of travis, but similar idea on travis, just git clone it in each time).

> Btw, it could be that one doesn't want to use it/include it in a project for something simple or completely unrelated - ie when I want to compile tests for EnTT. I won't use/don't think it's worth it to use a bazooka to kill a mosquito! Therefore a note aimed to help in a smooth transition could help. That's all. +1 smile

Something like polly shouldn't be included with any project ever anywhere, toolchains are system specific things, not project specific things, like on my desktop computer I have it cloned into `~/cpp/polly` and from any other project directory I just link to the toolchain I want in that directory when I `cmake -H. -Bbuild.debug -D...` to generate the build files, whether I leave it off for global gcc usage or I link to it to use clang or build for android or cross-compile to windows or so.  The local system is what should be aware about toolchains, not a project (at most it should test for available features it might need or optimize for).  :-)

ruslo:
> Btw, it could be that one doesn't want to use it/include it in a project for something simple or completely unrelated - ie when I want to compile tests for EnTT. I won't use/don't think it's worth it to use a bazooka to kill a mosquito! 

The only requirement for Hunter is that project should not touch flags:
* https://docs.hunter.sh/en/latest/creating-new/create/cmake.html#default-behavior

Because toolchain used to specify global flags:
* https://docs.hunter.sh/en/latest/overview/customization/toolchain-id.html

 It's not general CMake rule, it's Hunter specific rule. However you may find it very convenient.

So adding `USE_LIBCPP=OFF` to default `CMAKE_ARGS` (on Hunter side, not on project side) will do the trick. The rest is up to you (Polly, toolchains, CI, etc.).

skypjack:
Uhm... interesting approach. :+1: Let me think twice if it fits with my ways or not before to switch.  
So far, I always included `polly` as part of the projects in which I used it (mostly mobile stuff). Reasons were mainly that those projects are company wide ones and that was much simpler than telling to all the developers they have to set toolchains and bla bla bla. :smile:  
But I see what you mean, good point.

m-waka:
@OvermindDL1 I am not really convinced, of course you can have your custom Toolchain files as part of your project, containing the flags you optimized your library with e.g. (Nobody is forced to use it) But I also would'nt include the whole polly project within my project.

OvermindDL1:
> I am not really convinced, of course you can have your custom Toolchain files as part of your project, containing the flags you optimized your library with e.g. (Nobody is forced to use it) But I also would'nt include the whole polly project within my project.

Oh if you want to include custom toolchain files then by all means include those with the project, that's an "expected base set", but those are toolchain specific still, where if you want to compile for, say, android, or cross-compile to windows, then you need other toolchains for those, whether those are included in the project as well or not.  The user compiling it can still copy and override values in those as they wish for their own platform.  Regardless, those still don't belong in the CMakeLists.txt files as hard setting some flags on the toolchain in those suddenly means you fail when the user has to use a toolchain that you didn't anticipate that doesn't support those flags that are being added.

If a project is designed to be built for android for example, I'd absolutely see including android toolchain files in the `cmake/toolchains` directory in the project as that is the 'expected compile type' of the library (just as always make sure they don't make assumptions about the host OS paths or versions or so!), though personally I still wouldn't as there are so many versions of the android SDK and it's better just to feature check for the features needed and let the host OS use it's own toolchains that know the locations and setup of the android sdk for everything far more optimized than a generic script could do (and there are plenty of generic scripts out anyway if anyone needs some).

ruslo:

OvermindDL1:
For note, EnTT is still pending it's new version release, so later still.

skypjack:
Oh, @OvermindDL1 was that blocking this issue? I didn't get it, sorry, I thought v2 was fine to start with.

OvermindDL1:
That was the blocking issue yeah, v2's CMake build system was a little bit too broken to be able to use it here without either forking it or porting the changes backwards.

PR #1701, opened 2019-01-09
Force hard link when unpacking dir

bkotzz:
When unpacking the directory, there is a step where hunter creates a hard link from files in your Cellar directory, to files in your Install directory for the toolchain+config combination of that build run. The hard link will fail if the files already exist in the install directory.

I encountered this error when building protobuf 3.6.1, as hunter attempted to build protobuf from source (cache miss), but the files in the Install folder already existed.

```
[hunter ** INTERNAL **] Link script failed: 1, , Link files using Python: /usr/bin/python3
CMake Warning at /home/ubuntu/monorepo/hunter/hunter/scripts/link-all.cmake:70 (message):
  Python script failed:
  /usr/bin/python3
[hunter ** INTERNAL **] /home/ubuntu/monorepo/hunter/hunter/scripts/link-all.py
[hunter ** INTERNAL **] --list
[hunter ** INTERNAL **] /home/ubuntu/monorepo/hunter/hunter/_Base/Cellar/8536507487ca468150322f46f4147bd881d8f721/8536507/files.list
[hunter ** INTERNAL **] --cellar
[hunter ** INTERNAL **] /home/ubuntu/monorepo/hunter/hunter/_Base/Cellar/8536507487ca468150322f46f4147bd881d8f721/8536507/raw
[hunter ** INTERNAL **] --dest
[hunter ** INTERNAL **] /home/ubuntu/monorepo/hunter/hunter/_Base/xxxxxxx/b6a0ae2/9bcf801/Install,
  1, Exception caught: [Errno 5] Input/output error:
  '/home/ubuntu/monorepo/hunter/hunter/_Base/Cellar/8536507487ca468150322f46f4147bd881d8f721/8536507/raw/bin/protoc'
  ->
  '/home/ubuntu/monorepo/hunter/hunter/_Base/xxxxxxx/b6a0ae2/9bcf801/Install/bin/protoc'


  , Some job failed

  (may help: https://stackoverflow.com/a/2009505/2288008)


Link files using shell: /bin/bash
CMake Error at /home/ubuntu/monorepo/hunter/hunter/scripts/link-all.cmake:92 (message):
  Shell script failed:
  /bin/bash
[hunter ** INTERNAL **] /home/ubuntu/monorepo/hunter/hunter/_Base/Cellar/8536507487ca468150322f46f4147bd881d8f721/8536507/link-all.sh
[hunter ** INTERNAL **] /home/ubuntu/monorepo/hunter/hunter/_Base/xxxxxxx/b6a0ae2/9bcf801/Install,
  1, , ln: failed to create hard link
  '/home/ubuntu/monorepo/hunter/hunter/_Base/xxxxxxx/b6a0ae2/9bcf801/Install/bin/protoc':
  File exists
```

I'm not sure why it was in this state, but a simple fix is to force the hard link.

---

PTAL @ruslo

### DIFF
--- a/cmake/modules/hunter_unpack_directory.cmake
+++ b/cmake/modules/hunter_unpack_directory.cmake
@@ -101,7 +101,7 @@ function(hunter_unpack_directory cache_sha1)
       file(
           APPEND
           "${shell_link_script}"
-          "ln \\\n  \"\${HUNTER_CELLAR_RAW_DIRECTORY}/${x}\" \\\n  \"\$1/${x}\"\n\n"
+          "ln -f \\\n  \"\${HUNTER_CELLAR_RAW_DIRECTORY}/${x}\" \\\n  \"\$1/${x}\"\n\n"
        )
     endforeach()
     # }


ruslo:
> I'm not sure why it was in this state

It may happen if you interrupt link process (leaving it only as a partly finished) and then run build again.

> /.../hunter/_Base/**xxxxxxx**/b6a0ae2/9bcf801/Install/bin/protoc

Means you're using Hunter from Git, without packing it to archive. Use it with double care, see https://github.com/ruslo/hunter/wiki/dev.hunter.root#jenkinspy

ruslo:
> I still don’t yet understand the root cause

Like I said the only explanation I can think of is the interruption of script with link command, is it possible that at some point you stop the build with Ctrl+C? `xxxxxxx` looks suspicious in general, it's not robust and if you did too much magic it may cause troubles :)

> Here are some logs I captured when building protobuf 3.6 from hunter (on the left side of the diff) and protobuf 3.5 from hunter (on the right side of the diff)
https://www.diffchecker.com/ifRtXNjQ. It works with 3.5 but fails with 3.6. You can see that 3.6 has extra logs at the bottom, as it is doing the link-all step that 3.5 did not. Any thoughts?

Log from the right side of the diff is cut, I do expect that it will continue with the script with `ln` commands. Anyway the more interesting question is the state of directory `/home/ubuntu/monorepo/hunter/hunter/_Base/xxxxxxx/0498cbe/9bcf801/Install`. Do we have files there before Hunter instructions triggered?

As an example
```
  1, Exception caught: [Errno 5] Input/output error:		
484	  '/home/ubuntu/monorepo/hunter/hunter/_Base/Cellar/82020cb165e6b862e171ba98cf251a64d246f856/82020cb/raw/bin/protoc'		
485	  ->		
486	  '/home/ubuntu/monorepo/hunter/hunter/_Base/xxxxxxx/0498cbe/9bcf801/Install/bin/protoc'
```
Do `/home/ubuntu/monorepo/hunter/hunter/_Base/xxxxxxx/0498cbe/9bcf801/Install/bin/protoc` file exist before Hunter build? Do `/home/ubuntu/monorepo/hunter/hunter/_Base/xxxxxxx/0498cbe/9bcf801/Install/bin` directory exist? May be some package wrongly install `/home/ubuntu/monorepo/hunter/hunter/_Base/xxxxxxx/0498cbe/9bcf801/Install/bin` as a file and we can't use it as a directory?

bkotsopoulossc:
Nope, there was no interruption of the script with CTRL-C, both build runs finished. And even though I did cut out the logs in the diff by only showing the protobuf part, there was no more logs for protobuf after that. So the right side of the diff did not ever run the same link step, the build just went on to the next hunter dependency. 

And these logs were captured after deleting the entire _Base directory before running the build, so I am sure that the files and folders did not exist before the build, but the bin/protoc file and the others existed at the point where the error happened. 

It is also weird that this error only happens with the certain version of Protobuf

ruslo:
> the right side of the diff did not ever run the same link step, the build just went on to the next hunter dependency

I noticed that you haven't used `HUNTER_STATUS_DEBUG=ON` so there should be no logs even if link script run.

> but the bin/protoc file and the others existed at the point where the error happened

Some package do install same files?

> It is also weird that this error only happens with the certain version of Protobuf

Can you try to reproduce error? Please use `HUNTER_STATUS_DEBUG=ON` and don't use git based Hunter version (check there is no `/xxxxxxx/` in path to Hunter install dir, like `/_Base/xxxxxxx/0498cbe/9bcf801/`).

Issue #388, opened 2016-04-17
Add `python` package


ruslo:
Official:
* https://www.python.org/

Take a look:
* https://github.com/python-cmake-buildsystem/python-cmake-buildsystem

PR #1710, opened 2019-01-15
modify PYTHON_VERSION processing for MSVC compatibility

Algomorph:
1) Use standard FindPythonLibs instead of getting library location from interpreter,
which is more semantically correct and doesn't break on Windows
2) Replace single back- and forward-slashes in python paths of user_config.jam file with
escaped backslashes for MSVC only (otherwise, build breaks on MSVC)
3) Use PYTHON_ADD_MODULE command in Boost-python
4) Fix MSVC 2015 | Windows SDK 8.1 toolchain build (Derive include paths for Python from result of FindPythonLibs instead of using Python interpreter directly)
5) Update Boost package index to 39

  step by step carefully. **[Yes]**

* I've tested this package remotely and have excluded all broken builds (Edit: yep, all passed!).

  * https://ci.appveyor.com/project/Algomorph/hunter/builds/21647247
  * https://travis-ci.com/Algomorph/hunter/builds/97463949

---

### DIFF
--- a/examples/Boost-python/CMakeLists.txt
+++ b/examples/Boost-python/CMakeLists.txt
@@ -20,11 +20,12 @@ hunter_add_package(Boost COMPONENTS python)
 find_package(Boost CONFIG REQUIRED python35)
 find_package(PythonLibs ${PYTHON_VERSION} EXACT REQUIRED)
 
-add_library(foo foo.cpp)
+PYTHON_ADD_MODULE(foo foo.cpp)
 target_link_libraries(
     foo
     PUBLIC
     Boost::python35
+    ${PYTHON_LIBRARIES}
 )
 target_include_directories(
   foo
--- a/cmake/projects/Boost/atomic/hunter.cmake
+++ b/cmake/projects/Boost/atomic/hunter.cmake
@@ -18,5 +18,5 @@ hunter_download(
     Boost
     PACKAGE_COMPONENT
     atomic
-    PACKAGE_INTERNAL_DEPS_ID "39"
+    PACKAGE_INTERNAL_DEPS_ID "40"
 )
--- a/cmake/projects/Boost/chrono/hunter.cmake
+++ b/cmake/projects/Boost/chrono/hunter.cmake
@@ -18,5 +18,5 @@ hunter_download(
     Boost
     PACKAGE_COMPONENT
     chrono
-    PACKAGE_INTERNAL_DEPS_ID "39"
+    PACKAGE_INTERNAL_DEPS_ID "40"
 )
--- a/cmake/projects/Boost/context/hunter.cmake
+++ b/cmake/projects/Boost/context/hunter.cmake
@@ -18,5 +18,5 @@ hunter_download(
     Boost
     PACKAGE_COMPONENT
     context
-    PACKAGE_INTERNAL_DEPS_ID "39"
+    PACKAGE_INTERNAL_DEPS_ID "40"
 )
--- a/cmake/projects/Boost/contract/hunter.cmake
+++ b/cmake/projects/Boost/contract/hunter.cmake
@@ -18,5 +18,5 @@ hunter_download(
     Boost
     PACKAGE_COMPONENT
     contract
-    PACKAGE_INTERNAL_DEPS_ID "39"
+    PACKAGE_INTERNAL_DEPS_ID "40"
 )
--- a/cmake/projects/Boost/coroutine/hunter.cmake
+++ b/cmake/projects/Boost/coroutine/hunter.cmake
@@ -18,5 +18,5 @@ hunter_download(
     Boost
     PACKAGE_COMPONENT
     coroutine
-    PACKAGE_INTERNAL_DEPS_ID "39"
+    PACKAGE_INTERNAL_DEPS_ID "40"
 )
--- a/cmake/projects/Boost/date_time/hunter.cmake
+++ b/cmake/projects/Boost/date_time/hunter.cmake
@@ -18,5 +18,5 @@ hunter_download(
     Boost
     PACKAGE_COMPONENT
     date_time
-    PACKAGE_INTERNAL_DEPS_ID "39"
+    PACKAGE_INTERNAL_DEPS_ID "40"
 )
--- a/cmake/projects/Boost/exception/hunter.cmake
+++ b/cmake/projects/Boost/exception/hunter.cmake
@@ -18,5 +18,5 @@ hunter_download(
     Boost
     PACKAGE_COMPONENT
     exception
-    PACKAGE_INTERNAL_DEPS_ID "39"
+    PACKAGE_INTERNAL_DEPS_ID "40"
 )
--- a/cmake/projects/Boost/fiber/hunter.cmake
+++ b/cmake/projects/Boost/fiber/hunter.cmake
@@ -18,5 +18,5 @@ hunter_download(
     Boost
     PACKAGE_COMPONENT
     fiber
-    PACKAGE_INTERNAL_DEPS_ID "39"
+    PACKAGE_INTERNAL_DEPS_ID "40"
 )
--- a/cmake/projects/Boost/filesystem/hunter.cmake
+++ b/cmake/projects/Boost/filesystem/hunter.cmake
@@ -18,5 +18,5 @@ hunter_download(
     Boost
     PACKAGE_COMPONENT
     filesystem
-    PACKAGE_INTERNAL_DEPS_ID "39"
+    PACKAGE_INTERNAL_DEPS_ID "40"
 )
--- a/cmake/projects/Boost/graph/hunter.cmake
+++ b/cmake/projects/Boost/graph/hunter.cmake
@@ -18,5 +18,5 @@ hunter_download(
     Boost
     PACKAGE_COMPONENT
     graph
-    PACKAGE_INTERNAL_DEPS_ID "39"
+    PACKAGE_INTERNAL_DEPS_ID "40"
 )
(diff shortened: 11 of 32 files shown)


examples/Boost-python/CMakeLists.txt
```diff
@@ -20,7 +20,7 @@ hunter_add_package(Boost COMPONENTS python)
 find_package(Boost CONFIG REQUIRED python35)
 find_package(PythonLibs ${PYTHON_VERSION} EXACT REQUIRED)
 
-add_library(foo foo.cpp)
+PYTHON_ADD_MODULE(foo foo.cpp)
```
ruslo: Same question as [here](https://github.com/ruslo/hunter/pull/1708#discussion_r247857654): have you tried to build this example on Windows?

It still doesn't work for me, this time link error:
```
LINK : fatal error LNK1104: cannot open file 'python34.lib'
```

Algomorph: Yes, I have built this example on two different Windows platforms, successfully, multiple times, and I tried with the toolchain you suggested (vs-14-2015-sdk-8-1) as well, it also works for me.

Appveyor MSVC tests all seem to also succeed:
https://ci.appveyor.com/project/Algomorph/hunter/builds/21647247

Algomorph: My question would be, in your case, "why is it trying to link to python34.lib in the first place?", since the example is for python 3.5, not 3.4. The call to find PythonLibs is as follows:
https://github.com/ruslo/hunter/blob/eeed387c9f28bac06b25eb84ea5d9d3ee798740d/cmake/projects/Boost/schemes/url_sha1_boost_library.cmake.in#L306

Since "EXACT" is specified and we're specifying 3.5, not 3.4, something seems inconsistent here.

ruslo: > Yes, I have built this example on two different Windows platforms, successfully, multiple times, and I tried with the toolchain you suggested (vs-14-2015-sdk-8-1) as well, it also works for me.

Have you used `jenkins.py` and `Boost-python` example?

> Appveyor MSVC tests all seem to also succeed:
https://ci.appveyor.com/project/Algomorph/hunter/builds/21647247

There is no Boost-python example there, it just proves that old functionality was not broken (which is a good start).

I can reproduce the problem on AppVeyor (Python 3.4, `vs-14-2015-win64-sdk-8-1`, `Boost-python`), see:
- https://ci.appveyor.com/project/ingenue/hunter/builds/22119665

> why is it trying to link to python34.lib in the first place?

I have Python 3.4 installed on my machine, so tweaked example accordingly.

Algomorph: It seems like the one thing that's different from all environments I've tested in is the specific python version 3.4. I'm going to see if I can replicate the setup with Python 3.4 and vs-14-2015-win64-sdk-8-1 on my Windows machines.

Algomorph: I'm having a lot of trouble with Python 3.4. Tried it on two machines, but I'm getting "Python Interpreter for Python version 3.4 not found." Will investigate further.

Algomorph: Ok, boiling down to potential bug with FindPythonInterp on Windows...
`find_package(PythonInterp 3.4 EXACT REQUIRED QUIET)` throws an error with an installation of Python 3.4 when Python 3.6 is already installed. Gonna try to find workaround first or uninstall 3.6 and see how it goes.

Algomorph: (NB: Workaround was removing Python 3.6 from Path and adding 3.4, apparently. I didn't expect the FindPythonInterp script to be so primitive.)

Algomorph: @ruslo, I am unable to replicate the bug so far locally with 3.4. Everything goes smoothly for me after the small hiccup with the FindPythonInterp. I guess I'll keep trying (maybe, newer CMake version), but any insight as to the difference between our setups (and difference in setup in AppVeyor from an ordinary Windows workstation) would be appreciated.

image

Algomorph: > Have you used `jenkins.py` and `Boost-python` example?

And also, how would I do that?

Algomorph: (3) if it's only Python 3.4 that's the issue, I propose to include a note in the documentation about the PYTHON_VERSION parameter -- that it doesn't support Python 3.4, merge the PR, and open a separate issue about that (I am fairly certain that only a small percentage of Hunter users are using Windows machines running a Python minor release from 2014).

ruslo: > Workaround was removing Python 3.6 from Path and adding 3.4, apparently. I
> didn't expect the FindPythonInterp script to be so primitive

The path to tools is not standardized on Windows, how do you expect it to find Python without looking to PATH?

> > Have you used `jenkins.py` and `Boost-python` example?

> And also, how would I do that?

Really, just now!? It should be your main tool for testing package, see:
- https://docs.hunter.sh/en/latest/creating-new/create/cmake.html#testing-locally

Algomorph: > The path to tools is not standardized on Windows, how do you expect it to find Python without looking to PATH?

I would expect it to look up Python install path in the registry, either at: 
`[HKEY_LOCAL_MACHINE\SOFTWARE\Python\PythonCore\${VERSION}\InstallPath]`         
or at:
`[HKEY_CURRENT_USER\SOFTWARE\Python\\PythonCore\${VERSION}\InstallPath]`

The path to python.exe within and the path to the "Scripts" folder shouldn't vary too much from version to version, and there are only so many releases to take into account. Simple checks could be done to ensure this is a standard installation. My hypothesis is that an overwhelming majority of Windows python users use the official installers provided on Python.org and don't change the directory structure for this not to work.

Algomorph: > Really, just now!? It should be your main tool for testing package, see:
> 
> * https://docs.hunter.sh/en/latest/creating-new/create/cmake.html#testing-locally

I don't see how what I'm doing manually via CMake + polly is in any way different than what I would be doing with jenkins.py. I cannot fix the bug if I cannot replicate it. If you think jenkins.py can help me replicate the bug, then I'll take a look into that, but that scenario would imply there is a difference between how the testing is set up and how Hunter users on Windows are going to be using the package.

ruslo: > I would expect it to look up Python install path in the registry, either at:

You can send a patch to CMake, this functionality is not part of Hunter. Just for your info you can add Python to Hunter and manage version with Hunter, see: https://github.com/ruslo/hunter/issues/388

>I don't see how what I'm doing manually via CMake + polly is in any way different than what I would be doing with jenkins.py

It's much easier to make a mistake. The way you have used [here](https://github.com/ruslo/hunter/pull/1710#discussion_r256818916) for testing was kept for compatibility and designed for advanced usage, I will remove it shortly.

> that scenario would imply there is a difference between how the testing is set up and how Hunter users on Windows are going to be using the package

No, it's not, it's pretty similar.

> I am unable to replicate the bug so far locally with 3.4

I can reproduce the issue on my machine and on AppVeyor machine. You can check AppVeyor machine if you believe that my setup is broken in any way. Also assuming that you haven't used `jenkins.py` for testing my advice will be to start the build from scratch with `jenkins.py --clear-except-download` and make sure all packages built from the sources with latest instructions from Hunter.

Algomorph: Ok, I will do a jenkins.py build and report back.

Algomorph: I was able to reproduce the bug with the Jenkins build. I'll try to figure out what's different between the jenkins.py build and what I'm doing manually with CMake+polly, maybe that will shed some light on this issue. My first guess is that a different Boost configuration is somehow being used to build the binaries, which causes some problems for MSVC. I'll look into it.


(Error: `fatal error LNK1104: cannot open file 'python34.lib' [C:\Users\Algomorph\Code\hunter\_testing\_builds\vs-14-2015-win64-sdk-8-1\foo.vcxproj`)

Algomorph: NB: Using the same CMake cache file from within CMake _directly_ results in a successful configuration.
This means the jenkins.py build does something beyond just running CMake which I'm not expecting. Any ideas what?
image

Algomorph: Ok, for some reason it wasn't able to auto-link python libraries, not sure still what the difference was. But including ${PYTHON_LIBRARIES} in target_link_libraries solved the issue (commit submitted).

Algomorph:
> Please rebase your changes.

Done.

ruslo:
> > Please rebase your changes.

> Done.

I think you have squashed it, not [rebased](https://git-scm.com/book/en/v2/Git-Branching-Rebasing).

Algomorph:
Oh, true, my bad, I misunderstood. I guess you wanted me to rebase on master. One minute.

Algomorph:
Ok, I think I got it right now.

ruslo:
> Ok, I think I got it right now.

You've lost `PACKAGE_INTERNAL_DEPS_ID`. It was 38 -> 39 change but while you working on it 39 was used [here](https://github.com/ruslo/hunter/pull/1746/files). Please do 39 -> 40 for this change.

Algomorph:
Ok. Anything else?

ruslo:
Testing:
- https://ci.appveyor.com/project/ingenue/hunter/builds/22758711
- https://travis-ci.org/ingenue/hunter/builds/500520748?utm_source=github_status&utm_medium=notification

ruslo:
Uploading:
- https://ci.appveyor.com/project/ingenue/hunter/builds/22766149
- https://travis-ci.org/ingenue/hunter/builds/500676193?utm_source=github_status&utm_medium=notification

ruslo:
Thanks for the fix, released: https://github.com/ruslo/hunter/releases/tag/v0.23.124

Issue #751, opened 2017-04-05
CMAKE_VS_DEVENV_COMMAND is empty


alienzj:
1> Command line: C:\PROGRAM FILES (X86)\MICROSOFT VISUAL STUDIO\2017\COMMUNITY\COMMON7\IDE\COMMONEXTENSIONS\MICROSOFT\CMAKE\CMake\bin\cmake.exe  -G "Visual Studio 15 2017" -DCMAKE_INSTALL_PREFIX:PATH="C:\Users\zhujie\AppData\Local\CMakeBuild\877c8177-4761-4033-89b6-882311601906\build\install"  --DCMAKE_BUILD_TYPE="Debug" "D:\projects\metapp"
1> Working directory: C:\Users\zhujie\AppData\Local\CMakeBuild\877c8177-4761-4033-89b6-882311601906\build\x86-Debug
1> -- [hunter] Calculating Config-SHA1
1> -- [hunter] Calculating Toolchain-SHA1
1> -- [hunter] HUNTER_ROOT: D:/program/hunter_root
1> -- [hunter] [ Hunter-ID: 50eaca2 | Config-ID: a9b8ca2 | Toolchain-ID: 5ae8cae ]
1> 
1> [hunter ** INTERNAL **] CMAKE_VS_DEVENV_COMMAND is empty
1> [hunter ** INTERNAL **] [Directory:D:/projects/metapp]
1> 
1> ------------------------------ WIKI -------------------------------
1>     https://github.com/ruslo/hunter/wiki/error.internal
1> -------------------------------------------------------------------
1> 
1> CMake Error at D:/program/hunter_root/_Base/Download/Hunter/0.18.36/50eaca2/Unpacked/cmake/modules/hunter_wiki.cmake:12 (message):
1> Call Stack (most recent call first):
1>   D:/program/hunter_root/_Base/Download/Hunter/0.18.36/50eaca2/Unpacked/cmake/modules/hunter_internal_error.cmake:13 (hunter_wiki)
1>   D:/program/hunter_root/_Base/Download/Hunter/0.18.36/50eaca2/Unpacked/cmake/modules/hunter_setup_msvc.cmake:117 (hunter_internal_error)
1>   D:/program/hunter_root/_Base/Download/Hunter/0.18.36/50eaca2/Unpacked/cmake/modules/hunter_finalize.cmake:104 (hunter_setup_msvc)
[...]

PR #1729, opened 2019-02-04
msvc: Support Microsoft Build Tools 2017

hjmallon:
Microsoft Build Tools 2017 is a distribution of the Visual Studio compiler without the UI. It doesn't have devenv but cmake still works using msbuild. With the 2017 version no ENV variables are installed to find vcvarsall but we can use the path to MSBuild to find it (like we can with Devenv in Visual Studio proper).

I was able to reproduce the problem using:
`TOOLCHAIN=vs-15-2017-win64-cxx17 PROJECT_DIR=examples/ZLIB/ ./jenkins.py`

Possibly wording on the devenv wiki page needs changing to add reference to MSBuild?

This issue was mentioned by me here: https://github.com/ruslo/hunter/issues/751#issuecomment-460349156

### DIFF
--- a/cmake/modules/hunter_setup_msvc.cmake
+++ b/cmake/modules/hunter_setup_msvc.cmake
@@ -115,6 +115,9 @@ macro(hunter_setup_msvc)
     hunter_status_debug(
         "CMAKE_VS_DEVENV_COMMAND: '${CMAKE_VS_DEVENV_COMMAND}'"
     )
+    hunter_status_debug(
+        "CMAKE_VS_MSBUILD_COMMAND: '${CMAKE_VS_MSBUILD_COMMAND}'"
+    )
 
     string(COMPARE EQUAL "${_hunter_vcvarsall_path}" "" _is_empty)
     if(_is_empty)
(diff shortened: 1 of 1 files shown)


ruslo:
> Possibly wording on the devenv wiki page needs changing to add reference to MSBuild?

Feel free to add any useful info there.

Released: https://github.com/ruslo/hunter/releases/tag/v0.23.96

cmake/modules/hunter_setup_msvc.cmake
```diff
@@ -115,32 +115,45 @@ macro(hunter_setup_msvc)
...
+        if(NOT is_devenv_empty AND IS_ABSOLUTE "${CMAKE_VS_DEVENV_COMMAND}")
+          get_filename_component(_hunter_vcvarsall_path
+              "${CMAKE_VS_DEVENV_COMMAND}" DIRECTORY
           )
-        endif()
-        if(NOT IS_ABSOLUTE "${CMAKE_VS_DEVENV_COMMAND}")
+          set(_hunter_vcvarsall_path
+              "${_hunter_vcvarsall_path}/../../VC/Auxiliary/Build"
+          )
+        elseif(is_msbuild_empty AND IS_ABSOLUTE "${CMAKE_VS_MSBUILD_COMMAND}")
```
ruslo: `NOT` is missing? It should be
```cmake
elseif(NOT is_msbuild_empty AND IS_ABSOLUTE "${CMAKE_VS_MSBUILD_COMMAND}")
```
?

hjmallon: Yes I think it should match the Devenv case. Sorry about that.

ruslo: Fix: https://github.com/ruslo/hunter/commit/556dcfea6b4b8051404de4eaac897ef5086c96e0

Released: https://github.com/ruslo/hunter/releases/tag/v0.23.110

> Yes I think it should match the Devenv case. Sorry about that.

> we can use the path to MSBuild to find it

Well, I was thinking it's exactly the branch that do the fix for you. Have you tested it?

`if(NOT is_devenv_empty AND IS_ABSOLUTE "${CMAKE_VS_DEVENV_COMMAND}")` branch used on my local machine and in AppVeyor testing, that's why bug wasn't caught.

hjmallon: You are right. I negated the check to check the error message case, then accidentally pushed the result. I have sent you a beer via PayPal by way of apology.

PR #1819, opened 2019-04-15
Add 2.3.4-p0 version for utf8 with default header location

sjdrc:

I have not done CI testing for this since the change in the package is only moving the install directory for includes.

see: https://github.com/hunter-packages/utf8/commit/fd445050845b6bce1c94fd20dcce9afff38f8b44

---


ruslo:
> I have not done CI testing for this since the change in the package is only moving the install directory for includes.

And why exactly you decide that tests not needed in this case?

Update instructions are here:
- https://docs.hunter.sh/en/latest/creating-new/update.html

Closing.

PR #1852, opened 2019-04-29
jenkins: allow relative POLLY_ROOT paths

NeroBurner:

---

### DIFF
--- a/jenkins.py
+++ b/jenkins.py
@@ -111,6 +111,7 @@ def run():
   os.makedirs(testing_dir, exist_ok=True)
 
   if os.name == 'nt':
+    # path too long workaround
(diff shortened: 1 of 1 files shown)


jenkins.py
```diff
@@ -143,7 +144,7 @@ def run():
   else:
     which = 'which'
 
-  polly_root = os.getenv('POLLY_ROOT')
+  polly_root = os.path.abspath(os.getenv('POLLY_ROOT'))
```
ruslo: Should be checked for `None`:
- https://github.com/ruslo/hunter/commit/9e43e66c172663c5f4234549be75c9a6c9255bf4

See:
- https://travis-ci.org/ingenue/hunter/jobs/531915099

NeroBurner: Thanks!

PR #1862, opened 2019-05-11
Cxx20

madmongo1:
* My change will work with CMake 3.2 (minimum requirement for Hunter). **[No]**

---

This change adds c++20 support, which is now supported by latest CMake


cmake/modules/hunter_get_lang_standard_flag.cmake
```diff
@@ -4,7 +4,7 @@
 include(hunter_status_debug)
 
 function(hunter_get_lang_standard_flag LANG OUTPUT)
-  set(CXX_standards 17 14 11 98)
+  set(CXX_standards 20 17 14 11 98)
```
ruslo: Applied: https://github.com/ruslo/hunter/commit/355a7aeb120a5a92dc06529435d4d292a60b5f22

cmake/projects/SocketIOClient/hunter.cmake
```diff
@@ -0,0 +1,25 @@
...
+
+include(hunter_add_version)
+include(hunter_cacheable)
+include(hunter_cmake_args)
+include(hunter_download)
+include(hunter_pick_scheme)
+
+hunter_add_version(
+        PACKAGE_NAME
+        SocketIOClient
```
ruslo: Not relevant.

PR #1864, opened 2019-05-12
Add support for PACKAGE_UNRELOCATABLE_TEXT_FILES_OPTIONAL

epvbergen:

---

Implement support for PACKAGE_UNRELOCATABLE_TEXT_FILES_OPTIONAL, so that packages where the set of output files is dependent on user serviceable parts (CMAKE_ARGS) will not fail if the full set of files that must be patched is not present. Works exactly like PACKAGE_UNRELOCATABLE_TEXT_FILES otherwise.

### DIFF
--- a/cmake/modules/hunter_patch_unrelocatable_text_files.cmake
+++ b/cmake/modules/hunter_patch_unrelocatable_text_files.cmake
@@ -25,28 +25,33 @@ function(hunter_patch_unrelocatable_text_files)
   hunter_assert_not_empty_string("${x_TO}")
   hunter_assert_not_empty_string("${x_INSTALL_PREFIX}")
 
-  foreach(text_file ${HUNTER_PACKAGE_UNRELOCATABLE_TEXT_FILES})
+  foreach(text_file 
+      ${HUNTER_PACKAGE_UNRELOCATABLE_TEXT_FILES}
+      ${HUNTER_PACKAGE_UNRELOCATABLE_TEXT_FILES_OPTIONAL})
     set(text_full_path "${x_INSTALL_PREFIX}/${text_file}")
-    if(NOT EXISTS "${text_full_path}")
+    if(EXISTS "${text_full_path}")
+      hunter_status_debug("Patching text file: ${text_full_path}")
+      file(STRINGS "${text_full_path}" lines)
+
+      set(output_content "")
+      foreach(line ${lines})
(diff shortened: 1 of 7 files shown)


cmake/projects/gstreamer/hunter.cmake
```diff
@@ -36,7 +36,7 @@ hunter_cacheable(gstreamer)
 hunter_download(
     PACKAGE_NAME gstreamer
     PACKAGE_INTERNAL_DEPS_ID "4"
-    PACKAGE_UNRELOCATABLE_TEXT_FILES
+    PACKAGE_UNRELOCATABLE_TEXT_FILES_OPTIONAL
```
ruslo: I don't think they **all** should be optional; some files should present always (?).

cmake/modules/hunter_patch_unrelocatable_text_files.cmake
```diff
@@ -25,28 +25,33 @@ function(hunter_patch_unrelocatable_text_files)
...
+        string(REPLACE "${x_FROM}" "${x_TO}" line "${line}")
+        set(output_content "${output_content}\n${line}")
+      endforeach()
+
+      # if file is a link we should remove it first, otherwise we will
+      # update original file too
+      file(REMOVE "${text_full_path}")
+
+      file(WRITE "${text_full_path}" "${output_content}\n")
+    elseif(text_file IN_LIST HUNTER_PACKAGE_UNRELOCATABLE_TEXT_FILES_OPTIONAL)
```
ruslo: > My change will work with CMake 3.2 (minimum requirement for Hunter). [Yes]

`IN_LIST` is CMake 3.3 feature:
- https://cmake.org/cmake/help/latest/release/3.3.html#commands

ruslo: I don't mind updating to CMake 3.3, just send a separate pull request with documentation update:
- https://docs.hunter.sh/en/latest/quick-start/cmake.html

epvbergen: Missed that. Thanks, will do.

cmake/projects/gstreamer/hunter.cmake
epvbergen: Agreed for gstreamer itself, will fix that. For the plugins I guess most of not all will be optional.

ruslo:

PR #1869, opened 2019-05-18
add note about HUNTER_DOWNLOAD_SERVER behaviour

headupinclouds:
Add a note regarding the expected behaviour of the `HUNTER_DOWNLOAD_SERVER` variable.  It is applied only to standard `VERSION` variants of `hunter_config()` entries.  This comment gives a hint to the user that they can transform `URL/SHA1` entries explicitly with `hunter_download_server_url()` if they desire.

### DIFF
--- a/docs/reference/user-variables.rst
+++ b/docs/reference/user-variables.rst
@@ -351,8 +351,7 @@ To create new URLs the following template is used:
     ``VERSION`` variant of :ref:`hunter_config` entries, which is the case for all default
     Hunter package definitions.  Custom package definitions introduced with a ``URL``/``SHA1``
     variant on :ref:`hunter_config` in a project's local configuration, such as those included through
-    ``FILEPATH`` or ``LOCAL`` arguments to ``HunterGate()``, will be unaffected by this variable,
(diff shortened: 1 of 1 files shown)


docs/reference/user-variables.rst
```diff
@@ -345,6 +345,17 @@ To create new URLs the following template is used:
...
     This is the same structure as Hunter uses for its own :ref:`Download <layout deployed download>` directory.
 
+.. note::
+
+    ``HUNTER_DOWNLOAD_SERVER`` will be applied only to packages enabled with the standard 
+    ``VERSION`` variant of ``hunter_config()`` entries, which is the case for all default
+    Hunter package definitions.  Custom package definitions introduced with a `URL/SHA1``
+    variant on ``hunter_config()`` in a project's local configuration, such as those included through 
+    ``FILEPATH`` or ``LOCAL`` arguments to ``HunterGate()``, will be unaffected by this variable,
+    but can be similarly transformed through explicit calls to ``hunter_download_server_url()``.
```
ruslo: `hunter_download_server_url` is an internal module at this moment. I think if we need this functionality, then it's better to upgrade `hunter_config`.

PR #1901, opened 2019-06-24
Updated gstreamer and plugins to version 1.14.5

ErniBrown:

  * https://ci.appveyor.com/project/ErniBrown/hunter/builds/25498433
  * https://travis-ci.com/ErniBrown/hunter/builds/116653332


ErniBrown:
PS: I don't know why travis report the build [17.1](https://travis-ci.com/ErniBrown/hunter/jobs/210455067) as "in progress", but if you check the log it is completed, and exited with 0... maybe a bug, maybe some scheduler delay...

cmake/projects/gst_plugins_ugly/hunter.cmake
```diff
@@ -40,6 +51,7 @@ hunter_download(
     "lib/gstreamer-1.0/libgstasf.la"
     "lib/gstreamer-1.0/libgstdvdlpcmdec.la"
     "lib/gstreamer-1.0/libgstdvdsub.la"
-    "lib/gstreamer-1.0/libgstrmdemux.la"
```
ruslo: Since we are keeping the old version in `hunter.cmake` (user can switch to the old version in local config) we can't just remove this. I think the right approach here will be to check the version, see:
- https://github.com/ruslo/hunter/blob/v0.23.193/cmake/projects/xcb/hunter.cmake#L50-L52

ErniBrown: Ok, I'm working on that. By the way, I added the x264 dependency for the ugly plugin, but it is not needed. should I add an option, for example GST_PLUGIN_BASE_WITH_H264 ->add the dependency? If so is there another project I can use as a template?

ruslo: > should I add an option, for example GST_PLUGIN_BASE_WITH_H264 ->add the dependency?

It's already in `CMAKE_ARGS`, user can specify custom `DEPENDS_ON_PACKAGES` without `x264`. But probably it will affect `PACKAGE_UNRELOCATABLE_TEXT_FILES` so it will not work in practice. Workaround will be to use optional `PACKAGE_UNRELOCATABLE_TEXT_FILES`: https://github.com/ruslo/hunter/pull/1864

Custom schemes are hairy, as usual.

ErniBrown: But can I mix PACKAGE_UNRELOCATABLE_TEXT_FILES and PACKAGE_UNRELOCATABLE_TEXT_FILES_OPTIONAL? I tried, but I've got an error that "PACKAGE_UNRELOCATABLE_TEXT_FILES_OPTIONAL file" is not found...

ErniBrown: Yes, it is not working, if compiled without the x264 dependency it complains about the missing file. PACKAGE_UNRELOCATABLE_TEXT_FILE_OPTIONALS is still a PR, what would you like me to do now?

ruslo: > what would you like me to do now?

I'm okay with defaults if both new and old versions are working.

ErniBrown:
I've got another problem, the old version, 1.10.4 is not compiling for me. basically it's missing the opengl support. I supposed it was a problem of version 1.14.5, but it's not. It may be related to ubuntu switching to wayland, but I'm not completely sure.
In the end it is the same problem as x264, I can put it by default, and have opengl dependency added by default, completely remove it to avoid the complain about missing .la file, or wait for the PACKAGE_UNRELOCATABLE_TEXT_FILE_OPTIONALS PR to be completed. Since gl is something quite important, but not all the platform need it, i think the best option is to wait for the PR.

cmake/projects/gst_plugins_ugly/hunter.cmake
ruslo: > I'm okay with defaults if both new and old versions are working.

I've added `gst_plugins_good-1.10` example for testing.

ruslo:
> I can put it by default, and have opengl dependency added by default, completely remove it to avoid the complain about missing .la file

Is there an option to enable OpenGL or it's automatically picked if found?

ErniBrown:
> Is there an option to enable OpenGL or it's automatically picked if found?

There are some options, for enabling disabling, which can be set to yes/no/auto:

> --enable-opengl         Enable Desktop OpenGL support [default=auto]
  --enable-gles2          Enable OpenGL|ES 2.0 support [default=auto]
  --enable-egl            Enable EGL support [default=auto]
  --enable-wgl            Enable WGL support [default=auto]
  --enable-glx            Enable GLX support [default=auto]
  --enable-cocoa          Enable Cocoa support [default=auto]
  --enable-x11            Enable x11 support [default=auto]
  --enable-wayland        Enable Wayland support (requires EGL) [default=auto]

ruslo:
https://github.com/ruslo/hunter/blob/v0.23.214/README.rst

Issue #1739, opened 2019-02-16
Add readline


insufficientchocolate:
* I'm interested in package `readline`
* Here is official site of package https://tiswww.case.edu/php/chet/readline/rltop.html

* This package is using [Git]: https://git.savannah.gnu.org/git/readline.git

* I can't find any active mirror on github

* Used build system: `autotools`

* This package has no dependencies

---

tstack:
I'm currently working on this and an ncursesw package.  Mostly waiting for travis to get back on its feet at this point...

Issue #1191, opened 2017-11-12
Interface/implementation packages


ruslo:
There are packages that do implement same interface differently, examples:
* [musl](https://en.wikipedia.org/wiki/Musl) vs [glibc](https://en.wikipedia.org/wiki/GNU_C_Library)
* [libc++](http://libcxx.llvm.org/) vs [libstdc++](https://gcc.gnu.org/onlinedocs/libstdc++/faq.html#faq.what)
* [BoringSSL](https://github.com/google/boringssl) vs [OpenSSL](https://github.com/openssl/openssl)

It looks this way: there is standard, say "Foo Standard" that defines headers and libraries. Say library should define API:
```cpp
namespace foo {
int mult_by_2(int a);
}
```

"Foo Standard" on user's side:
```cpp
#include <foo/foo.hpp>

int main() {
  int value = 42;
  std::cout << "Result: " << foo::mult_by_2(value) << std::endl;
}
```

Implementation libraries OpenFoo and LibreFoo:
```cpp
// OpenFoo
int mult_by_2(int a) {
  return a * 2;
}
```

```cpp
// LibreFoo
int mult_by_2(int a) {
  return a << 1;
}
```

Because of the `#include <foo/foo.hpp>` interface defined by "Foo Standard" and because of the ODR rule these libraries can't be used simultaneously. Currently this code will lead to illegal Hunter state (effectively one library will overwrite files of other library):
```cmake
hunter_add_package(OpenFoo)
hunter_add_package(LibreFoo)
```

It should be implemented something like this:
```cmake
hunter_add_package(Foo)
```

With some reasonable default choice, say `OpenFoo`. User should be able to switch to different implementation by `CMAKE_ARGS`:
```cmake
hunter_config(
    Foo
    VERSION ${HUNTER_Foo_VERSION}
    CMAKE_ARGS FOO_IMPLEMENTATION=LibreFoo
)
```

Note that this will switch version **for all 3rd parties** (which is correct and intended).

User should never call `hunter_add_package(OpenFoo)` or `hunter_add_package(LibreFoo)` explicitly (add some checker?).

xsacha:
I have this issue with CURL right now where the CMake does not allow me to use BoringSSL.
The (upstream) CMakeLists actually does a find_package(OpenSSL) if CMAKE_USE_OPENSSL but, unlike the configure path, does not allow me to choose an alternative.

isaachier:
[...]

PR #1912, opened 2019-07-08
add readline-6.3 package

tstack:
* I've followed [this guide](https://docs.hunter.sh/en/latest/creating-new/create/cmake.html)

* I have submitted CI configs to https://github.com/ingenue/hunter targeting `pkg.template` branch,
  see this merged pull request https://github.com/ingenue/hunter/pull/380

* My change will resolve this "New package" request: https://github.com/ruslo/hunter/issues/1739

---

### DIFF
--- a/cmake/find/Findreadline.cmake
+++ b/cmake/find/Findreadline.cmake
@@ -1,0 +1,20 @@
+
+# The readline library is autotools-based, but does not include
+# pkgconfig files.  So, a find module is needed to locate the outputs.
+find_library(readline_library "libreadline.a")
+find_library(history_library "libhistory.a")
+find_path(readline_INCLUDE_DIRS readline/readline.h)
+
+add_library("readline::readline" UNKNOWN IMPORTED)
+set_target_properties("readline::readline"
+	PROPERTIES
+	INTERFACE_INCLUDE_DIRECTORIES "${readline_INCLUDE_DIRS}"
+	IMPORTED_LOCATION "${readline_library}"
+)
+
+add_library("readline::history" UNKNOWN IMPORTED)
+set_target_properties("readline::history"
+	PROPERTIES
+	INTERFACE_INCLUDE_DIRECTORIES "${readline_INCLUDE_DIRS}"
+	IMPORTED_LOCATION "${history_library}"
+)
--- a/cmake/find/Findreadline.cmake
+++ b/cmake/find/Findreadline.cmake
@@ -5,11 +5,14 @@ find_library(readline_library "libreadline.a")
 find_library(history_library "libhistory.a")
 find_path(readline_INCLUDE_DIRS readline/readline.h)
 
+find_package(ncursesw CONFIG REQUIRED)
+
 add_library("readline::readline" UNKNOWN IMPORTED)
 set_target_properties("readline::readline"
 	PROPERTIES
 	INTERFACE_INCLUDE_DIRECTORIES "${readline_INCLUDE_DIRS}"
 	IMPORTED_LOCATION "${readline_library}"
+	INTERFACE_LINK_LIBRARIES PkgConfig::ncursesw
 )
 
 add_library("readline::history" UNKNOWN IMPORTED)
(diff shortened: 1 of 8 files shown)


cmake/projects/readline/hunter.cmake
```diff
@@ -0,0 +1,32 @@
...
+    017b92dc7fd4e636a2b5c9265a77ccc05798c9e1
+)
+
+hunter_configuration_types(readline CONFIGURATION_TYPES Release)
+hunter_pick_scheme(DEFAULT url_sha1_autotools)
+
+hunter_cacheable(readline)
+hunter_download(
+    PACKAGE_NAME readline
+    PACKAGE_COMPONENT readline
```
ruslo: I don't think `PACKAGE_COMPONENT` is needed.

cmake/find/Findreadline.cmake
```diff
@@ -0,0 +1,20 @@
+
+# The readline library is autotools-based, but does not include
+# pkgconfig files.  So, a find module is needed to locate the outputs.
```
ruslo: It seems that version 8.0 has `*.pc` files.

examples/readline/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+cmake_minimum_required(VERSION 3.2)
+
+# Emulate HunterGate:
+# * https://github.com/hunter-packages/gate
+include("../common.cmake")
+
+project(download-readline)
+
+# DOCUMENTATION_START {
+# readline depends on a termcap implementation.  One
```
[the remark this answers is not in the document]
tstack: The dependency can be provided by other packages, though, not just ncurses.  So, I didn't want to make a hard dependency.  Is there another way to handle that?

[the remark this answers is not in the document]
ruslo: `CMAKE_ARGS` can be customized by a user, so it should not be a problem:
- https://github.com/ruslo/hunter/wiki/example.custom.config.id

I'm not sure about the details, but I guess this one is related:
- https://github.com/ruslo/hunter/issues/1191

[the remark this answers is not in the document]
tstack: Added the dependency

docs/packages/pkg/readline.rst
```diff
@@ -0,0 +1,21 @@
...
+  single: terminal ; readline
+
+.. _pkg.readline:
+
+readline
+========
+
+-  `Official <https://tiswww.case.edu/php/chet/readline/rltop.html>`__
+-  `Example <https://github.com/ruslo/hunter/blob/master/examples/readline/CMakeLists.txt>`__
+-  Added by `Tim Stack <https://github.com/tstack>`__ (`pr-N <https://github.com/ruslo/hunter/pull/N>`__)
```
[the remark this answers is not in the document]
tstack: updated

cmake/projects/readline/hunter.cmake
tstack: removed

cmake/find/Findreadline.cmake
tstack: I'd still like to use the 6.3 version.

ruslo: The thing about writing `Find*.cmake` modules is that they are either hard or even impossible to create correctly. E.g., as you said instead of "ncursesw" user can specify something else. Or a user may want to build a shared library, and the name will be `libreadline.so` instead of `libreadline.a`. I've tried a test with `HUNTER_BUILD_SHARED_LIBS=ON`, and it picks a `/usr/lib/x86_64-linux-gnu/libreadline.a` from the system instead of the library from Hunter root because there is only `~/.hunter/.../Install/lib/libreadline.so` present.

ruslo:
Testing:
- https://travis-ci.org/ingenue/hunter/builds/555960398?utm_source=github_status&utm_medium=notification
- https://ci.appveyor.com/project/ingenue/hunter/builds/25830122

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.23.200

PR #1918, opened 2019-07-10
Update documentation for lib PNG

tjom:
* I've checked this [Git style guide](https://0.readthedocs.io/en/latest/git.html). **Yes**
* I've checked this [CMake style guide](https://0.readthedocs.io/en/latest/cmake.html). **Yes**
* My change will work with CMake 3.2 (minimum requirement for Hunter). **Yes**
* I will try to keep this pull request as small as possible and will try not to mix unrelated features. **Yes**

---

I added a note to the lib PNG documentation that the PNG::PNG target will only be available for CMake versions greater or equal to 3.11.


docs/packages/pkg/PNG.rst
```diff
@@ -18,9 +18,10 @@ PNG
   :start-after: # DOCUMENTATION_START {
   :end-before: # DOCUMENTATION_END }
 
-For compatibility with
+For compatibility with the
 `FindPNG <https://cmake.org/cmake/help/latest/module/FindPNG.html>`__
-module, the ``ALIAS`` target ``PNG::PNG`` can be used too:
+module, the ``ALIAS`` target ``PNG::PNG`` can be used too (requires CMake
```
ruslo: That's not applied for the current state. Please submit this change with a new PNG version.

Update instructions:
- https://docs.hunter.sh/en/latest/creating-new/update.html

cmake/projects/PNG/hunter.cmake
```diff
@@ -87,6 +87,28 @@ hunter_add_version(
...
+    "https://github.com/hunter-packages/libpng/archive/v1.6.26-p6.tar.gz"
+    SHA1
+    dee61a3f48ed2d436a4e7fb8cdaf3cabaa3e604e
+)
+
+hunter_add_version(
+    PACKAGE_NAME
+    PNG
+    VERSION
+    "1.6.36-p3"
```
ruslo: Have you tested it? If yes, please update `hunter_default_version`, otherwise please remove this.

docs/packages/pkg/PNG.rst
tjom: Right, I added the latest 1.6.26 and 1.6.36 packages and updated the default version to the latest 1.6.26 package.

cmake/projects/PNG/hunter.cmake
tjom: I only tested it locally on Linux with gcc and the respective mingw cross compiler. That is why I left the default at 1.6.26, but since there is already a package for 1.6.36 I thought adding it would allow hunter users to upgrade manually and see if it works for their toolchain. But if you require CI testing I will remove it.

ruslo: I think there are issues with 1.36 version, see:
- https://github.com/hunter-packages/libpng/pull/5

So it's better to remove this version until they are fixed.

You can always do:
```cmake
hunter_config(
    PNG
    VERSION "1.6.36-p3"
    URL"https://github.com/hunter-packages/libpng/archive/v1.6.36-p3.tar.gz"
    SHA1 3ad1207c649eb17d840d75bef048fe9ee12fa00a
)
```
in your local config. See documentation:
- https://docs.hunter.sh/en/latest/reference/user-modules/hunter_config.html

tjom: > I think there are issues with 1.36 version, see:
> 
> * [hunter-packages/libpng#5](https://github.com/hunter-packages/libpng/pull/5)
> 
> So it's better to remove this version until they are fixed.

Okay, I will remove it from the PR.

> You can always do:
> 
> ```cmake
> hunter_config(
>     PNG
>     VERSION "1.6.36-p3"
>     URL"https://github.com/hunter-packages/libpng/archive/v1.6.36-p3.tar.gz"
>     SHA1 3ad1207c649eb17d840d75bef048fe9ee12fa00a
> )
> ```
> 
> in your local config.

Ah right, I forgot that you could also specify the URL, thank you!

ruslo:
Testing:
- https://ci.appveyor.com/project/ingenue/hunter/builds/25875010
- https://travis-ci.org/ingenue/hunter/builds/556753795?utm_source=github_status&utm_medium=notification

ruslo:
Released: https://github.com/ruslo/hunter/releases/tag/v0.23.203

